PR #51, opened 2022-07-18
 Add Pact Number Functions

ggobugi27:
- Moves `createExp` from `lang` to `util`, removes `JSON.stringify`ing arguments. 
- Creates `PactNumber` class with functions to print numbers in integers, strings, JSON.stringifed numbers, and`int` or `decimal` objects used in envData and clist. 
- Replaces `mkPactInt`  with `new PactNumber(val).toPactInteger`
- Replaces `mkPactDecimal` with `new PactNumber(val).toPactDecimal`


Randynamic (approved):

Randynamic:
Looks good!!! @alber70g Can you check this as well so we can merge this PR?

packages/libs/kadena.js/src/util/PactNumber.ts
```diff
@@ -0,0 +1,50 @@
...
+  public toStringifiedDecimal(): string {
+    return JSON.stringify(this.toDecimal());
+  }
+
+  public toPactDecimal(): IPactDecimal {
+    return {
+      decimal: this.toDecimal(),
+    };
+  }
+}
```
alber70g, line 50: Should we move it to a separate lib? We have two categories, low level libs, and consumer libs. This could be part of a low-level lib?

packages/libs/kadena.js/src/util/tests/createExp.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  const actual = createExp(
+    '+',
+    new PactNumber('2').toStringifiedInteger(),
+    new PactNumber('3').toStringifiedDecimal(),
+  );
+  console.log(actual);
+  const expected = '(+ "2" "3.0")';
+
+  expect(expected).toEqual(actual);
+});
```
alber70g, line 33: Can you wrap these in `describe("createExp", () => { it("takes in a function...) } }`

alber70g (changes requested):

rush.json
```diff
@@ -427,6 +427,13 @@
       "shouldPublish": true
     },
 
+    {
+      "packageName": "@kadena/pactjs-utils",
+      "projectFolder": "packages/libs/pactjs-utils",
+      "tags": ["libs"],
+      "shouldPublish": false
+    },
+
```
alber70g, line 436: Minuity: sorting in this file is done within the categories by `packageName`. So `@kadena/pactjs-utils` should be below `@kadena/integration-tests`

alber70g (dismissed):
Looks good!

packages/libs/pactjs-utils/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./node_modules/@kadena/heft-rig/profiles/default/tsconfig-base.json",
+  "compilerOptions": {
+    "types": ["heft-jest", "node"]
```
alber70g, line 4: Does it work without "node"? We found out that adding this to the types, the libs potentially don't work when using in the browser.

alber70g (approved):

PR #59, opened 2022-08-02
feat: pactjs client generator

alber70g:
This PR adds the ability to parse a Pact contract.
This can be used in the `pact-cli` to generate a TypeScript definition for `@kadena/pactjs`. This generated d.ts is a package called `@kadena/pactjs-generated`. This contains an interface that extends the `@kadena/pactjs::IPactModules` interface.

It can be used like so:

1. Generate the d.ts from a `coin.contract.pact`

```sh
npx @kadena/pactjs-cli generate --file "path/to/the/file/coin.contract.pact"
# this needs to be executed inside a package. 
# It relies on the existence of a package.json to define where the `@kadena/pactjs-generated` will be placed
```

2. Add the package to `types` in `tsconfig.json`

```json
{ "compilerOptions": { "types": ["@kadena/pactjs-generated"] } }
```

3. Consume the `@kadena/pactjs`

```ts
import { IFunctionResponse, Pact } from '@kadena/pactjs';

const builder: IFunctionResponse = Pact.modules.coin.transfer(
  'k:kadena',
  'k:kadena',
  1,
);

async function main(): Promise<void> {
  console.log('generate()', builder.generate());

  const res: Awaited<ReturnType<IFunctionResponse['call']>> =
    await builder.call();
  const text: string = await res.text();
  console.log('call()', text);
  console.log('returnType', builder.returnType);
}

main().catch(console.error);
```

## MVP:
- [x] fix the runtime of Pactjs to create an instance for each call (this is a js Proxy [3 MDN Proxy] that intercepts property gets and function apply calls) [4 Pactjs runtime]
- [x] put generated d.ts in a namespaced d.ts. `coin.d.ts` and `index.d.ts` should export
- [x] generation should output something that can be signed in Chainweaver
- [x] extend @kadena/pactjs interface to allow for adding metadata (sender, caps and signers) to each transaction [5 template metadata]
- [x] allow parsing `defcaps`
- [x] build very thin interface to allow signing (and maybe a concrete implementation to connect to Chainweaver somehow)  
  - [x] chainweaver quicksign POST /v1/quicksign
  - [x] chainweaver sign POST /v1/sign
- [x] sane defaults for `addMeta`
- [x] allow cli to parse templates (akin [kadena-io/tx-library](https://github.com/kadena-io/tx-library/blob/master/create-gas-station/template.yaml))
- [ ] allow to write custom command (without the generator). Requires refactoring to split the transaction builder and command generator

## Discussion
- is there a need for a cap to have a signer. From the chainweaver `/quickSign` endpoint it's mandatory, but for `/sign` it can be picked in the UI

## Near future:
- [ ] allow cli to parse contracts straight from the chain (e.g. generate --module coin --network testnet)
- [ ] allow to use utility types in the function calling the command (e.g.`Pact.modules.coin['transfer-create']("k:from","k:to", new Guard(), new PactNumber(100))`

## Further down the road
- [ ] analyse more of the contract and see which caps are mandatory (require-capability and compose-capability) (requires more work on the grammar)
- [ ] execute signing and posting to the api
- [ ] don't forget the `type: cont` 
- [ ] add zod.dev as runtime validation
- [ ] suggestions welcome...

[4 Pactjs runtime] https://github.com/kadena-community/kadena.js/pull/59/files#diff-4d9620e1499b60bb38913adb209e07414ed2ce77c1bcc52c261990cabb17fb09R14-R36
[5 template metadata] https://github.com/kadena-io/tx-library/blob/master/safe-transfer/template.yaml#L5-L21


packages/libs/pactjs-generator/src/lexer.js
```diff
@@ -0,0 +1,169 @@
...
+    lt: '<',
+    NOT_IS: /!=/,
+    NL: { match: /\n/, lineBreaks: true }
+});
+/**
+ * @internal
+ */
+function getModuleAndMethods(contract, logger) {
+    exports.lexer.reset(contract);
+    var token = undefined;
```
Randynamic, line 54: can we lose the `vars` => `let`?

Randynamic: Shouldn't our linter break on this?

alber70g: This is generated. It's a js file

packages/libs/pactjs-generator/src/lexer.ts
```diff
@@ -0,0 +1,266 @@
...
+    return acc.concat(cur);
+  }, []);
+}
+
+/**
+ * @internal
+ */
+export const flattenOnce = (arr: any[]) => {
+  return arr.reduce((acc, cur) => {
+    if (Array.isArray(cur)) {
```
Randynamic, line 67: Do we need this check as it does exactly the same as 68

packages/libs/pactjs-generator/src/tests/pact.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            name: 'address',
+            type: 'string',
+          },
+        },
+      },
+    });
+  });
+});
+
+// describe('pact generator', () => {
```
Randynamic, line 61: can we remove the unused tests?

Randynamic (dismissed):
Really awesome work @alber70g !!

packages/libs/pactjs-generator/src/lexer.ts
alber70g: Wdym?

packages/libs/pactjs-generator/src/tests/pact.test.ts
alber70g: Will do when finalizing this PR

packages/libs/pactjs/src/pact.ts
```diff
@@ -0,0 +1,207 @@
...
+    },
+    apply(
+      target: unknown,
+      that: unknown,
+      args: Array<string | number | boolean>,
+    ) {
+      // when the expression is called, finalize the call
+      // e.g.: `Pact.modules.coin.transfer(...someArgs)`
+      log('apply', args);
+      const final: string = `(${transaction.code} ${args
```
ggobugi27, line 189: You can use `createExp(transaction.code, ...args)` from pactjs

packages/libs/pactjs/src/PactExpression.ts
```diff
@@ -0,0 +1,33 @@
...
+      get(target: unknown, prop: string, receiver: unknown) {
+        if (prop === 'generate') {
+          expression = `(${path.join('.')}${args.length > 0 ? ' ' : ''}${args
+            .map(parseType)
+            .join(' ')})`;
+          return () => expression;
+        }
+        path.push(prop);
+        return builder;
+      },
```
ggobugi27, line 24: Same as last comment, `createExp`. Also, there can be nested pact codes, like `(func (read-keyset 'hi ))`, so we should add pact code as input type? 
 There also needs to be separation between strings and pact values, i.e.) `(func some-val "arg")` where we don't want some-val to be wrapped in quotations but we want `arg` to be stringified.  - I think something to add in pactjs

packages/libs/pactjs/src/IPactCommand.ts
```diff
@@ -0,0 +1,38 @@
+/**
```
ggobugi27, line 1: We have `ICap`, `IMetaData (=IPublicMeta)`, `ISignedTransaction(= ISignedCommand)` , `ICommandPayload(=IUnsignedTransactionCommand)` , etc in @kadena/types : https://github.com/kadena-community/kadena.js/blob/master/packages/libs/types/src/PactCommand.ts
Should we have one place for all pact value / tx types?

packages/libs/pactjs-test-project/src/transfer-create.ts
```diff
@@ -0,0 +1,41 @@
...
+const data: Record<string, unknown> = {
+  ks: {
+    keys: ['554754f48b16df24b552f6832dda090642ed9658559fef9f3ee1bb4637ea7c94'],
+    pred: 'keys-all',
+  },
+};
+
+const unsignedTransaction: IUnsignedTransaction = Pact.modules.coin[
+  'transfer-create'
+](sender, receiver, () => 'read-keyset ks', amount)
```
ggobugi27, line 24: `createExp("read-keyset", "ks")` can be used here

packages/libs/pactjs-test-project/src/transfer.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '@kadena/pactjs';
+
+const sender: string = 'croesus';
+const receiver: string =
+  'k:554754f48b16df24b552f6832dda090642ed9658559fef9f3ee1bb4637ea7c94';
+// const sender: string = receiver;
+const senderNoK: string =
+  '2993f795d133fa5d0fd877a641cabc8b28cd36147f666988cacbaa4379d1ff93';
+// const senderNoK: string = sender.split('k:')[1];
+const amount: number = 1000.0;
```
ggobugi27, line 14: `PactNumber` can be used here

packages/libs/pactjs/src/PactExpression.ts
alber70g: Yes. I'm now using `parseType` to "stringify" to parse the arguments based on their inputs, but that doesn't fix the type for `some-value` vs `"some-string"`. The generator should probably be aware of the type that's been passed. Maybe we need to create PactString, PactValue, PactSymbol, etc.

packages/libs/pactjs-test-project/src/transfer-create.ts
alber70g: Yes that's one option. I'd rather have a list of classes that are exported from @kadena/pactjs that can be used as `guard`. Something like `type PactGuard = ReadKeyset | Keyset | OtherGuard`. Same goes for the other types. I'd like to have the flexiblity where we can map the javascript primitives dynamically to Pact Types (like PactGuard) so the consumer isn't required to use those classes.

packages/libs/pactjs-test-project/src/transfer.ts
alber70g: This is the consuming code. I'm not really a fan of using the complex types in the consuming code. I'd like the level of entry to be as low as possible. We now dynamically convert number into PactNumber. See also https://github.com/kadena-community/kadena.js/pull/59#discussion_r964748673

packages/libs/pactjs/src/IPactCommand.ts
alber70g: This is great! I'll use those from @kadena/types. I'm still not sure what would be the best way to define types. I'm much a proponent of defining the type at the place where it's being used, not separate them too much... So I'm conflicted about how we should use this.

PR #68, opened 2022-09-01
Added the unique API to crypto utils

Randynamic:
## Reason:
Moved implementation from `kadena.js` package to `cryptography-utils`


ggobugi27 (approved):

ggobugi27:
Looks like the imports failed in `createListenRequest`, and other `kadena.js/api` functions

packages/libs/cryptography-utils/src/unique.ts
```diff
@@ -0,0 +1,38 @@
...
+export function unique(hashes: Array<IBase64Url>): Array<IBase64Url> {
+  const isUnique: IHashes = {};
+  return hashes.filter((hash) => {
+    if (!isUnique[hash]) {
+      isUnique[hash] = true;
+      return true;
+    }
+    return false;
+  });
+}
```
alber70g, line 38: Wouldn't it be better to leave filtering doubles to the consuming library?

Randynamic: I agree as this should only be a helper function.
But we first need to comply with the previous `pact-lang-api` API before deprecating or changing the API

packages/libs/cryptography-utils/src/unique.ts
```diff
@@ -0,0 +1,36 @@
...
+export function unique(array: Array<string>): Array<string> {
+  const isUnique: ICache = {};
+  return array.filter((item: string) => {
+    if (!isUnique[item]) {
+      isUnique[item] = true;
+      return true;
+    }
+    return false;
+  });
+}
```
alber70g, line 36: Maybe we can make this generic?
```suggestion
export function unique<T extends string>(hashes: Array<T>): Array<T> {
  const isUnique: Set<T> = {} as Set<T>;
   hashes.forEach((hash) => {
    if (!isUnique.has(hash)) {
      isUnique.add(hash);
    }
  });
  return Array.from(isUnique)
}
```

packages/libs/cryptography-utils/src/unique.ts
alber70g: Can we add the [`@deprecated`](https://api-extractor.com/pages/tsdoc/tag_deprecated/) tag?

ggobugi27 (approved):

PR #80, opened 2022-10-20
Cleanup: move project to src folder, eslint config, tsconfig, update …

ashwinvandijk:
Multiple cleanup actions have been taken, including actually using the eslint plugins, moving the explorer to a `src` folder, and cleaning up redundant lines from `.gitignore`.


Randynamic (approved):

packages/apps/explorer/.eslintrc.js
```diff
@@ -3,6 +3,14 @@
 require('@rushstack/eslint-config/patch/modern-module-resolution');
 
 module.exports = {
-  extends: ['@kadena/eslint-config/profile/lib'],
+  extends: [
+    '@kadena/eslint-config/profile/lib',
+    'eslint-config-airbnb-typescript',
+    '@next/eslint-plugin-next',
+    'eslint-config-prettier',
```
alber70g, line 10: Not sure if this is necessary. We should have an automatic prettier run on commit, therefore there's no need to have also a linter. Unless this is for the other plugins? https://github.com/kadena-community/kadena.js/blob/master/common/git-hooks/pre-commit

packages/apps/explorer/.eslintrc.js
```diff
@@ -3,6 +3,14 @@
 require('@rushstack/eslint-config/patch/modern-module-resolution');
 
 module.exports = {
-  extends: ['@kadena/eslint-config/profile/lib'],
+  extends: [
+    '@kadena/eslint-config/profile/lib',
+    'eslint-config-airbnb-typescript',
```
alber70g, line 8: I think this conflicts with the rules we have from the rig. I think this can be removed (might result in a bunch of errors that need to be fixed)

packages/apps/explorer/.eslintrc.js
```diff
@@ -3,6 +3,14 @@
...
+    '@next/eslint-plugin-next',
+    'eslint-config-prettier',
+    'eslint-plugin-jsx-a11y',
+    'eslint-plugin-react',
+    'eslint-plugin-unused-imports',
```
alber70g, line 13: This is also part of the rig rules

packages/apps/explorer/package.json
```diff
@@ -1,18 +1,20 @@
...
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
-    "test": "echo test",
+    "test": "",
+    "lint-staged": "lint-staged",
     "prisma:generate": "npx prisma generate",
     "create:info:test": "curl --request GET 'http://web:3000/api/info/create?network=Testnet'",
     "create:info:main": "curl --request GET 'http://web:3000/api/info/create?network=Mainnet'"
   },
```
alber70g, line 15: Could be that the fixed ci requires two new commands `_phase:build` and `_phase:test`. But as this is an app, and not a lib or tool I like it better if there's no phases here.

packages/apps/explorer/src/pages/mainnet/chain/[chainId]/block/[hash]/txs.tsx
```diff
@@ -7,7 +7,7 @@ import s from 'components/common/Transaction/TransactionDetails.module.css';
 import HistoryPage from 'components/common/Transaction/components/HistoryPage/HistoryPage';
 import { Loader } from 'components/common/Loader/Loader';
 import { NetworkName } from '../../../../../../utils/api';
-import { hasWindow } from 'utils';
+import { hasWindow } from 'src/utils';
 import { useBlockData } from 'utils/hooks';
```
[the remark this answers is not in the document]
Comment (ashwinvandijk) [reply packages/apps/explorer/src/pages/mainnet/chain/[chainId]/block/[hash]/txs.tsx:11]: Correct, I missed this one when moving the source to the `src` folder. As you mentioned we could also import the hooks from `utils`, but I prefer having the imports a bit organised by splitting up between the "normal" utils and the hooks.

packages/apps/explorer/package.json
ashwinvandijk: I've left these out for now

PR #85, opened 2022-10-26
Feat/kadena explorer lite

ashwinvandijk:
In the previous PR the plugins for eslint we're in the package.json, but weren't enforced on committing. This MR fixes this, and a bunch of eslint-errors/warnings. All the errors are fixed, there a still a lot of warnings that need to be addressed.


alber70g:
I'll approve. These are the typescript lint errors right?

Did you run `rush change`?

ashwinvandijk:
It's the test-coverage that's failing the build, there aren't any tests for the explorer and so the coverage isn't high enough. I don't think we want to add tests at this moment, so maybe exclude the complete explorer from the coverage for now?

Randynamic:
Please rebase with master an run `rush change` than it will work

alber70g (dismissed):

Randynamic (approved):

PR #92, opened 2022-10-28
version update

Randynamic:


alber70g (changes requested):
We shouldn't change minor to patch even in prerelease or alpha. Otherwise people don't know what might be breaking if we set everything to patch.

See #93

PR #102, opened 2022-11-08
feat(client): add `send` to PactCommand

alber70g:
- makes `signWithChainweaver` compatible with changes in `CommandBuilder`
I've updated the @kadena/client 
- transaction to have 3 new functions for /send /local /poll endpoints
- signWithChainweaver now returns a reference to the passed transactions and adds signatures to it

Updated the pactjs-test-projects transfer.ts accordingly
Changed some minor things about chainweb-node-client (needs thorough review, as I'm not aware who are using this)

TODO:

- [x] add `send`
- [x] add `local`
- [x] add `poll` to a successful transaction
- [x] add/update tests
- [x] update examples
- [x] update test project
- [x] update docs
- [ ] update launch post
- [ ] release new minor version

Another PR:
- [ ] remove `sigs` and `hash` and `cmd` when a non-malleable transactionbuilder gets changed
- [ ] add `setSigner` to `sign` straight from the `CommandBuilder`
- [ ] add `client`class and do server-type detection (Pact, P2P) and environment (dev (`pact -s`), testnet, mainnet)
  - [ ] be able to poll on the transaction
- [ ] change ICommandBuilder to be more like [Pact Commands](https://api.chainweb.com/openapi/pact.html)
- [ ] make 'nonce' changeable
- [ ] allow `type: cont` cross-chain transactions

---
  - https://app.asana.com/0/0/1203329012218796


packages/libs/chainweb-node-client/src/poll.ts
```diff
@@ -21,10 +20,10 @@ import fetch from 'cross-fetch';
 export function poll(
```
Randynamic, line 20: should we refactor this into an async function?

Randynamic (dismissed):

packages/libs/client/src/pact.ts
```diff
@@ -106,41 +142,38 @@ class PactCommand
...
       hash,
-      sigs: unsignedTransactionCommand.signers.reduce((acc, signer) => {
-        acc[signer.pubKey] = null;
-        return acc;
-      }, {} as Record<string, null>),
-      // sigs: unsignedTransactionCommand.signers.reduce((acc, signer) => {
-      //   acc.push({ hash, sig: undefined, pubKey: signer.pubKey });
-      //   return acc;
-      // }, [] as SignCommand[]),
+      sigs: this.sigs.map((s) => (s === undefined ? { sig: '' } : s)),
```
EnoF, line 160: can `s` never be `null`?

packages/libs/client/src/pact.ts
```diff
@@ -150,24 +183,80 @@ class PactCommand
...
+   * @alpha
+   */
+  public async send(apiHost: string): Promise<ISendResponse> {
+    const sendResponse = await send({ cmds: [this.createCommand()] }, apiHost);
+    this.requestKey = sendResponse.requestKeys[0].toString();
+    return sendResponse;
+  }
+
+  public poll(apiHost: string): Promise<IPollResponse> {
+    if (this.requestKey === undefined) {
```
EnoF, line 229: Is there a specific reason not to check on `falsy`, i.e. `!this.requestKey`?

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
```diff
@@ -16,25 +19,30 @@ describe('signWithChainweaver', () => {
...
-      })
-      .createTransaction();
-    const body = JSON.stringify({ reqs: [unsignedCommand] });
+    ).addCap('GAS', 'signer-key');
+
+    const { cmd, hash } = unsignedCommand.createCommand();
+    const sigs = unsignedCommand.sigs.reduce((sigs, sig, i) => {
+      const pubkey = unsignedCommand.signers[i].pubKey;
+      sigs[pubkey] =
+        sig === undefined ? null : sig.sig === undefined ? null : sig.sig;
```
EnoF, line 40: Can't we simplefy this to: `sig?.sig ?? null`?

packages/libs/pactjs-test-project/src/example-templates/generated-tx-templates/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  "transfer.json": (args: {
+    "fromAcct": string,
+    "toAcct": string,
+    "amount": string,
+    "chain": string,
+    "network": string,
+    "fromKey": string,
+  }): ICommandBuilder<{}> => {
+    const parts = [
+      "{\n  \"code\": \"(coin.transfer \\\"",
```
EnoF, line 51: Would it make sense to rewrite the string to a template string? So we could omit the escaping `\`?

EnoF (dismissed):
Just some small suggestions, nice stuff :+1:

packages/libs/client/src/pact.ts
EnoF: you could also simplify this to

```ts
this.sigs.map((s) => (s ?? { sig: '' }))
```

packages/libs/client/src/pact.ts
alber70g: See comment above

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
alber70g: See comment above

Issue #109, opened 2022-11-30
[@kadena/client] add a way to generate a apiHost based on a template and the tx contents


alber70g:
`PactCommand.setApiHost(apiHostTemplate: string)`
- `apiHostTemplate` is a string that contains `{apiVersion}` `{networkId}` and `{chainId}` and will be used when `.local()` `.send()` and `.poll()` are called.

```
tx.setApiHost("https://api.testnet.chainweb.com/chainweb/{apiVersion}/{networkId}/chain/{chainId}/pact")
```

alber70g:
We probably want a `Client` class that we can pass to the `Pact` instance so all derived `ICommandBuilder` instances have the same client.

Pseudo implementation:

```ts
import { createClient, Pact } from '@kadena/client';

const testnetPact = new Pact(createClient("https://api.testnet.chainweb.com/chainweb/{apiVersion}/{networkId}/chain/{chainId}/pact"));

export const Pact = testnetPact;
```

Issue #47, opened 2022-07-15
Make libraries node *and web* compatible


alber70g:
I think most packages will be used in both node and web environments. Therefore we need to make sure they are both web and node compatible.

Libraries like `node-fetch` can not be used in that case. As an alternative we could use [`cross-fetch`](https://www.npmjs.com/package/isomorphic-fetch)

We also have to think of a way of assuring this is the case, with tests or otherwise.

- [ ] add `web-lib` profile to `heft-rig`
- [x] change `node-fetch` to node _and web_ compatible lib
- [x] buffer is not browser compatible https://github.com/kadena-community/kadena.js/blob/76f148cf69c356204620120c38b9a4d25125d829/packages/libs/cryptography-utils/src/hexToBin.ts#L7
- [x] #77
- [ ] Find out whether we can run our tests on headless chrome, next to node, so we don't have to have separate e2e tests

PR #110, opened 2022-12-01
Cookbook Setup and Initial script

eileenmguo:
## Reason:

To create a Kadena.js version of the pact-lang-api cookbook 

Made minor updates to @kadena/client docs. Some other issues that I ran into was having to install additional kadena packages to get things working, but I will leave that for Albert to review and update in case he wants to make additional changes there.

## Check off the following:

- [x] I have added tutorials.


packages/tools/cookbook/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "scripts": {
+    "build": "",
+    "_phase:build": "",
+    "lint": "npx eslint ./mixins --ext .js,.ts --fix && npx eslint ./profile --ext .js,.ts --fix",
```
Takadenoshi, line 12: These paths do not exist. We can do one of:

- add linting 
-  remove this

packages/tools/cookbook/contracts/coin.module.pact
```diff
@@ -0,0 +1,681 @@
+(module coin GOVERNANCE
```
Takadenoshi, line 1: I don't mind including this file but maybe we can add a line in the README about checking the canonical repo for any coin contract upgrades - if/when we go to coin v6 we may forget to update this

Takadenoshi: Actually: is this included just for reference or is it required by code?

Takadenoshi:
Good stuff! LGTM aside from some very minor Qs and comments above.

packages/tools/cookbook/package.json
eileenmguo: I'll remove it 👍

packages/tools/cookbook/contracts/coin.module.pact
Takadenoshi: Oh cool, wasn't sure if it was there for users' reference

packages/tools/cookbook/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+  <picture>
+
+    <source srcset="https://github.com/kadena-community/kadena.js/raw/master/common/images/Kadena.JS_logo-white.png" media="(prefers-color-scheme: dark)"/>
+
+    <img src="https://github.com/kadena-community/kadena.js/raw/master/common/images/Kadena.JS_logo-black.png" width="200" alt="kadena.js logo" />
+
+  </picture>
+
+</p>
```
alber70g, line 16: This part should be an image but on preview it's shown as a codeblock. Maybe autoformatting is breaking this?

packages/tools/cookbook/package.json
```diff
@@ -0,0 +1,22 @@
...
+  },
+  "dependencies": {
+    "@kadena/client": "workspace:*",
+    "@kadena/chainweb-node-client": "workspace:*"
+  },
+  "devDependencies": {
+    "@types/node": "^16.0.0",
+    "@kadena/pactjs-cli": "workspace:*",
+    "@kadena/heft-rig": "workspace:*"
+  }
```
alber70g, line 21: Can we add the `lint-staged` command? See [for reference](https://github.com/kadena-community/kadena.js/blob/feat/cookbook/packages/libs/bootstrap-lib/package.json#L35-L43)

We probably want to have the whole package json look like the one from `bootstrap-lib`

packages/tools/cookbook/src/accounts/transfer-create.ts
```diff
@@ -0,0 +1,58 @@
+import { Pact, signWithChainweaver } from '@kadena/client';
+import { ISendResponse } from '@kadena/chainweb-node-client';
+import { accountKey, apiHost, pollTransactions } from '../utils';
+
+const HELP = `Usage example: \n\nts-node transfer-create.js k:{senderPublicKey} k:{receiverPublicKey} amount`;
+const NETWORK_ID = 'testnet04';
+const API_HOST = apiHost('1', 'testnet.', NETWORK_ID);
```
alber70g, line 7: #109 (trying to link this line with that issue)

packages/tools/cookbook/src/utils/account-key.ts
```diff
@@ -0,0 +1,5 @@
+// Returns the public key for the provided account name assuming
+// the naming follows the recommended convention: "k:publicKey"
+export function accountKey(account: string): string {
+  return account.split(':')[1];
```
alber70g, line 4: Can we do a check if this account is a namespaced account, and throw an error when it isn't. Gives a bit more info for the end user when this doesn't work (when they don't read the docs ;) )

packages/tools/cookbook/src/utils/api-host.ts
```diff
@@ -0,0 +1,9 @@
+// Composes the apiHost
+export function apiHost(
```
alber70g, line 2: Linking issue #109

packages/tools/cookbook/src/utils/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './account-key';
+export * from './api-host';
+export * from './poll-transactions';
+export * from './print-local';
+export * from './print-sig-data';
```
alber70g, line 5: We probably don't want to export any of these functions as the cookbook will not be used as a library?

And if we want to offer these functions, they should be exposed from different libs

packages/tools/cookbook/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./node_modules/@kadena/heft-rig/profiles/default/tsconfig-base.json",
+  "compilerOptions": {
+    "types": [".kadena/pactjs-generated", "node"]
```
alber70g, line 4: `node` should be removed here. See also #47

alber70g (changes requested):
The lib needs to change a bit to be Rush compatible. For reference you can take the [bootstrap-lib](https://github.com/kadena-community/kadena.js/blob/feat/cookbook/packages/libs/bootstrap-lib)

For all exported functions we need JSDoc. Check out this VSCode extension: [William's Document This](https://marketplace.visualstudio.com/items?itemName=WilliamRagstad.wr-docthis))

packages/tools/cookbook/src/utils/index.ts
eileenmguo: The intention was not to offer this functions; I was just exporting them for ease of use within the repository. Is there a reason why we wouldn't want to export them this way?

packages/tools/cookbook/package.json
eileenmguo: Took what seemed like the relevant items from bootstrap-lib!

packages/tools/cookbook/src/utils/index.ts
eileenmguo: ahh i see 👍

packages/tools/cookbook/src/utils/print-transactions-local.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Blocking/sync call to submit a command for non-transactional execution.
+ * In a blockchain environment this would be a node-local “dirty read”
+ *
+ * @export
+ * @param {((ICommandBuilder<ICoinCaps> & IPactCommand)[])} signedTransactions
+ * @param {string} apiHost
+ * @return {Promise<void>}
+ */
+export async function printTransactionsLocal(
```
alber70g, line 12: In typescript jsdoc, named [tsdoc](https://tsdoc.org/) you don't need to add the type. Does eslint not give warnings about this?
Same goes for `@export`, because there is an `export` keyword

alber70g (dismissed):
One minor comment. Applies to all exported files and tsdoc

packages/tools/cookbook/src/utils/print-transactions-local.ts
eileenmguo: Hmm I don't get any lint warnings, but I also wasn't getting any lint warnings for missing tsdoc :/

alber70g: Ah I see. It's only when it's being exported from the file that's in the `package.json`.`main` entry. On build, it checks the exports to see if there's a [release tag](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/#release-tags). 

So for this it's not needed I think, as it's not exposed as public api.

PR #120, opened 2022-12-09
feat(graph): retrieve blocks once per second from db

alber70g:


packages/apps/graph-client/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js
+  features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out
+[the Next.js GitHub repository](https://github.com/vercel/next.js/) - your
+feedback and contributions are welcome!
+
+## Deploy on Vercel
```
Randynamic, line 47: should we keep this in here?

packages/apps/graph/src/graph/Query/hello.ts
```diff
@@ -0,0 +1,10 @@
+import { builder } from '../builder';
```
Randynamic, line 1: Do we want to keep this in here as well?

Randynamic (approved):

packages/apps/graph-client/README.md
alber70g: I'll get rid of it in future PR

packages/apps/graph/src/graph/Query/hello.ts
alber70g: It's already removed, but didn't want to change this PR. Will be in next PR

packages/apps/graph/prisma/schema.prisma
```diff
@@ -0,0 +1,143 @@
...
+  @@index([from_acct, height(sort: Desc), idx], map: "transfers_from_acct_height_idx")
+  @@index([to_acct, height(sort: Desc), idx])
+  @@map("transfers")
+}
+
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
```
eileenmguo, line 136: Was this left intentionally?

eileenmguo (approved):
Looks good! Couldn't comment directly on the file, but might be good to remove the vercel svg at some point

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { keyframes, styled } from '../../styles/stitches.config';
+import { Box } from '../box';
+import { Text } from '../text';
+
+import { TimeTicker } from './time-ticker';
+
+import { RocketIcon, TimerIcon } from '@radix-ui/react-icons';
+import React from 'react';
+
+const Container: any = styled('div', {
```
nillo, line 11: can we abstract components/wrapper/styling to seperate files ex styles? i donno the current pattern but just a thought for reusability.

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
```diff
@@ -0,0 +1,94 @@
...
+  width: '$blockWidth',
+  background: '$$color',
+  borderRadius: '$md',
+  display: 'flex',
+  alignItems: 'center',
+  px: '$2',
+  py: '$1',
+});
+
+interface IChainBlockProps {
```
nillo, line 53: can we have a interfaces.ts for the each component.

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
```diff
@@ -0,0 +1,94 @@
...
+  color: string;
+  block?: Block;
+  textColor: string;
+}
+
+export function ChainBlock(props: IChainBlockProps): JSX.Element {
+  const { color, textColor, block } = props;
+
```
nillo, line 61: early return

if (!block) return null

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
```diff
@@ -0,0 +1,94 @@
...
+  py: '$1',
+});
+
+interface IChainBlockProps {
+  color: string;
+  block?: Block;
+  textColor: string;
+}
+
+export function ChainBlock(props: IChainBlockProps): JSX.Element {
```
nillo, line 59: return value: JSX.Element | null

packages/apps/graph-client/src/components/chainweb/chainweb-header.tsx
```diff
@@ -0,0 +1,49 @@
...
+export function ChainwebHeader(): JSX.Element {
+  return (
+    <Box
+      css={{
+        display: 'flex',
+        alignItems: 'center',
+      }}
+    >
+      <Box css={{ width: '$blockWidth', mx: '$1' }} />
+      {[...new Array(20)].map((_, index) => (
```
nillo, line 16: dont use key as index

packages/apps/graph-client/src/pages/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  flexDirection: 'column',
+  my: '5rem',
+});
+
+export default function Home(): JSX.Element {
+  const { loading, data } = useGetBlocksSubscription();
+  const dataString = JSON.stringify(data);
```
nillo, line 20: usePrevious or https://github.com/kentcdodds/use-deep-compare-effect
to prevent performanceloss, this hook has a memoize function within it ,preventing this from happening on each rerender.

packages/apps/graph-client/src/pages/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }, [dataString]);
+
+  return (
+    <div>
+      <Head>
+        <title>Kadena Graph Client</title>
+        <link
+          rel="icon"
+          href="https://kadena.io/wp-content/uploads/2021/10/Favicon-V1.png"
```
nillo, line 36: use static

nillo (changes requested):
some minor feedback

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
eileenmguo: The `Container` component that is still rendered when there is no block data also applies styling so we wouldn't want to return null in this instance

packages/apps/graph-client/src/components/chainweb/chainweb-header.tsx
eileenmguo: Normally using the index as a key is an anti-pattern since it can cause more of the component tree to potentially rerender, but in this case I don't think it causes any issues since the array never changes. The array is just being used to create 20 labels using the index. Are you ok with leaving this as is or do you have a recommendation for a better way to implement this?

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
eileenmguo: I would be fine with doing this! I'm not sure if there is a current pattern we follow so I just left them in the same file. I think it could be cleaner to move them to their own file so I'll make this update 👍

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
KdWarlordz: i can live with both your suggestions:)

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
KdWarlordz: a yes , missed that

packages/apps/graph-client/src/components/chainweb/chainweb-header.tsx
KdWarlordz: nope if the data is static and there is no removal of items from array causing it to re-render , your solution will do fine:)

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
nillo: a yes , missed that

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
nillo: i can live with both your suggestions:)

packages/apps/graph-client/src/components/chainweb/chainweb-header.tsx
nillo: ur solution is fine (sorry commented from old account)

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
Randynamic: As long as we don't use the interface in other places IMHO it's better to have them in the same file

packages/apps/graph-client/src/pages/index.tsx
eileenmguo: useDeepCompareEffect didn't work since data can be undefined (there is an open issue on this) so I opted for usePrevious + lodash.isEqual comparison

packages/apps/graph-client/src/utils/hooks/use-parsed-blocks.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { useCallback, useState } from 'react';
+
+interface IUseParseBlocksReturn {
+  allBlocks: Record<number, Block[]>;
+  addBlocks: (blocks: Block[]) => void;
+}
+
+export function useParsedBlocks(): IUseParseBlocksReturn {
+  const [allBlocks, setAllBlocks] = useState<Record<number, Block[]>>({});
```
alber70g, line 11: You're using the `Block` type here. This is a nice and easy way to define the interface for `useParsedBlocks` but it's confusing and breaks typings.

You're essentially saying: `addBlocks` can accept only blocks that satisfy the `Block` type can be passed, but that's not what you want.

Therefore, it's best to make an interface that is _just_ what you need for this code to work.

Based on the implementation, you probably want this interface:

```ts
interface IBlock {
  height: number;
  chainid: number;
}
```

eileenmguo: makes sense! It has been updated to what you suggested

nillo (approved):

eileenmguo (approved):

packages/apps/graph/prisma/schema.prisma
alber70g: It's being generated by the prisma cli. Will remove at some point.

packages/apps/graph-client/src/components/chainweb/chain-block.tsx
alber70g: I'd leave them in the component file where it's used. Lowers cognitive load when needed to change

packages/apps/graph-client/src/utils/hooks/use-parsed-blocks.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+interface IUseParseBlocksReturn {
+  allBlocks: Record<number, IBlock[]>;
+  addBlocks: (blocks: IBlock[]) => void;
+}
+
+export function useParsedBlocks(): IUseParseBlocksReturn {
+  const [allBlocks, setAllBlocks] = useState<Record<number, IBlock[]>>({});
+
+  const addBlocks = useCallback(
```
alber70g, line 15: useCallback can be typed with a generic. Then you don't need to type the arguments

```suggestion
  const addBlocks = useCallback<(blocks: IBlock[]) => void>(
```

PR #126, opened 2022-12-12
[libs/client] Adds a `callPollUntilTimeout` function to poll if a transaction succeeded

ash-vd:


packages/libs/client/src/tests/pact.test.ts
```diff
@@ -197,6 +200,101 @@ describe('Pact proxy', () => {
...
+    try {
+      await builder.callPollUntilTimeout('fake-api-host.local.co');
+    } catch (error) {
+      expectingError = error;
+    }
+
+    expect(expectingError.message).toContain('`requestKey` not found');
+  });
+
+  fit('returns a response after polling a succeeding transaction', async () => {
```
alber70g, line 231:
```suggestion
  it('returns a response after polling a succeeding transaction', async () => {
```

alber70g (changes requested):

alber70g (approved):

PR #129, opened 2022-12-15
[libs/client] Add pactjs helpers

ash-vd:


packages/libs/client/package.json
```diff
@@ -13,6 +13,9 @@
     "start": "ts-node --transpile-only src/index.ts",
     "lint-staged": "lint-staged",
     "build": "heft build --clean",
+    "pactjs:retrieve:contract": "pactjs retrieve-contract --out contracts/coin.contract.pact --module coin",
+    "pactjs:generate:contract": "pactjs contract-generate --file contracts/coin.contract.pact",
+    "dev:postinstall": "rushx pactjs:retrieve:contract; rushx pactjs:generate:contract",
```
ash-vd, line 18: If we use `postinstall` here it will also run on `npm i`.

alber70g (approved):

PR #134, opened 2022-12-20
[tools/cookbook] Updates the transfer-create recipe to use `pollUntil`

ash-vd:
depends on #135


Randynamic (dismissed):

packages/tools/cookbook/package.json
```diff
@@ -10,7 +10,10 @@
     "_phase:build": "",
     "_phase:test": "",
     "build": "",
+    "dev:postinstall": "rushx pactjs:retrieve:contract; rushx pactjs:generate:contract",
```
LindaOrtega, line 13: This worked after running `mkdir ./contracts` for me. Maybe it can be 
```
"dev:postinstall": "mkdir -p ./contracts; rushx pactjs:retrieve:contract; rushx pactjs:generate:contract",
```

packages/tools/cookbook/src/accounts/transfer-create.ts
```diff
@@ -49,22 +47,20 @@ async function transferCreate(
     .addCap('coin.TRANSFER', senderPublicKey, sender, receiver, amount)
     .setMeta({ sender }, NETWORK_ID);
 
-  const signedTransactions = await signWithChainweaver(transactionBuilder);
+  await signWithChainweaver(transactionBuilder);
 
-  const sendRequests = signedTransactions.map((tx) => {
-    console.log(`Sending transaction: ${tx.code}`);
-    return tx.send(API_HOST);
+  await transactionBuilder.send(API_HOST);
```
LindaOrtega, line 52: (for my own education)
I noticed we're no longer keeping a reference to the `signedTransactions`. Does the `transactionBuilder` keep a copy of the signed transactions in its state now?

packages/tools/cookbook/src/utils/account-key.ts
```diff
@@ -10,9 +6,15 @@ const HELP: string =
...
  */
 export function accountKey(account: string): string {
-  if (!namespacedAccountRegex.test(account)) {
-    console.info(HELP);
+  const regExp: RegExp = new RegExp(/^([a-z]+\:)([a-z0-9]{64})$/g);
+  const [, namespace, publicKey] = regExp.exec(account) || [];
+
+  if (!namespace || !publicKey) {
+    console.info(
+      `Account name ${account} doesn't follow the recommended convention: 'k:publicKey'`,
```
LindaOrtega, line 14: (nit) 
Do we want the error message here to highlight that not just `k:` accounts are allowed? For example, `t:publicKey` would pass the regex check.

LindaOrtega (dismissed):
LGTM!
I had a some small nitpick suggestions. I also added a question about the transaction builder since I'm re-familiarizing myself with the repo.

packages/tools/cookbook/src/accounts/transfer-create.ts
ash-vd: Nothing changed in the `transactionBuilder`, we just use it differently in the cookbook now. There was no need to keep track of multiple transactions here, since there is only one. Line 52 awaits the transaction to succeed. When you reject the transaction in Chainweaver or it fails in another way it will[ throw an error](https://github.com/kadena-community/kadena.js/blob/master/packages/libs/client/src/signing/signWithChainweaver.ts#L85).

Randynamic (approved):

alber70g (changes requested):
`requestKey` should not be on the `ICommandBuilder` but in `IPactCommand`.
Btw, `requestKey` is the same as the `hash`.
However, it should not be stored at all and should be generated from the command.

Another reason why the hash/requestKey is not exposed is that we do not want to put the user at risk signing something where only the hash is used to sign, instead of the full command.

alber70g:
Oh, I see it's already on IPactCommand. Why is it needed on `ICommandBuilder`?

nillo (approved):

ash-vd:
I've removed the `requestKey` from the `ICommandBuilder` as it was indeed not needed there.

While this PR is only to add the `pollUntil` to the cookbook, I guess it does make sense to change that function to a version we all like. I do however think that it's better to open a new issue for that, as we're using the requestKey in the older `poll` function as well. I don't see the problem with using the key for the polling functions, as there is no signing being done there.

alber70g (approved):

PR #135, opened 2022-12-20
[libs/client] Rename `callPollUntilTimeout` to `pollUntil` and add `onPoll` callback

ash-vd:


packages/libs/client/src/pact.ts
```diff
@@ -277,15 +281,24 @@ export class PactCommand
...
-  public callPollUntilTimeout(
+  public pollUntil(
     apiHost: string,
-    interval: number = 5000,
-    timeout: number = 60000,
+    options: {
+      interval?: number;
+      timeout?: number;
+      onPoll?: (result: IPollResponse) => void;
+    },
```
alber70g, line 295: Would it be an idea to destruct options to the arguments and do onPoll first. 
Then we can set defaults there, and get rid of the ones on line 301.

Especially with the 5 seconds interval and 3 minutes timeout as defaults, you wouldn't set them, but you _would_ want to set the onPoll almost always.

```suggestion
    onPoll?: (result: IPollResponse) => void;
    interval: number = 1000 * 5;  // 5 seconds
    timeout: number = 1000 * 60 * 3; // 3 minutes
```

alber70g (approved):

alber70g (dismissed):

packages/libs/client/src/pact.ts
ash-vd: Decided to keep the `options` to be more flexible. Another benefit I see is that we don't need a breaking change if we ever decide to add another optional param.

alber70g (approved):

PR #146, opened 2023-01-18
chore(lint): adds 'no-eslint-disable' rule

alber70g:


packages/apps/explorer/.eslintignore
```diff
@@ -3,3 +3,5 @@
 !.*.js
 **/dist/**
 **/coverage/**
+**/*.ts
```
jermaine150, line 6: What's the reason these files are ignored / is there a plan to revisit them later?

alber70g: We have a bunch of React specific linting errors/warnings that I'm not sure how to deal with yet. Especially every function needs to have a return type, which basically for every component is `JSX.Element`. Maybe it'd be better to customize a react set of rules for some projects

jermaine150: Got it. As discussed in Slack we'll take it offline and discuss on how to handle these.

jermaine150 (dismissed):

nillo (dismissed):
besides ts comments lgtm

Randynamic (approved):

jermaine150 (approved):

PR #150, opened 2023-01-26
feat(graph): adds completedBlockHeights and blocksFromHeight

alber70g:


packages/apps/graph/src/graph/objects/Block.ts
```diff
@@ -17,23 +19,44 @@ export default builder.prismaNode('Block', {
...
+      async resolve(query, parent, { events }, context, info) {
         return prismaClient.transaction.findMany({
           ...query,
           where: {
-            block: {
-              equals: parent.hash,
+            block: parent.hash,
+            requestkey: {
+              in: (
+                await prismaClient.$queryRaw<{ requestkey: string }[]>`
```
nillo, line 52: lgtm , just one small comment can't you abstract this yo a separate function, or and apply for line 33 and 52 since they are equal?

nillo (approved):

packages/apps/graph/src/graph/objects/Block.ts
alber70g: Yes, will do.

ash-vd (approved):

PR #154, opened 2023-01-31
chore(domain): adds domain model, based on APIs, and various inputs

alber70g:


docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  pactId: String
+  rollback: Boolean # | null
+  step: Int
+  data: Object
+}
+
+enum ChainwebNetwork {
+  mainnet01
+  testnet04
+}
```
edmundnoble, line 160: If you want to be able to interact with test versions including devnet versions, this may be best kept as a string.

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+
+type Cut {
+  height: BigInt
+  weight: Hash
+}
```
edmundnoble, line 165: Cuts have some more information, a cut in general is a mapping from chain IDs to block header hashes. The other properties are derived from those block headers.

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+type Block {
+  creationTime: MsFromPosixEpoch
+  parent: Block
+  height: BigInt
+  hash: Hash
+  chainId: Int
+  weight: Base64Url
+  featureFlags: Int
+  epochStart: MsFromPosixEpoch
+  adjacents: [Block]
```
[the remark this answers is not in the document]
larskuhtz: Not sure what would be a good solution. Maybe simply `Object`? 

Or maybe using a union type for adjacents:

```gql
type Adjacents0 {
   2: BlockHash!
   3: BlockHash!
   5: BlockHash!
}
type Adjacents1 {
   # ...
}
# ...

union Adjacents = Adjacents0 | Adjacents1 | ... | Adjacents19   
```

Or event creating one Block type per chain? But that would be pretty verbose and would make queries more complicated.

```gql
interface Block {
    # ...
}
type Block0 implements Block { 
    adjacents: Adjacents0!
}
type Block1 {
    adjacents: Adjacents1!
}
# ...
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  payloadHash: Base64Url # fk BlockPayload
```
larskuhtz, line 27:
```suggestion
  payloadHash: BlockPayloadHash!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  chainwebVersion: ChainwebNetwork
+  target: Base64Url
```
larskuhtz, line 29:
```suggestion
  target: BlockWeight!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+
+  payload: BlockPayload
+}
+
+type BlockPayload {
+  transactions: [Transaction] # pact realm
+  minerData: Account
+  transactionsHash: Base64Url
+  outputsHash: Base64Url
+  payloadHash: Base64Url
```
larskuhtz, line 40:
```suggestion
  transactionsHash: PayloadTransactionsHash!
  outputsHash: PayloadOutputsHash!
  payloadHash: PayloadHash
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
+scalar Hash
+scalar MsFromPosixEpoch
+scalar Base64Url
+scalar Int64
```
larskuhtz, line 4: The uses of this type below do not include numerical semantics.

```suggestion
scalar Word64
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+scalar BigInt
```
larskuhtz, line 5:
```suggestion
scalar BlockHeight
scalar BlockWeight
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+scalar Base64UrlTransaction
+scalar Object
+
```
larskuhtz, line 8: There is only a well defined fixed set of Chains for a given version of this API. Next time the chainweb graph changes, we would have to update this API in any case.

```suggestion
scalar ChainId

```

Or maybe even use an enum?

```gql
enum ChainId
    CHAIN_0
    CHAIN_1
    CHAIN_2
    # ...
    CHAIN_19
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+type Signature {
+  sig: String
+}
+
+type Block {
+  creationTime: MsFromPosixEpoch
+  parent: Block
+  height: BigInt
+  hash: Hash
+  chainId: Int
```
larskuhtz, line 22:
```suggestion
  chainId: ChainId!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  weight: Base64Url
+  featureFlags: Int
```
larskuhtz, line 24: This is just 8 bytes without any numerical semantics:

```suggestion
  featureFlags: Word64!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  epochStart: MsFromPosixEpoch
+  adjacents: [Block]
+  payloadHash: Base64Url # fk BlockPayload
+  chainwebVersion: ChainwebNetwork
+  target: Base64Url
+  nonce: Int64
```
larskuhtz, line 30: Just 8 bytes (64 bits) without any numerical meaning:

```suggestion
  nonce: Word64!
```

docs/domain/domain-model.gql
larskuhtz: It seems that the GraphQL does not support dictionaries. So there has to be a key for each chain:

```suggestion
type CutChain {
    height: BlockHeight!
    hash: BlockHash!
}

type CutHashes {
    0: CutChain!
    1: CutChain!
    2: CutChain!
    3: CutChain!
    4: CutChain!
    5: CutChain!
    6: CutChain!
    7: CutChain!
    8: CutChain!
    9: CutChain!
    10: CutChain!
    11: CutChain!
    12: CutChain!
    13: CutChain!
    14: CutChain!
    15: CutChain!
    16: CutChain!
    17: CutChain!
    18: CutChain!
    19: CutChain!
}

type Cut {
    hashes: CutHashes!
    origin: String
    weight: BlockWeight!
    height: BlockHeight!
    instance: ChainwebNetwork!
    id: CutHash!
}
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+
+type Block {
+  creationTime: MsFromPosixEpoch
+  parent: Block
+  height: BigInt
+  hash: Hash
+  chainId: Int
+  weight: Base64Url
+  featureFlags: Int
+  epochStart: MsFromPosixEpoch
```
larskuhtz, line 25:
```suggestion
  epochStart(format: DateFormat = ISO8601): Date!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  adjacents: [Block]
+  payloadHash: Base64Url # fk BlockPayload
+  chainwebVersion: ChainwebNetwork
```
larskuhtz, line 28:
```suggestion
  chainwebVersion: ChainwebNetwork!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  target: Base64Url
+  nonce: Int64
+
+  payload: BlockPayload
+}
+
+type BlockPayload {
+  transactions: [Transaction] # pact realm
+  minerData: Account
```
larskuhtz, line 37:
```suggestion
  transactions: [Transaction!]! # pact realm
  minerData: Account!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+  transactionsHash: Base64Url
+  outputsHash: Base64Url
+  payloadHash: Base64Url
+  coinbase: CoinbaseTransaction
```
larskuhtz, line 41:
```suggestion
  coinbase: CoinbaseTransaction!
```

docs/domain/domain-model.gql
```diff
@@ -0,0 +1,165 @@
...
+
+enum KeysetPredicateOptions {
+  KeysAllPredicate
+  KeysTwoPredicate
+  KeysAnyPredicate
+}
+
+type Transaction {
+  command: TransactionCommand
+  commandResult: TransactionCommandResult
```
larskuhtz, line 90: Some services may not have the results available (e.g. they are not needed for validation and in the future we decide to make storage of results in nodes optional). So, it probably makes sense to keep results `nullable`.

```suggestion
  command: TransactionCommand!
  commandResult: TransactionCommandResult
```

larskuhtz (commented):
The review proposes quite a few new custom scalars. Not sure if we want that. However, the chainweb data types (Block and BlockPayload) are part of the Merkle tree and thus have pretty git types and semantics. In particular hashes are just opaque identifiers.

Recursively specified types seem to be a potential source of DOS vectors. But I don't know how queries on recursively specified types are implemented in GraphQL and whether there is a build in solution for this. Also, references to adjacent blocks form a DAG (a huge one) and if GraphQL queries aren't careful about sharing, this would result in exponential blowup in the response. So, I propose to avoid recursive types.

Also almost (but not all fields) are `non-nullable`.

The appropriate GraphQL representation for `adjacents` of a `Block` is an open question.

docs/domain/domain-model.gql
larskuhtz: It seems that the use of numbers as keys (either directly or as string) isn't supported in GraphQL. So, the current JSON wouldn't work.

docs/domain/domain-model.gql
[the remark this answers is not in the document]
larskuhtz: Another problem is that the use of numbers as field keys (either directly or quoted as strings) is not supported by GraphQL. So, it is not clear how to best encode our chains. (Unfortunately, enums aren't supported as keys neither.)

PR #156, opened 2023-02-02
fix(client/signWithChainweaver) KIP-0015 update

ggobugi27:
## Reason:
`signWithChainweaver` fails with Chainweaver 2.2.3 which uses the upgraded signing-api: https://github.com/kadena-io/signing-api/pull/19, which changes the endpoint from `/quickSign` to `/quicksign` with request/response body changes. 

- Update `signWithChainweaver` function to support the new signing-api, and add necessary interfaces
- Fix `IUnsignedSignatureWithHash` to include `{sig: null}` option
- Change `pubkey` in `addSignatures` in `ICommandBuilder` to `pubKey`
- Update `signWithChainweaver` test

## Check off the following:

- [ V ] I have reviewed my changes and run the appropriate tests.
- [ V ] I have have run `rush change` to add the appropriate change logs.
- [ V ] I have added/edited docs.
- [ X ] I have added tutorials.
- [ X ] I have double checked and DEFINITELY added docs.


nillo (approved):

packages/libs/client/src/pact.ts
```diff
@@ -214,7 +214,7 @@ export class PactCommand
     // convert to IUnsignedTransaction
     const command: ICommand = {
       hash,
-      sigs: this.sigs.map((s) => (s === undefined ? { sig: '' } : s)),
+      sigs: this.sigs.map((s) => (s === undefined ? { sig: null } : s)),
```
EnoF, line 217: is it possible to receive `sigs: [null]`? Should we in that case check `!s` instead of `s === undefined`?

packages/libs/client/src/pact.ts
```diff
@@ -377,10 +377,12 @@ export class PactCommand
...
 
-  public addSignatures(...sigs: { pubkey: string; sig: string }[]): this {
-    sigs.forEach(({ pubkey, sig }) => {
+  public addSignatures(
+    ...sigs: { pubKey: string; sig: string | null }[]
+  ): this {
+    sigs.forEach(({ pubKey: key, sig }) => {
       const foundSignerIndex = this.signers.findIndex(
-        ({ pubKey }) => pubKey === pubkey,
+        ({ pubKey }) => pubKey === key,
```
EnoF, line 385: you could also opt to not destructure the pubkey and compare like:

```
sigs.forEach(({ pubKey, sig }) => {
  const foundSignerIndex = this.signers.findIndex(
    (signer) => signer.pubKey === pubKey
```
That way you don't have to rename pubkey to a more abstract key

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
```diff
@@ -36,16 +36,15 @@ describe('signWithChainweaver', () => {
...
     const sigs = unsignedCommand.sigs.reduce((sigs, sig, i) => {
       const pubkey = unsignedCommand.signers[i].pubKey;
-      sigs[pubkey] =
-        sig === undefined ? null : sig.sig === undefined ? null : sig.sig;
+      sigs[pubkey] = sig?.sig || null;
       return sigs;
       // eslint-disable-next-line @rushstack/no-new-null
     }, {} as { [pubkey: string]: string | null });
 
     const body = JSON.stringify({ reqs: [{ cmd, hash, sigs }] });
```
EnoF, line 44: Shouldn't this have changed from `reqs` to `cmdSigDatas`?

packages/libs/client/src/pact.ts
Randynamic: normally I'm not for this but is this particular case it makes a lot of sense for the readability of the code

Randynamic (dismissed):

packages/libs/client/src/signing/signWithChainweaver.ts
```diff
@@ -35,24 +28,25 @@ export async function signWithChainweaver<T1 extends string, T2>(
   ...transactions: (IPactCommand & ICommandBuilder<Record<T1, T2>>)[]
 ): Promise<(IPactCommand & ICommandBuilder<Record<T1, T2>>)[]> {
   const quickSignRequest: IChainweaverQuickSignRequestBody = {
-    reqs: transactions.map((t) => {
+    cmdSigDatas: transactions.map((t) => {
```
CryptoPascal31, line 31: As per KIP-0015, cmdSigDatas should be commandSigDatas

https://github.com/kadena-io/KIPs/blob/master/kip-0015.md#request

Maybe related to this:
https://github.com/kadena-io/chainweaver/issues/797

Randynamic: @CryptoPascal31, great catch. We need to change this as well to `commandSigDatas`.

ggobugi27: Thanks for the review. Let's finish this PR as the latest chainweaver is implemented with `cmdSigDatas` and update the function once chainweaver is released with the updated signing-api.

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
ggobugi27: Updated

Randynamic (approved):

packages/libs/client/src/pact.ts
```diff
@@ -220,7 +220,8 @@ export class PactCommand
     // convert to IUnsignedTransaction
     const command: ICommand = {
       hash,
-      sigs: this.sigs.map((s) => (s === undefined ? { sig: '' } : s)),
+      // eslint-disable-next-line @rushstack/no-new-null
+      sigs: this.sigs.map((s) => ({ sig: s?.sig ?? null })),
```
alber70g, line 224: Should this `null` value be dealt with in `local()` function

packages/libs/client/src/signing/signWithChainweaver.ts
```diff
@@ -63,31 +48,41 @@ export async function signWithChainweaver<T1 extends string, T2>(
...
+function isASigner(signer: IQuicksignSigner): signer is {
+  pubKey: string;
+  sig: string;
+} {
+  console.log(
+    'pubKey' in signer &&
+      'sig' in signer &&
+      signer.sig !== null &&
+      signer.pubKey.length > 0,
+  );
```
alber70g, line 81: Should the `console.log` stay?

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
```diff
@@ -6,20 +6,18 @@ jest.mock('cross-fetch', () => {
...
 
 describe('signWithChainweaver', () => {
   jest.setTimeout(1000);
 
-  it('makes a call on 127.0.0.1:9467/v1/quickSign with transaction', async () => {
+  it('makes a call on 127.0.0.1:9467/v1/quicksign with transaction', async () => {
     (fetch as jest.Mock).mockResolvedValue({
       status: 200,
-      text: () => JSON.stringify({ results: [] }),
+      text: () => JSON.stringify({ responses: [] }),
```
alber70g, line 20: We can use the types from the generated openapi spec to define the response

Eg 

```suggestion
      text: () => JSON.stringify({ responses: [] } as IQuicksignResponse),
```

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
```diff
@@ -122,4 +141,37 @@ describe('signWithChainweaver', () => {
       { sig: 'transfer-key-sig' },
     ]);
   });
+
+  it('Wallet signs but does not have the signer key and returns sig null', async () => {
```
alber70g, line 145: The `it` already refers to the `describe('signWithChainweaver')`. Also, we're not testing the wallet but the `signWithChainweaver` function.

```suggestion
  it('signs but does not have the signer key and returns sig null', async () => {
```

packages/libs/pactjs-test-project/src/example-templates/generated-tx-templates/index.ts
```diff
@@ -2,114 +2,114 @@
 import { buildUnsignedTransaction, ICommandBuilder } from '@kadena/client';
```
alber70g, line 2: I think this file needs to be formatted before commit

alber70g (changes requested):

packages/libs/client/src/pact.ts
```diff
@@ -58,7 +58,7 @@ export interface ICommandBuilder<
   poll(apiHost: string): Promise<IPollResponse>;
   addSignatures(
     ...sig: {
-      pubkey: string;
+      pubKey: string;
```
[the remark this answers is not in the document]
ggobugi27: signing-api takes in `{pubKey: .. , sig: ..}` and so I thought it'd be nice to keep them the same

packages/libs/client/src/signing/signWithChainweaver.ts
ggobugi27: removed

packages/libs/client/src/pact.ts
ggobugi27: `sigs` are signed before sent to `/local`, but now we take in invalid sig strings when `signatureVerification` is turned off. We can let sig be undefined here, but we need to create new type instead of `ICommand`. Let's do that in the PR where we work on the new `local`.
I wanted to have remove `sig: ''` here because `signing-api` assumes the signature is signed if it sees the empty string.

packages/libs/client/src/signing/tests/signWithChainweaver.test.ts
ggobugi27: I've added `IQuicksignResponseBody` interface

alber70g (approved):

PR #159, opened 2023-02-02
fix(client): generated code accepts IPactDecimal and IPactInteger for pact-contract-types `decimal` and `integer`

alber70g:


common/changes/@kadena/pactjs-generator/fix-client-convert-decimal-integer-ipactdecimal-ipactinteger_2023-02-02-14-57.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/pactjs-generator",
+      "comment": "BREAKING: Pact.modules functions will accept for numbers only IPactDecimal or IPactInteger",
```
EnoF, line 5: Since it is a breaking change, shouldn't we mark the type as minor? or major?

EnoF (approved):

common/changes/@kadena/pactjs-generator/fix-client-convert-decimal-integer-ipactdecimal-ipactinteger_2023-02-02-14-57.json
alber70g: Technically yes, but we're still in alpha versions and if we set this to minor/major it'll update the version accordingly... minor disadvantage of rush

PR #171, opened 2023-02-20
Adding angular template to create-kadena-app 

jermaine150:
Adding angular template to create-kadena-app 
Enabled support for angular in the cli
Cleanup old files

Next step is to include all template projects in the build and further preparations for release create-kadena-app

## Check off the following:

- [X] I have added/edited docs.
- [X] I have added tutorials.
- [X] I have double checked and DEFINITELY added docs.


packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
```diff
@@ -0,0 +1,36 @@
+<div>
+ <main class="grid">
+   <div class="card">
+     <div class="header">
+       <h1 class="title">Welcome to <span>Kadena!</span></h1>
+     </div>
+     <p>This is the Kadena starter template using nextjs to help you get started on your blockchain development. Use the form below
```
davyengone, line 7: @jermaine150 this template mentions Nextjs instead of Angular

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
```diff
@@ -0,0 +1,36 @@
...
+   </div>
+   <div class="card">
+     <h3 class="cardTitle">Interact with the blockchain</h3>
+     <section class="cardSection">
+       <h4>My Account</h4>
+       <input id="account" [(ngModel)]="account" placeholder="Please enter a valid k:account" />
+     </section>
+     <section class="cardSection">
+      <h4>Write Message</h4>
+       <textarea id="messageToWrite" [(ngModel)]="messageToWrite" [disabled]="account === '' || writeInProgress"></textarea>
```
davyengone, line 18: @jermaine150 here you're using a 2-way binding which is not necessary. I believe the actual scenario is to write into `messageToWrite` when the user types. You can achieve this with `[ngModel]="messageToWrite"` then when you click below, the variable `messageToWrite` will be populated with the user's input.

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
```diff
@@ -0,0 +1,36 @@
...
+       <h1 class="title">Welcome to <span>Kadena!</span></h1>
+     </div>
+     <p>This is the Kadena starter template using nextjs to help you get started on your blockchain development. Use the form below
+       to interact with the Kadena blockchain using <code>@kadena/client</code> and edit <code>src/app/app.component.html</code> to get started.</p>
+   </div>
+   <div class="card">
+     <h3 class="cardTitle">Interact with the blockchain</h3>
+     <section class="cardSection">
+       <h4>My Account</h4>
+       <input id="account" [(ngModel)]="account" placeholder="Please enter a valid k:account" />
```
davyengone, line 14: Same as the comment below about `ngModel` bi-directional binding!

davyengone (changes requested):
Hi @jermaine150, I added a few comments. I hope it helps.

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
kdlmaster: Thanks, i've updated it.

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
jermaine150: @davyengone like we discussed, `[ngModel]="messageToWrite"` will give one-way binding but only from source to view, while in this case we need it the other way around. 
An alternative would be to use `(ngModelChange)="currentHero.name=$event">` but I would prefer 2-way binding over that.

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
jermaine150: See the remark on the previous comment

packages/tools/create-kadena-app/templates/angular/src/app/app.component.html
davyengone: Yes, let's go with that approach. The Reactiveform approach is complicated for such an example

davyengone (approved):

Randynamic (approved):

packages/tools/create-kadena-app/templates/angular/src/app/app.service.ts
```diff
@@ -0,0 +1,60 @@
...
+      .setMeta({
+        ttl: 28000,
+        gasLimit: 100000,
+        chainId: kadenaChainId,
+        gasPrice: 0.000001,
+        sender: account,
+      }, kadenaNetworkId)
+
+      await signWithChainweaver(transactionBuilder)
+
```
alber70g, line 27: Should we add a comment indicating that this is changed by reference?
```suggestion

      // `signWithChainweaver` modifies the `transactionBuilder` that's passed, 
      //   so we use this further down
      // It also returns an array with the transactionBuilders that are passed 
      //   (if one is passed it'll be an array of 1)

      await signWithChainweaver(transactionBuilder)

```

packages/tools/create-kadena-app/templates/angular/src/environments/environment.ts
```diff
@@ -0,0 +1,12 @@
+import { ChainId, ChainwebNetworkId } from '@kadena/types';
+
+// this would normally be your mainnet configuration, but our test contract has only been deployed on testnet
```
alber70g, line 3:
```suggestion
// this would normally be your mainnet configuration, but our test contract `free.cka-message-store` has only been deployed on testnet
```

packages/tools/create-kadena-app/templates/angular/tsconfig.app.json
```diff
@@ -0,0 +1,14 @@
+/* To learn more about this file see: https://angular.io/config/tsconfig. */
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./out-tsc/app",
+    "types": [".kadena/pactjs-generated"]
+  },
```
alber70g, line 7: This is extended from `./tsconfig.json` so not necessary

```suggestion
    "outDir": "./out-tsc/app",
  },
```

alber70g (approved):
Minor points and nitpicks

packages/tools/create-kadena-app/templates/angular/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "angular",
+  "version": "0.1.0",
+  "description": "Angular starter project with @kadena/client integration",
+  "scripts": {
+    "build": "ng build",
+    "ng": "ng",
+    "pactjs:generate:contracts": "pactjs contract-generate --file ./pact/message-store.pact; pactjs contract-generate --file ./pact/root/coin-v5.pact",
```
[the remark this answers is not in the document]
jermaine150: The goal of this is to have a working, typed, interaction with the blockchain, which is the main reason we generate the contracts. I think it's a good idea to refer to general examples / tutorial of contract generation if that's available from the resources section. (we now only refer to the general docs section and the pact smart contract).

Btw, issue #172 Is already fixed right? I'll comment there

packages/tools/create-kadena-app/templates/angular/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "rxjs": "~7.8.0",
+    "tslib": "^2.3.0",
+    "zone.js": "~0.12.0"
+  },
+  "devDependencies": {
+    "@angular-devkit/build-angular": "^15.1.4",
+    "@angular/cli": "~15.1.4",
+    "@angular/compiler-cli": "^15.1.0",
+    "@kadena/pactjs-cli": "^0.0.9",
+    "@kadena/types": "^0.0.3",
```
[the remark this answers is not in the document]
jermaine150: Yes, types are actually being used through `@kadena/client`, I'll remove `@kadena/types`

PR #193, opened 2023-03-27
Retrieve and generate as one command

### LINKED ISSUE #173: [@kadena/pactjs-cli] allow the user to retrieve and generate in one command
**Is your feature request related to a problem? Please describe.**
Currently the commands are split up between retrieval of the contract and generation of the typings

**Describe the solution you'd like**
I like to have one command for retrieval and generation

**Describe alternatives you've considered**
Running both after one another but this requires the storing of a file. 

**Additional context**
Related to this is the `namespace` that's not included in the retrieved contract. So the namespace is lost in type generation. See also #127

ash-vd:
## Reason:
With these changes it's possible to retrieve a contract from chain by using the `--contract` option for the `generate` command from `pact-cli`.

Closes #173 

---
  - https://app.asana.com/0/0/1204080496441777


packages/libs/pactjs-generator/src/contract/generation/StringContractDefinition.ts
```diff
@@ -12,10 +12,24 @@ export class StringContractDefinition implements IContractDefinition {
...
+    this._setNamespace(namespace);
+  }
+
+  private _setNamespace(namespace: string): void {
+    if (!namespace) return;
+
+    Object.keys(this._raw).forEach((module) => {
+      this._raw[module].namespace = namespace;
+    });
   }
```
alber70g, line 33: Maybe better to leave namespace as optional value. So `namespace: string | undefined`. If it's undefined, it'll take it from the `parser()`. When it's not `undefined`, it'll overwrite whatever comes from the `parser()`

packages/tools/pactjs-cli/src/contract-generate/index.ts
```diff
@@ -170,7 +63,45 @@ export function contractGenerateCommand(
...
+      'Generate d.ts from Pact contract from the blockchain',
+    )
+    .option(
+      '--api',
+      'The API to use for retrieving the contract, i.e. https://api.chainweb.com/chainweb/0.0/mainnet01/chain/8/pact',
+    )
+    .addOption(
+      new Option(
+        '--chain <chainId>',
+        'The chainId to retrieve the contract from, i.e. 8. Defaults to 1.',
```
alber70g, line 77:
```suggestion
        'The chainId to retrieve the contract from, e.g. 8. Defaults to 1.',
```

packages/tools/pactjs-cli/src/contract-generate/tests/generate.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+const mockedRetrieveContractFromChain =
+  retrieveContractFromChain as jest.MockedFunction<
+    typeof retrieveContractFromChain
+  >;
+mockedRetrieveContractFromChain.mockImplementation(
+  mockedRetrieveContractFromChain as jest.MockedFunction<
+    typeof retrieveContractFromChain
+  >,
+);
```
alber70g, line 41: Maybe something like `mock-fs` could be used? https://www.npmjs.com/package/mock-fs

alber70g (dismissed):

Randynamic:
@all-contributors please add @ash-vd for Code

Randynamic:
@all-contributors please add @alber70g for Reviewed Pull Requests

Randynamic:
@all-contributors please add @alber70g for review

packages/tools/pactjs-cli/src/contract-generate/tests/generate.test.ts
ash-vd: I'd like to suggest keeping these these changes out of this PR. Streamlining the tests should be much easier after #197 or similar.

packages/libs/pactjs-generator/src/contract/generation/FileContractDefinition.ts
```diff
@@ -13,10 +13,24 @@ export class FileContractDefinition implements IContractDefinition {
   private _logger: ILogger;
   private _raw: Output;
 
-  public constructor(filePath: PathLike, logger: ILogger = () => {}) {
+  public constructor(
```
EnoF, line 16: as the amount of params are increasing, do we want to consider destructuring an object over a list of arguments? Do we expect this constructor to grow in the near future?

EnoF: Actually disregarding whether we want to grow the list, if I now want to provide a logger, but no namespace, I would have to provide a `null` or `undefined` in between. I'd rather have a object destructured as api in such case

packages/tools/pactjs-cli/src/contract-generate/tests/generate.test.ts
```diff
@@ -0,0 +1,156 @@
...
+const createAndRunProgram = async (
+  type: 'file' | 'chain',
+  options?: Partial<ContractGenerateOptions>,
+): Promise<void> => {
+  const program = new Command('generate');
+  (program.error as unknown) = mockProgramError;
+  const action = generate(program, '0.0.0');
+  let result;
+
+  if (type === 'file') {
```
EnoF, line 50: why don't we just return the result without assigning it to a result? With early returns it becomes easier to understand

packages/tools/pactjs-cli/src/utils/callLocal.ts
```diff
@@ -0,0 +1,40 @@
+import fetch from 'cross-fetch';
+
+export async function callLocal(
+  apiHost: string,
+  body: string,
+): Promise<{
```
EnoF, line 6: Not sure but can't you do something like `ReturnValueOf<fetch>`?

isaozler (approved):

packages/tools/pactjs-cli/src/utils/callLocal.ts
ash-vd: Don't think so. We're not returning the response of fetch as-is, but we're building our own object.

packages/libs/pactjs-generator/etc/pactjs-generator.api.md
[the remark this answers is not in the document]
ash-vd: This file is generated by [api-extractor](https://api-extractor.com/pages/setup/generating_docs/#using-api-documenter-to-generate-markdown), which spits out .md files. When we add documentation in the code like below the documentation will be added to these files as well. Since nothing is documented right now, it's just the code that's displayed.

```/**
 * This is some example documentation.
 * @alpha
 */```

File: packages/libs/pactjs-generator/src/contract/generation/FileContractDefinition.ts [hunk 7]
```diff
@@ -13,10 +13,28 @@ export class FileContractDefinition implements IContractDefinition {
   private _logger: ILogger;
   private _raw: Output;
 
-  public constructor(filePath: PathLike, logger: ILogger = () => {}) {
+  public constructor({
+    filePath,
```
Takadenoshi, line 17: `path` as above

packages/tools/pactjs-cli/src/utils/callLocal.ts
```diff
@@ -0,0 +1,40 @@
...
+  textResponse: string | undefined;
+  jsonResponse:
+    | {
```
Takadenoshi, line 9: What does the leading pipe `|` do here? I'm not familiar with this syntax

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -0,0 +1,186 @@
...
+
+import { ContractGenerateOptions } from './';
+
+import { Command } from 'commander';
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import mkdirp from 'mkdirp';
+import { dirname, join } from 'path';
+import rimraf from 'rimraf';
+
+export const TARGET_PACKAGE: '.kadena/pactjs-generated' =
```
Takadenoshi, line 18: Windows issue here due to incompatible slash separators? could wrap in normalize

Takadenoshi: As far as I can tell this is used with `path.join` so it isn't a problem yes, but if used as-is it could blow up.

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -0,0 +1,181 @@
...
+  '.kadena/pactjs-generated' as const;
+
+const shallowFindFile = (path: string, file: string): string | undefined => {
+  while (!existsSync(join(path, file))) {
+    path = join(path, '..');
+    if (path === '/') {
```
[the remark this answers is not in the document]
Takadenoshi:
```suggest
    if (path === '/' || path === '\\') {
```

packages/tools/pactjs-cli/src/utils/callLocal.ts
ash-vd: No idea either, but Prettier won't let me remove it.

packages/libs/pactjs-generator/etc/pactjs-generator.api.md
```diff
@@ -8,12 +8,13 @@
...
 import { PathLike } from 'fs';
 
-// Warning: (ae-forgotten-export) The symbol "IContractDefinition" needs to be exported by the entry point index.d.ts
-//
 // @alpha (undocumented)
 export class FileContractDefinition implements IContractDefinition {
-    // Warning: (ae-forgotten-export) The symbol "ILogger" needs to be exported by the entry point index.d.ts
-    constructor(filePath: PathLike, logger?: ILogger);
+    constructor({ filePath, namespace, logger, }: {
+        filePath: PathLike;
```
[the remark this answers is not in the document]
ash-vd: We use `filePath` in other places as well (not with a named param though), so I'd like to keep it consistent. Another argument is that a `path` could also be a folder, `filePath` makes it more explicit.

packages/tools/pactjs-cli/src/contract-generate/generate.ts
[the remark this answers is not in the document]
ash-vd: We're going to pick this up in a new issue: #244

Randynamic (approved):

PR #194, opened 2023-03-28
add new grid component

sstraatemans:
- [ ] I have added tutorials.


packages/libs/react-components/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+  children?: ReactNode;
+  colStart?: IColSpan;
+  colEnd?: IColSpan;
+  rowStart?: IColSpan;
+  rowEnd?: IColSpan;
+  area?: string;
+  bg?: IThemeColors;
+}
+
+const checkSpan = (
```
eileenmguo, line 21: What is this checking? Should it be validating that the colEnd is past colStart?

packages/libs/react-components/src/styles/stitches.config.ts
```diff
@@ -23,7 +24,16 @@ const sizes: Record<string, string> = {
...
+} as const;
+
+// eslint-disable-next-line @kadena-dev/typedef-var
+export const breakpoints = {
+  sm: `(min-width: ${640 / 16}rem)`,
+  md: `(min-width: ${768 / 16}rem)`,
+  lg: `(min-width: ${1024 / 16}rem)`,
+  xl: `(min-width: ${1280 / 16}rem)`,
+  '2xl': `(min-width: ${1536 / 16}rem)`,
+} as const;
```
eileenmguo, line 36: What is the reasoning for exporting these?

packages/libs/react-components/src/typings/index.ts
```diff
@@ -0,0 +1,6 @@
+import { IThemeBreakpointProps } from '../styles';
+
+export type Primitive = number | string | boolean;
+export type ICSSPropType =
+  | Primitive
+  | Partial<Record<IThemeBreakpointProps, Primitive>>;
```
eileenmguo, line 6: Is there a reason you opted for `color: { "sm": "$blue500", "md": "$gray500"}` instead of `color: { "@sm": "$blue500", "@md": "$gray500"}`?

The reason I ask is because I think the `@` symbol makes it very clear that we are referring to a breakpoint and this is a style `at size sm`. It also would help us reduce the number of types

eileenmguo: Additionally, I think it would be valuable to make this type more strict by making it a generic so that we can pass in the property we will apply it to. 
For example: backgroundColor 
`ICSSPropType<Stitches.PropertyValue<"backgroundColor">>`

packages/libs/react-components/src/components/Grid/GridContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const GridContainer: FC<IGridContainerProps> = ({
+  children,
+  gap,
+  templateAreas,
+  templateRows,
+  templateColumns,
+}) => {
+  const styles = convertStyles({
+    gridGap: gap ?? {},
```
eileenmguo, line 30: Is this necessary? My understanding is that gap can only be undefined or the value itself so would this work if it were just written as the following? `gridGap: gap,`

packages/libs/react-components/src/utils/convertStyles.ts
```diff
@@ -0,0 +1,27 @@
+import { Primitive } from '../typings';
+
+export const convertStyles = (
+  styles: Record<string, Primitive | Record<string, Primitive>>,
+): Record<string, Primitive | Record<string, Primitive>> => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const newStyles = {} as any;
```
eileenmguo, line 7: can we use unknown here?

packages/libs/react-components/src/utils/convertStyles.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (value === undefined) return;
+
+    if (typeof value === 'object') {
+      return Object.keys(value).forEach((mediaQueryKey) => {
+        if (newStyles['@' + mediaQueryKey] === undefined) {
+          newStyles['@' + mediaQueryKey] = {};
+        }
+
+        const prop = styles[key];
+        if (typeof prop === 'object') {
```
eileenmguo, line 18: I'm a bit confused. Isn't `prop` the same as `value` so it would be of type "object" by default?

packages/libs/react-components/src/utils/convertStyles.ts
```diff
@@ -0,0 +1,27 @@
...
+          newStyles['@' + mediaQueryKey][key] = prop[mediaQueryKey];
+        }
+      });
+    }
+    return (newStyles[key] = value);
+  });
+
+  return newStyles;
```
eileenmguo, line 26: I noticed this doesn't account for sizes smaller than what is defined. I think it would be beneficial to implement an `@initial` like stitches does so they can also set a default value
https://stitches.dev/docs/responsive-styles#setting-an-initial-variant

packages/libs/react-components/tsconfig.json
```diff
@@ -2,20 +2,7 @@
...
-    "outDir": "dist/esm",
-    "allowJs": true,
-    "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true,
-    "forceConsistentCasingInFileNames": true
-  },
-  "include": ["src/**/*"]
+    "outDir": "lib"
+  }
 }
```
eileenmguo, line 8: what was the reasoning for removing this? I believe this change will break the usage of this package since package.json is still pointing at `dist/esm`

eileenmguo (changes requested):
I noticed this PR doesn't take the design system spacing into account. Is there a reason for that? I think it would be beneficial to offer a variant that allows users to choose a predefined space from the design system so they don't have to think about which of the theme spaces aligns with the design. 

This is something that would be shared with the stack component so I can also just implement that in my PR and apply it to the grid component as well. Just wanted to check if there was a rational for not including it!

packages/libs/react-components/src/utils/convertStyles.ts
sstraatemans: it will break the function.
Albert and I took the best part of the morning to get this function properly typed. we failed. :/

packages/libs/react-components/src/components/Grid/GridContainer.tsx
sstraatemans: hmm..thinking about it again. I added undefined to the Primitive type (used in the convertStyles function)
makes it much better. thanks

packages/libs/react-components/src/styles/types.ts
```diff
@@ -0,0 +1,7 @@
+import { colors } from './colors';
+import { breakpoints, sizes } from './stitches.config';
+
+export type IThemeColors = `$${keyof typeof colors}`;
+export type IThemeSizes = `$${keyof typeof sizes}`;
+export type IThemeBreakpointProps = keyof typeof breakpoints;
+export type IStyleBreakpointProps = `@${keyof typeof breakpoints}`;
```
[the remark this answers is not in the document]
sstraatemans: makes sense

packages/libs/react-components/src/styles/stitches.config.ts
sstraatemans: for the typing. 
These are not in the theme object. so need them exported

packages/libs/react-components/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+export type IColSpan = number | 'auto';
+
+export interface IGridItemProps {
+  children?: ReactNode;
+  colStart?: IColSpan;
+  colEnd?: IColSpan;
+  rowStart?: IColSpan;
+  rowEnd?: IColSpan;
+  area?: string;
+  bg?: IThemeColors;
```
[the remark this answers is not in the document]
sstraatemans: but then we use all the colors. that is really flexible

packages/libs/react-components/src/components/Grid/GridItem.tsx
sstraatemans: we can add that check. 
But is is really just creating that string.

packages/libs/react-components/src/utils/convertStyles.ts
sstraatemans: it didnt' but now that you mentioned it, i refactored and it looks nicer now

packages/libs/react-components/src/utils/convertStyles.ts
sstraatemans: shouldnt the initial be in the component it self already?

packages/libs/react-components/tsconfig.json
sstraatemans: I have no idea...I was desperately trying to get the build to run

packages/libs/react-components/src/utils/convertStyles.ts
sstraatemans: not needed anymore

packages/libs/react-components/src/components/Grid/GridContainer.tsx
sstraatemans: not needed anymore

packages/libs/react-components/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,47 @@
+import { IThemeColors, styled } from './../../styles';
+
+import React, { FC, ReactNode } from 'react';
+
+const BaseItem = styled('div', {
+  gridColumnStart: 'auto',
+});
+
+export type IColSpan = number | 'auto';
```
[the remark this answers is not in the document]
ash-vd: I think we should only prepend an I when it's an actual interface. If one wants to extend it because "it's an interface" it doesn't work. We could do `ColSpanType` if you're not happy with just `ColSpan` (but I think `ColSpan` is fine

packages/libs/react-components/src/components/Grid/GridItem.tsx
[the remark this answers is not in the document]
ash-vd: This is a tricky one. Is the general idea that we only allow the user to use colors from the theme? In my opinion it wouldn't hurt to be flexible here and allow all colors to be used.

packages/libs/react-components/src/components/Grid/GridContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+  spacing?: Stitches.VariantProps<typeof BasicContainer>['spacing'];
+  templateRows?: Stitches.CSSProperties['gridTemplateRows'];
+  templateColumns?: Stitches.CSSProperties['gridTemplateColumns'];
+  templateAreas?: Stitches.CSSProperties['gridTemplateAreas'];
+}
+
+const BasicContainer = styled('div', {
+  display: 'grid',
+  gridTemplateColumns: 'repeat(12, 1fr)',
+  gridGap: '$md',
```
eileenmguo, line 17: This is a matter of preference, but I think it's a bit clearer to define the default gridGap using the default variant

packages/libs/react-components/src/components/Grid/GridItem.tsx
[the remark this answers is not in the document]
eileenmguo: Yeah, I also think it's better to only prefix it if it is an Interface as it indicates that it is extendable. My preference is to go with `ColSpan`, but `ColSpanType` is also fine

packages/libs/react-components/src/components/Grid/GridItem.tsx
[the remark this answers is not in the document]
eileenmguo: Yeah, this one I'm a bit torn on, but I'm leaning towards providing the flexibility. I think it's clear that we should only be using colors from the theme, but I also feel that we can trust our team to use this correctly. I'm fine with either approach at this point

packages/libs/react-components/src/styles/index.ts
```diff
@@ -7,6 +7,8 @@ export {
   getCssText,
   createTheme,
   globalCss,
+  sizes,
```
eileenmguo, line 10: Why do we need to export sizes?

packages/libs/react-components/tsconfig.json
```diff
@@ -2,20 +2,7 @@
...
-    "jsx": "react",
-    "module": "ESNext",
-    "declaration": true,
-    "declarationDir": "types",
-    "sourceMap": true,
-    "outDir": "dist/esm",
-    "allowJs": true,
-    "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true,
-    "forceConsistentCasingInFileNames": true
```
eileenmguo, line 18: I would revert this change since this is this configuration is what transpiles the esm version of the lib

eileenmguo (changes requested):

packages/libs/react-components/src/styles/index.ts
sstraatemans: removed

packages/libs/react-components/src/components/Grid/GridContainer.tsx
```diff
@@ -0,0 +1,70 @@
+import { styled } from './../../styles';
+
+import type * as Stitches from '@stitches/react';
```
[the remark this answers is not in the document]
sstraatemans: done

packages/libs/react-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+const meta: Meta<typeof Grid.Container> = {
+  title: 'Grid',
+  component: Grid.Container,
+  argTypes: {
+    spacing: {
+      options: Object.keys(theme.sizes).map((key) => key),
```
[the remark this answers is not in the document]
eileenmguo: Ohh actually I think this should be taken from the variants since we are restricting it only be able to use 2xs, xs, sm...?

packages/libs/react-components/src/components/Grid/GridItem.tsx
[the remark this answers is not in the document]
sstraatemans: I like the ColsSpanType

packages/libs/react-components/src/components/Grid/GridItem.tsx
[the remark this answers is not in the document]
sstraatemans: fixed this

packages/libs/react-components/src/components/Grid/Grid.stories.tsx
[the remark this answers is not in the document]
eileenmguo: > Ohh actually I think this should be taken from the variants since we are restricting it only be able to use 2xs, xs, sm...?

👀

packages/libs/react-components/src/components/Grid/GridContainer.tsx
eileenmguo: You should be able to remove `gridGap: "$md"` now that the default variant is defined

packages/libs/react-components/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+export type ColSpanType = number | 'auto';
+
+export interface IGridItemProps {
+  children?: ReactNode;
+  colStart?: ColSpanType;
+  colEnd?: ColSpanType;
+  rowStart?: ColSpanType;
+  rowEnd?: ColSpanType;
+  area?: CSSProperties['gridArea'];
+  bg: PropertyValue<'backgroundColor', typeof config>;
```
eileenmguo, line 21: 🥳

packages/libs/react-components/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+  rowStart,
+  rowEnd,
+  area,
+  bg,
+}) => {
+  const styles = {
+    gridColumn: createSpan(colStart, colEnd),
+    gridRow: createSpan(rowStart, rowEnd),
+    gridArea: area,
+    backgroundColor: bg,
```
eileenmguo, line 45: maybe it would be good to add this util to stitches so we can always use `bg` to define backgroundColor? What do you think?

packages/libs/react-components/src/styles/colors/index.ts
```diff
@@ -1,13 +1,14 @@
 // TODO: Update to match design system
-export const colors: Record<string, string> = {
+// eslint-disable-next-line @kadena-dev/typedef-var
+export const colors = {
```
eileenmguo, line 3: Should we add the type back so we can remove the eslint disable?

packages/libs/react-components/src/styles/stitches.config.ts
```diff
@@ -4,8 +4,10 @@ import { colors, colorsDark } from './colors';
 
 import type { CSS as StitchesCSS } from '@stitches/react';
 import { createStitches, PropertyValue } from '@stitches/react';
+import { ConfigType } from '@stitches/react/types/config';
 
-const sizes: Record<string, string> = {
+// eslint-disable-next-line @kadena-dev/typedef-var
```
eileenmguo, line 9: Should we add the type back here too? I think it would be better to use eslint-disable as little as possible

packages/libs/react-components/src/styles/stitches.config.ts
```diff
@@ -169,4 +173,6 @@ export const darkTheme = createTheme('darkTheme', {
   },
 });
 
+export type IThemeColors = `$${keyof typeof theme.colors}`;
```
eileenmguo, line 176: Would it be worth using ScaleValue here instead of composing it ourselves?

packages/libs/react-components/tsconfig.test.json
```diff
@@ -0,0 +1,21 @@
+{
```
eileenmguo, line 1: nice! glad this works! 

Some things that still need to change: 
- I think it would be better to name this tsconfig.esm.json since this transpiles the lib for esm
- We need to update the package.json `build` script from 
```"build": "tsc & tsc -p ./tsconfig.cjs.json",```
to
```"build": "tsc -p ./tsconfig.esm.json & tsc -p ./tsconfig.cjs.json",```

eileenmguo: Also the `tsconfig.cjs.json` will need to extend the `tsconfig.esm.json` since it was previously extending that tsconfig and overwriting certain values

packages/libs/react-components/src/styles/stitches.config.ts
sstraatemans: for now we dont need this import anymore.
will remove it

packages/libs/react-components/src/styles/stitches.config.ts
sstraatemans: yes

packages/libs/react-components/tsconfig.test.json
sstraatemans: done

packages/libs/react-components/src/components/Grid/GridItem.tsx
sstraatemans: We need to config inside the config, which is not possible it seems.
lets look at this later

isaozler (approved):

packages/libs/react-components/src/styles/colors/index.ts
eileenmguo: can we remove the eslint-disable?

packages/libs/react-components/src/components/Grid/GridContainer.tsx
eileenmguo: not sure if you saw this since the original comment was resolved

eileenmguo (approved):
🎉

PR #200, opened 2023-03-31
feat(@kadena/ui): Add Typography components

eileenmguo:
## Reason:

Added the basic typography components based on the typography file in the design system:
- Heading
- Text

All typography components are composed from the `BaseText` styled component which defines any shared styling. Each component accepts an `as` prop and `variant` prop which allows the user to decide to select a style that does not match the semantic html element. These are the flexible text building blocks for the design system, but preset sizes that are used in applications will be added in a future PR.

Additionally, I changed the name of the `variants` files to `styles` so they can also hold styled components, not just variants.  Decided to move all styled components to the `styles` file per Steven's suggestion as we don't have to import variants back into the component's file

NOTE: Fonts are not ready to be used so those will be added in a separate PR in the future as well as additional props to the text components. Right now this gives the user very little flexibility to style the text components so I'm thinking we add to `BaseText` on an as needed basis

## Check off the following:

- [ ] I have added tutorials.


packages/libs/react-components/src/styles/stitches.config.ts
isaozler, line 1: @eileenmguo for the mono typeface the current weights are, normal=400, medium=500, semiBold=600, bold=700

packages/libs/react-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import * as variants from './variants';
+import { spacingVariant } from './styles';
```
sstraatemans, line 1: better.. nice!

packages/libs/react-components/src/styles/stitches.config.ts
eileenmguo: Oh I didn't see that in the typography page in figma and I just realized I forgot to apply the specific fontWeights 🫠. Should they match the fontWeights of the main typeface? Because then there is 300 for the non bold Heading (h1, h2)

packages/libs/react-components/src/components/Typography/Heading.tsx
```diff
@@ -0,0 +1,24 @@
+import { BaseText } from './styles';
+
+import type { VariantProps } from '@stitches/react';
+import React, { FC } from 'react';
+
+export interface IHeadingProps {
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  font?: VariantProps<typeof BaseText>['font'];
+  bold?: VariantProps<typeof BaseText>['bold'];
```
sstraatemans, line 9: I think I would prefer a variantname instead of having to set font and bold props.
So all the variants in the styleguide should have a name. and we use those names as a variant.

If we keep it like this and the style of a variant in the stylebook changes, we need to go through the whole codebase to change a prop, no?

packages/libs/react-components/src/components/Typography/Text.tsx
```diff
@@ -0,0 +1,20 @@
+import { BaseText, Text as StyledText } from './styles';
+
+import type { VariantProps } from '@stitches/react';
+import React, { FC } from 'react';
+
+export interface ITextProps {
+  as?: 'span' | 'p' | 'code';
+  font?: VariantProps<typeof BaseText>['font'];
+  bold?: VariantProps<typeof BaseText>['bold'];
```
sstraatemans, line 9: same remark as with the header

packages/libs/react-components/src/components/Stack/Stack.stories.tsx
```diff
@@ -19,34 +25,30 @@ const meta: Meta<typeof Stack> = {
...
       control: { type: 'select' },
     },
     justifyContent: {
       options: Object.keys(
-        variants.justifyContent,
-      ) as (keyof typeof variants.justifyContent)[],
-      control: { type: 'select' },
+        justifyContentVariant,
+      ) as (keyof typeof justifyContentVariant)[],
+      control: { type: 'radio' },
```
sstraatemans, line 35: 🔥 Like the radio control

packages/libs/react-components/src/components/Typography/Heading.tsx
sstraatemans: I would keep the `as` prop.
That seperates the styling from the html

Randynamic:
@all-contributors please add @eileenmguo for Code

packages/libs/react-components/src/components/Typography/Heading.tsx
eileenmguo: @Isa is working on compiling a list of "preset" variants of typography that are being used on the docs website that will be created using these base components. I am using the as prop to style the component right now. Is there a reason you want to separate the styling from the html?

eileenmguo: mmm I think I get it. Sometimes you want the semantic element to be different from the style. I'll add an additional prop `variant` in addition to the `as` prop

isaozler (approved):

PR #205, opened 2023-04-03
chore(@kadena/ui): Refactor variants

eileenmguo:
## Reason:

Refactored variants by moving them to a `variants` file so that we can keep the component file less cluttered and it makes it clear and easier to define options in storybook.

What do you think about this approach?

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/common/changes/@kadena/react-components/chore-improve-storybook-variants_2023-04-03-14-02.json
+++ b/common/changes/@kadena/react-components/chore-improve-storybook-variants_2023-04-03-14-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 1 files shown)


ash-vd (approved):

packages/libs/react-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -31,7 +33,7 @@ export const Primary: Story = {
   },
   render: ({ spacing }) => (
     <>
-      <Grid.Container spacing="2xl">
+      <Grid.Container spacing={spacing}>
```
sstraatemans, line 36: good catch

PR #229, opened 2023-04-14
clean up typography and icons

sstraatemans:
Some small changes in the typography, icons and iconbutton components

## Check off the following:

- [ ] I have added tutorials.


isaozler (approved):

packages/libs/react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -29,13 +32,19 @@ const Button = styled('button', {
 export const IconButton: FC<IIConButtonProps> = ({
   icon,
   onClick,
+  color = 'transparent',
```
eileenmguo, line 35: I believe this does not work with the type since "transparent" is not a key of colors. As we discussed earlier, we will need to come to a group consensus on how strict we want our components, but for the sake of consistency I would suggest changing the type to `PropertyValue` like we currently have in GridItem 
```export interface IGridItemProps {
  children?: ReactNode;
  colStart?: ColSpanType;
  colEnd?: ColSpanType;
  rowStart?: ColSpanType;
  rowEnd?: ColSpanType;
  area?: CSSProperties['gridArea'];
  bg?: PropertyValue<'backgroundColor', typeof config>;
}

File: packages/libs/react-components/src/components/Typography/styles.ts [hunk 2]
```diff
@@ -131,7 +131,6 @@ export const transformVariant = {
 
 export const BaseText = styled('span', {
   fontWeight: '$regular',
-  color: '$foreground',
```
eileenmguo, line 134: What is the reason for removing this? I'm not sure if the theme switching will work in storybook without this

packages/libs/react-components/src/styles/stitches.config.ts
```diff
@@ -119,6 +119,10 @@ export const {
     sizes: {
       ...sizes,
     },
+    zIndices: {
+      menu: 100,
```
eileenmguo, line 123: nitpick: might be good to make this a bit more specific like `navigationMenu`

sstraatemans: makes sense

packages/libs/react-components/src/components/Typography/styles.ts
sstraatemans: doing it like this will have the color always the same color.
removing it, it will inherit the font color of the parent element.
This will still work fine with theme switching.

We can do `color: "inherit"`
maybe that is more clear?

packages/libs/react-components/src/components/IconButton/IconButton.tsx
sstraatemans: yeah you are right...sorry. consistency is key here

sstraatemans: will change it with the icons now as well then, because I copied it from there :)

packages/libs/react-components/src/components/Typography/styles.ts
eileenmguo: Ahhh maybe if you apply the color in global styles it would work. I'm fine with it being removed as well, I would just suggest to check that this still works in storybook in dark mode 👍

packages/libs/react-components/src/components/IconButton/IconButton.tsx
sstraatemans: even better.. the `color` is already an attribute in `React.HTMLAttributes<HTMLButtonElement>`, which we extend. problem solved

packages/libs/react-components/src/components/IconButton/IconButton.stories.tsx
```diff
@@ -28,9 +26,8 @@ const meta: Meta<
       },
     },
     color: {
-      options: Object.keys(colors).map((k) => `$${k}`),
       control: {
-        type: 'select',
+        type: 'color',
```
eileenmguo, line 30: 😱

packages/libs/react-components/src/components/IconButton/IconButton.tsx
eileenmguo: Yes, but I think it inherits the type without our stitches tokens. If that's the case, I think defining it is necessary to override that type with a type that includes the tokens

packages/libs/react-components/src/components/Typography/styles.ts
sstraatemans: I did.... and it does 👍🏻

packages/libs/react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -1,20 +1,24 @@
...
 import React, { FC } from 'react';
 
 export interface IIConButtonProps
   extends React.HTMLAttributes<HTMLButtonElement> {
   icon: typeof SystemIcons[keyof typeof SystemIcons];
   onClick: React.MouseEventHandler<HTMLButtonElement>;
   title: string;
+  color: React.HTMLAttributes<HTMLButtonElement>['color'] &
+    PropertyValue<'backgroundColor', typeof config>;
 }
```
eileenmguo, line 14: Do you think this would work?

```suggestion
export interface IIConButtonProps
  extends Omit<React.HTMLAttributes<HTMLButtonElement>, "color"> {
  icon: typeof SystemIcons[keyof typeof SystemIcons];
  onClick: React.MouseEventHandler<HTMLButtonElement>;
  title: string;
  color: PropertyValue<'backgroundColor', typeof config>;
}
```

eileenmguo (approved):
Nice!

PR #232, opened 2023-04-18
feat(@kadena/ui): Add Button Component

eileenmguo:
Created a Button component that accepts children and an icon which is aligned to the right 
- Currently it only has one style variant which is `primaryFilled`, but it is setup so that all of the colors can be updated via named local tokens
- The button component can also be used as a link when passed `as='a'` and an href.
- I also added another theme dependant color called `primaryHighContrast` for the interactive states for the button. We can add these to the other colors when they are chosen
- Additionally I added a `neutral` color scale that only includes black and white as place holders for now, but it acts as a placeholder for the full neutral color scale when it is available which allows us to use a token to choose the same color regardless of the active theme

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/libs/react-components/src/components/IconButton/IconButton.tsx
+++ b/packages/libs/react-components/src/components/IconButton/IconButton.tsx
@@ -6,8 +6,10 @@ import React, { FC } from 'react';
 
 export interface IIconButtonProps
   extends Omit<React.HTMLAttributes<HTMLButtonElement>, 'color'> {
+  as?: 'button' | 'a';
   icon: typeof SystemIcons[keyof typeof SystemIcons];
-  onClick: React.MouseEventHandler<HTMLButtonElement>;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  href?: string;
   title: string;
   color?: PropertyValue<'backgroundColor', typeof config>;
 }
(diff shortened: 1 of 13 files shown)


packages/libs/react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -29,7 +29,7 @@ const Button = styled('button', {
   },
 });
 
-export const IconButton: FC<IIConButtonProps> = ({
+export const IconButton: FC<IIconButtonProps> = ({
```
sstraatemans, line 32: 🔥 nice catch

packages/libs/react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -4,7 +4,7 @@ import { SystemIcons } from '../Icons';
 import { PropertyValue } from '@stitches/react';
 import React, { FC } from 'react';
 
-export interface IIConButtonProps
+export interface IIconButtonProps
```
sstraatemans, line 7: can we add the `as` and `href` to the iconButton as well?

sstraatemans (approved):

isaozler (approved):

ash-vd (approved):

PR #234, opened 2023-04-20
create menu script

sstraatemans:
This PR has a script that walks through all the MD pages and creates a json with the page structure.
This can be used for the menu structure.

This file is imported in the _app serverside.
and send as props to the menus

## Reason:
The reason is that we don't want to maintain the menu structure by hand.
Because lots of pages will be added.

## Check off the following:

- [ ] I have added tutorials.


isaozler (approved):

packages/apps/kadena-docs/src/components/Layout/components/SideMenu/SideMenu.tsx
```diff
@@ -11,23 +11,26 @@ import {
...
     clickMenu,
     clickSubMenu,
-    hasSubmenu,
     setActive,
-    activeTitle,
-  } = useSideMenu(closeMenu);
+    activeItem,
+  } = useSideMenu(closeMenu, menuItems);
+
+  const hasSubmenu = activeItem?.children.length ?? 0;
```
ash-vd, line 33: Can we make this a boolean and ditch the `> 0` below? Or do we need the length for something else? Then it might be better to rename it to `submenuLength` or something similar?

ash-vd (approved):

PR #236, opened 2023-04-21
chore(@kadena/ui): Update component library to align with guidelines written in the README

eileenmguo:
## Reason:

Due to our discussion as a team, we came up with some guidelines moving forward with the component library. The readme was already updated, but this PR applies those changes to the components themselves. The following changes were made:

- Updated color props to only expose color combinations discussed with Isa (left out options that aren't used in the designs yet like `warning` for IconButtons - I think it's better to add them as we need them)
- Removed color props from components that don't need them 
- Updated color sets
- Some cleanup and restructuring

## Check off the following:

- [ ] I have added tutorials.


isaozler (approved):

packages/libs/react-components/src/components/IconButton/styles.ts
```diff
@@ -0,0 +1,53 @@
...
+    background: '$primarySurface',
+    $$svgColor: '$colors$primaryHighContrast',
+  },
+  positive: {
+    background: '$positiveSurface',
+    $$svgColor: '$colors$positiveHighContrast',
+  },
+  negative: {
+    background: '$negativeSurface',
+    $$svgColor: '$colors$negativeHighContrast',
```
sstraatemans, line 20: missing the gray background, used in the Header of the app

isaozler: That's indeed as we have discussed 👍

isaozler (approved):

PR #238, opened 2023-04-24
Ignore unnecessary files when bundling packages

### LINKED ISSUE #183: [@kadena/chainweb-node-client] published packages is 53mb
Fix so that this package isn't so big when publishing. It now includes the following directories which can be excluded

- .heft/build-cache
- .rush/temp
- temp

lukeribchester:
## Reason:

TBD

Closes #183 

- Add an entry to `.gitignore` to ignore IntelliJ Platform files.
- Add an `.npmignore` file to each of the following packages:
  - `libs/`:
    - `bootstrap-lib`
    - 
  - `tools/`:
    - `pactjs-cli`

## Check off the following:

- [ ] I have added tutorials.


ash-vd (approved):

packages/libs/bootstrap-lib/.npmignore
```diff
@@ -0,0 +1,13 @@
+# Ignore source, test, and Storybook files.
+src/
+lib/test*/
+*.test.*
+*.stories.*
```
Randynamic, line 5: @ash-vd do we have storybook in here?

packages/libs/client/.npmignore
```diff
@@ -1,34 +1,13 @@
...
-#
-# package.json
-# README (and its variants)
-# CHANGELOG (and its variants)
-# LICENSE / LICENCE
-
-#--------------------------------------------
-# DO NOT MODIFY THE TEMPLATE ABOVE THIS LINE
-#--------------------------------------------
-
```
Randynamic, line 29: I guess we need to put some of this back

packages/libs/bootstrap-lib/.npmignore
ash-vd: We don't, and although I don't think we will ever do in a lib, it doesn't hurt to ignore them anyway. On the other hand it would be cleaner to only ignore what's actually there... 

Now that I think of it I like the clean approach more, so only ignore files that make sense for the specific package.

packages/libs/client/.npmignore
lukeribchester: It depends on the chosen philosophy for removing unnecessary files (i.e. whitelist or blacklist).

**For this implementation nothing from the template is required.**

I noticed the few existing `.npmignore` files after finalising my blacklist approach — which I feel is more natural. A whitelist approach is more explicit and is likely to lead to future confusion about why certain files aren't published. The documentation in this template describes how certain files such as "CHANGELOG (and its variants)" are automatically included, however I found this to be incorrect when testing.

alber70g (approved):

ash-vd:
Thanks for this @lukeribchester! Unfortunately the build fails, could you rebase on master to see if that fixes it?

.gitignore
```diff
@@ -77,6 +77,9 @@ jspm_packages/
 # OS X temporary files
 .DS_Store
 
+# IntelliJ Platform files
+.idea/
+
```
isaozler, line 82: Maybe we can also add .vscode

isaozler (approved):

.gitignore
alber70g: Not always, we're using recommended extensions there and some settings specific to certain extensions (graphql)

PR #240, opened 2023-04-25
fix(@kadena/react-components): adding skipLibCheck to fix incorrect @types/react version resolution

Takadenoshi:
Attempt to fix build issue with react-components

Issue has to do with @types/react confusion - v18 used in react-components but v17 is resolved by typescript somehow

~removing skipLibCheck fixed this locally~ edit: incorrect

### DIFF
--- a/common/changes/@kadena/react-components/fix-react-components-build_2023-04-25-17-17.json
+++ b/common/changes/@kadena/react-components/fix-react-components-build_2023-04-25-17-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


Takadenoshi:
Build succeeded but testing fails - both 17 and 18 @types/react versions are used somehow. Will keep at it.

Takadenoshi:
Still not sure how the v17 types are finding their way into this sub project but explicitly setting skipLibCheck=true on all tsconfigs seems to fix both building and testing. We will revisit with this Albert.

EnoF (approved):

Randynamic (approved):

PR #243, opened 2023-04-26
Bugfix icon (re)sizing

isaozler:
## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/libs/react-components/src/components/Icons/styles.ts
+++ b/packages/libs/react-components/src/components/Icons/styles.ts
@@ -20,6 +20,10 @@ export const sizeVariant = {
 export const IconContainer = styled('span', {
   display: 'flex',
 
+  '> svg': {
+    width: '100%',
+  },
+
   variants: {
     size: sizeVariant,
   },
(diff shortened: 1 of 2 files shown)


packages/libs/react-components/src/components/Icons/styles.ts
```diff
@@ -20,6 +20,10 @@ export const sizeVariant = {
 export const IconContainer = styled('span', {
   display: 'flex',
 
+  '> svg': {
+    width: '100%',
+  },
```
eileenmguo, line 25: I'm a bit confused by this. Can you elaborate on the issue?

EnoF (approved):

packages/libs/react-components/src/components/Icons/styles.ts
eileenmguo: oh i see 👍

eileenmguo (approved):

packages/libs/react-components/src/components/Icons/styles.ts
isaozler: so if the icon is initially at size small and you resize the grid or icon to large the icon doesn't resize to the larger size

this because the svg has a maxWidth but not a width (which resizes the icon)

PR #245, opened 2023-04-27
Layout: Card

bearni95:
Card element as described [in Figma](https://www.figma.com/file/SabswJFAG1Y595zDWQVKBk/KadenaDocs?type=design&node-id=763-2657&t=a6R88XcMFGvcjuQY-0):

**Options**
- **Expand (boolean)**:  Makes the `width` and `height` be `min-content` when disabled, or `100%` when enabled
- **Color (one of):** Sets the color for the background of the card. Possible values are:
  - `default`
  - `primary`
  - `secondary`
  - `positive`
  - `warning`
  - `negative`


packages/libs/react-components/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+type ColorKeys = keyof typeof colors;
+
+const meta: Meta<
+  {
+    selectIcon: keyof typeof SystemIcons;
+    text: string;
+  } & ICardProps
```
eileenmguo, line 16: I don't believe you need these additional card props

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  minWidth: '546px',
+  width: 'min-content',
+  height: 'min-content',
+  boxShadow: '0px 0px 10px rgba(0, 0, 0, 0.04)',
+  borderRadius: '8px',
+  variants: {
+    color: {
+      default: {
+        backgroundColor: '$colors$defaultSurface',
+        color: '$colors$defaultContrast',
```
eileenmguo, line 30: You can access the color tokens like this `$defaultSurface` for these properties since stitches already knows that you will be using color tokens for backgroundColor and color

Here are their mappings: https://stitches.dev/docs/tokens#property-mapping

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  backgroundColor: '$colors$neutral1',
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'flex-start',
+  padding: '24px',
+  minWidth: '546px',
+  width: 'min-content',
+  height: 'min-content',
+  boxShadow: '0px 0px 10px rgba(0, 0, 0, 0.04)',
+  borderRadius: '8px',
```
eileenmguo, line 25: We generally want to avoid unique values to certain properties because we want all of these to be consistent and align with the design system. To achieve this, we add values like border radius to the theme so we can reuse the same values via a token. If this value isn't already in the theme, you should verify with @isaozler if it was intentional to use a new style, or if it needs to be integrated into the theme for reuse

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+];
+
+export const StyledCard = styled('div', {
+  boxSizing: 'border-box',
+  backgroundColor: '$colors$neutral1',
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'flex-start',
+  padding: '24px',
+  minWidth: '546px',
```
eileenmguo, line 21: these should use space/size tokens. For padding you may even want to use one of the breakpoint defined spaces (xs, sm, .etc)

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  'default',
+  'primary',
+  'secondary',
+  'positive',
+  'warning',
+  'negative',
+];
+
+export const StyledCard = styled('div', {
+  boxSizing: 'border-box',
```
eileenmguo, line 15: is this necessary? If so, we might want to move it to global styles

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+});
+
+export const StyledCardHeader = styled('div', {
+  width: '100%',
+  fontFamily: '$main',
+  fontStyle: 'normal',
+  fontWeight: '700',
+  fontSize: '20px',
+});
```
eileenmguo, line 66: You shouldn't need this as we already have our typography component `Header` which should be used for all headers

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+  flexDirection: 'column',
+  alignItems: 'flex-start',
+  padding: '24px',
+  minWidth: '546px',
+  width: 'min-content',
+  height: 'min-content',
+  boxShadow: '0px 0px 10px rgba(0, 0, 0, 0.04)',
+  borderRadius: '8px',
+  variants: {
+    color: {
```
eileenmguo, line 27: @isaozler Should the card component have so many color options? I would expect it to just be gray

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,84 @@
...
+        color: '$colors$neutral5',
+      },
+      negative: {
+        backgroundColor: '$colors$negativeSurface',
+        color: '$colors$neutral5',
+      },
+    },
+    expand: {
+      true: { width: '100%' },
+      height: { width: '100%' },
```
eileenmguo, line 55: I'm a bit confused by these variants, did you only mean to have true as an option?

eileenmguo (commented):
Good progress! Notes are mainly about using theme tokens and some more general usage questions

packages/libs/react-components/src/components/Card/styles.ts
bearni95: Yes, sorry! I've corrected it in the last commit, the idea is that the `true` variant sets both width and height to 100%

packages/libs/react-components/src/components/Card/styles.ts
isaozler: We have 2 colors for card surfaces, white and gray indeed

isaozler (approved):

packages/libs/react-components/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Card, ICardProps } from '.';
+import { CardBody, CardColors, CardFooter, CardHeading } from './';
+
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+type ColorType = 'default' | 'accent';
+const meta: Meta<
+  {
+    color: ColorType;
```
eileenmguo, line 11: Do we need to define a new ColorType or can we just use CardColors?

packages/libs/react-components/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const Card: FC<ICardProps> = ({
+  children,
+  expand = false,
+  color = CardColors.Default,
+}) => {
+  return (
+    // eslint-disable-next-line
+    <StyledCard color={color} expand={expand as any}>
```
eileenmguo, line 30: you should use `Stitches.VariantProps` for expand as well so you don't have to cast the type 👍

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const StyledCard = styled('div', {
+  backgroundColor: '$neutral1',
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'flex-start',
+  padding: '$lg',
+  width: 'max-content',
+  height: 'max-content',
+  boxShadow: '$shadows$card',
```
eileenmguo, line 18: I don't think you need to prefix with with shadows

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,53 @@
...
+  borderRadius: '$radii$lg',
+  variants: {
+    color: {
+      default: {
+        backgroundColor: '$defaultSurface',
+        color: '$defaultContrast',
+      },
+      accent: {
+        backgroundColor: '$colors$neutral2',
```
eileenmguo, line 27: You shouldn't need to prefix this with colors

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,53 @@
...
+        color: '$contrast',
```
eileenmguo, line 28: I don't believe this is a color defined in the theme?

packages/libs/react-components/src/styles/stitches.config.ts
```diff
@@ -120,6 +126,7 @@ export const {
     shadows: {
       // TODO: Update to match design system
       1: `0px 1px 2px 0 $colors$neutral3`,
+      card: `0px 0px 10px rgba(${hexToRgb(colors.neutral1)}, .04)`,
```
eileenmguo, line 129: Generally speaking, we want to avoid using alpha colors unless they are really necessary. Would it be possible to achieve a similar result using a color without alpha?

eileenmguo (changes requested):
Looking good so far! Just some minor comments. 

Also, in the future I think it's preferable to just update the changelog message to reflect the contents of the PR rather than creating new one for every change that happens during development. The reason is because we squash all of the commits when merging anyways so it's a bit cleaner to just have one changelog message that shows up in release notes.

packages/libs/react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -8,7 +8,7 @@ import React, { FC } from 'react';
 export interface IIconButtonProps
   extends Omit<React.HTMLAttributes<HTMLButtonElement>, 'color'> {
   as?: 'button' | 'a';
-  icon: typeof SystemIcons[keyof typeof SystemIcons];
+  icon: (typeof SystemIcons)[keyof typeof SystemIcons];
```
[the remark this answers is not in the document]
bearni95: Nopes, ran prettier on the wrong path! Undone on the latest commit

packages/libs/react-components/src/components/Card/styles.ts
bearni95: I've changed it into "Negative", but don't really see a color for that meaning in any other components

packages/libs/react-components/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { CardColor } from './styles';
+import { Card, ICardProps } from '.';
+import { CardBody, CardColors, CardFooter, CardHeading } from './';
+
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+const meta: Meta<
+  {
+    color: CardColor;
```
eileenmguo, line 11: I believe color is already part of ICardProps

packages/libs/react-components/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+    expand: boolean;
+  } & ICardProps
+> = {
+  title: 'Layout/Card',
+  argTypes: {
+    color: {
+      options: Object.values(CardColors),
```
eileenmguo, line 18: I would expect this to get the options from the variant object like the other stories do rather than a CardColors type. Is there a reason why you decided to do it this way?

packages/libs/react-components/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { StyledCard } from './styles';
+
+import { VariantProps } from '@stitches/react';
+import React, { FC } from 'react';
+
+export interface ICardProps {
+  children: React.ReactNode;
+  color?: VariantProps<typeof StyledCard>['color'];
+  expand?: boolean;
```
eileenmguo, line 11: Since expand is a Variant, I would recommend using the VariantProp typing for the sake of consistency

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+  flexDirection: 'row',
+  alignItems: 'flex-start',
+  paddingTop: '$lg',
+  gap: '$sm',
+  width: '100%',
+});
+
+export type CardColor = VariantProps<typeof StyledCard>['color'];
+
+export const CardColors: { [key: string]: CardColor } = {
```
eileenmguo, line 57: You can look at other components as examples, but we have been separating the variants into separate objects so that there is one source of truth that is used for the variants and is exported and used in stories

packages/libs/react-components/src/components/Card/styles.ts
eileenmguo: I didn't mean the name of the variant! I meant "$contrast" is not a defined color in the theme. If you go to `styles/colors` you can see all of the available color tokens

packages/libs/react-components/src/components/Card/Card.tsx
eileenmguo: 👀

packages/libs/react-components/src/components/Card/styles.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint @kadena-dev/typedef-var: 0 */
+
+import { styled } from '../../styles';
+
+export const colorVariant = {
```
isaozler, line 5: I would suggest to use theme: dark | light instead of default and negative as variants to keep it consistent

alber70g (approved):
I see you added multiple `rush change` documents. There's no need to do that. It should represent the change that will end up in the changelog when the package is released.

So just having a change with "adds Layout: Card" is sufficient

packages/libs/react-components/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { StyledCard } from './styles';
+
+import { VariantProps } from '@stitches/react';
+import React, { FC } from 'react';
+
+export interface ICardProps {
+  children: React.ReactNode;
+  fullWidth?: VariantProps<typeof StyledCard>['fullWidth'];
+  stack?: boolean;
```
eileenmguo, line 11: nit: This should never be anything outside of true or false, but since this is tied to a variant, it's generally best to make the type also tied to the variant as well. So I would type it like fullWidth

packages/libs/react-components/src/components/Input/Input.tsx
```diff
@@ -16,8 +16,8 @@ export interface IInputProps
   > {
   as?: 'input';
   leadingText?: string;
-  leftPanel?: typeof SystemIcons[keyof typeof SystemIcons];
-  rightPanel?: typeof SystemIcons[keyof typeof SystemIcons];
+  leftPanel?: (typeof SystemIcons)[keyof typeof SystemIcons];
```
eileenmguo, line 19: Do you know what is causing this reformating?

bearni95: It's the running of prettier on my end! Will revert and keep an eye out for these

bearni95: It actually seems to be running with the prettier command upon commiting, so not sure how to undo it!

ash-vd: To me this formatting does make sense. It's a bit the same as `(a as any).item`. What is weird is that the parens weren't there before, and now Prettier made them appear. Didn't we run Prettier before for this file?

packages/libs/react-components/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+            Learn Kadena&apos;s core concepts & tools for development in 15
+            minutes
+          </CardBody>
+          <CardFooter>
+            <Button title={'Button'} icon={SystemIcons.Information}>
+              Get Started
+            </Button>
+            <Button title={'Button'}>Hello World Tutorial</Button>
+          </CardFooter>
+        </Card>
```
eileenmguo, line 53: For future implementations: Sub components should be namespaced so you can use them with the following syntax
```
<Card>
  <Card.Heading>
    ...
  </Card.Heading>
</Card>
```

You can see an example of this in the export for the Grid component!

PR #246, opened 2023-04-27
Include prettier in the react-components package

bearni95:
Added `prettier` (2.8.5) dependency and a prettier npm command (`npm run prettier`) to the package.json for libs/react-components

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -708,6 +708,7 @@ importers:
       '@types/node': ^16.0.0
       '@types/react': ^18.0.28
       eslint: ^8.15.0
+      prettier: ^2.8.5
       react: ^18.2.0
       react-dom: ^18.2.0
       storybook: ^7.0.0-rc.3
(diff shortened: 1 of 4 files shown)


common/config/rush/pnpm-lock.yaml
```diff
@@ -708,6 +708,7 @@ importers:
       '@types/node': ^16.0.0
       '@types/react': ^18.0.28
       eslint: ^8.15.0
+      prettier: ^2.8.5
```
Randynamic, line 711: If I'm not mistaken we already have a global prettier config. We do need to update that one to the latest version.

eileenmguo: That was my understanding as well, but I think @alber70g knows best here. I think if the goal is to just fix formatting you can also run the `lint` script

bearni95: Prettier runs upon committing with the `lint-staged` command, but I've found no way to run something like `npm run prettier` other than including it in the package

ash-vd: You can run `rush lint` to lint everything in the monorepo, or run `rush lint -i .` to lint the project in the current working directory and the projects it depends on. It's also possible to for example run `rush lint -I @kadena/client` to specify a project.

We shouldn't add prettier as a dependency for each project, but rely on the one that's defined globally :).

bearni95: You're right, there are other ways without messing with the package.json

Thanks!!

PR #226, opened 2023-04-13
Chainweb stream client (broken branch)

Takadenoshi:
WIP chainweb-stream-client branch

Still need to extend the README a bit, but the approach and code should be mostly ready for review

Backend: https://github.com/kadena-io/chainweb-sse/

Will open a PR on the account tracker to show demo usage

PR #249, opened 2023-05-01
Chainweb Stream Client (working? build branch) 

Takadenoshi:
Reopening https://github.com/kadena-community/kadena.js/pull/226 with a new branch from current master to see if it fixes @types/react incorrect resolution issues


Takadenoshi:
Fixed some eslint/typing issues and it now passes. Since last review on the other branch we have these changes:

https://github.com/kadena-community/kadena.js/pull/249/commits/4a55c121c32d06b1fe43cb5c4dfeec576290deec

@EnoF @ash-vd can you take another look?

common/changes/@kadena/chainwebjs/chainweb-stream-client-new_2023-05-01-11-17.json
EnoF, line 1: is this a duplicate from the json above?

common/changes/@kadena/chainwebjs/chainweb-stream-client-new_2023-05-01-11-16.json
EnoF, line 1: should this json be here?  or should the one above this one replace this json? (common/changes/@kadena/chainweb-stream-client/chainweb-stream-client-new_2023-05-01-11-17.json)

Randynamic (approved):

packages/libs/chainweb-stream-client/src/examples/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const client: ChainwebStreamClient = new ChainwebStreamClient({
+  // required
+  type: 'account',
+  id: 'k:e7f7130f359fb1f8c87873bf858a0e9cbc3c1059f62ae715ec72e760b055e9f3',
+  host: 'http://localhost:4000/',
+
+  // optional
+  limit: 100,
+  connectTimeout: 15_000,
```
EnoF, line 11: didn't knew this was allowed :100: this makes it so much easier to read, nice!

packages/libs/chainweb-stream-client/src/index.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    // close event source; crucial - chrome reconnects, firefox does not
+    this._eventSource?.close();
+
+    // cancel connection timeout timer, if exists
+    if (this._connectTimer) {
+      clearTimeout(this._connectTimer);
+    }
+
+    let message = 'Connection error'; // default for "Event". No error information is available, presume disconnection
```
EnoF, line 187: if you'd extract the message retrieval you could remove the `let` and then this code would be easier to read

EnoF (approved):

packages/libs/chainweb-stream-client/src/examples/index.ts
Takadenoshi: the _ separator in numbers? I love this feature for large numbers

packages/libs/chainweb-stream-client/src/index.ts
Takadenoshi: good point, I refactored the errors a bit. Please check again

common/changes/@kadena/chainwebjs/chainweb-stream-client-new_2023-05-01-11-17.json
Takadenoshi: good catch, not sure why it compained to me. I deleted one and added a description on the other. We did change the @types/eventsource to pin to 1.10.0 (as 1.10.1 broke for ts 4.x)

webpro:
Just wondering: why are we using the `eventsource` package and not the native `EventSource` in the browser?

Takadenoshi:
@webpro EventSource re: first point,, EventSource is only touched if it does not exist, so in browsers that have it defined already this will not make a difference

regarding size, I checked the size of this in a react bundle and it is 4.92KB so I do not think it is a problem.

PR #257, opened 2023-05-02
Make readme a bit more readable

webpro:
The readme was a bit hard to read. This PR turns the list into a table so it's a tad easier to grasp everything this monorepo contains. Also the reading flow was a bit weird with links back and forth ("go here" and "go back").

For now I've left the content mostly as-is, just moving stuff around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,31 +60,10 @@ Kadena.js.
 
 See CONTRIBUTING.md on how to contribute.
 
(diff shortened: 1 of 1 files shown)


ash-vd:
Way better indeed 🚀

eileenmguo (approved):
Much better!

PR #259, opened 2023-05-02
fix(pactjs-cli) : append the export to index.d.ts if it's not already there

javadkh2:
## Reason:

contract-generate adds the export type to the index file even if the export is already there.  it happens if we run the script on a file twice. 
it can happen where a user updates the contract and wants to ensure the types are also updated.

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/tools/pactjs-cli/src/contract-generate/generate.ts
+++ b/packages/tools/pactjs-cli/src/contract-generate/generate.ts
@@ -141,7 +141,10 @@ export const generate =
       if (existsSync(indexPath)) {
         console.log(`Appending to existing file ${indexPath}`);
         const indexDts: string = readFileSync(indexPath, 'utf8');
-        writeFileSync(indexPath, indexDts + exportStatement);
+        // Append the export to the file if it's not already there.
+        if (!indexDts.includes(exportStatement)) {
+          writeFileSync(indexPath, indexDts + exportStatement);
+        }
       } else {
         console.log(`Writing to new file ${indexPath}`);
         writeFileSync(indexPath, exportStatement);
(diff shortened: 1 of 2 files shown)


Randynamic (approved):

ash-vd (approved):

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -141,7 +141,10 @@ export const generate =
       if (existsSync(indexPath)) {
         console.log(`Appending to existing file ${indexPath}`);
         const indexDts: string = readFileSync(indexPath, 'utf8');
-        writeFileSync(indexPath, indexDts + exportStatement);
+        // Append the export to the file if it's not already there.
+        if (!indexDts.includes(exportStatement)) {
+          writeFileSync(indexPath, indexDts + exportStatement);
```
Takadenoshi, line 146: Should we separate these with a newline or two (if they are not there?) Maybe something like

```
const separator = indexDts.endsWith('\n') ? '' : '\n\n';
const newIndexDts = [indexDts, exportStatement].join(separator);
writeFileSync(indexPath, newIndexDts);
```

javadkh2: Agree, adding new line makes it more readable. 
maybe only adding one new line is enough. 
 ```javascript
  writeFileSync(indexPath, indexDts + exportStatement + "\n");
  ```

Takadenoshi: My intent was for the newline to separate the inserted statement from the existing body - otherwise it would end up in the same line if the body does not end in a newline

PR #262, opened 2023-05-03
Update pnpm (v7) and Node.js (>=16)

webpro:
## Reason:

When starting out I couldn't start anything because the project demands Node.js <= v16. I'm at Node.js v18 and v20 is also released. So I figured we should basically turn this around now: let's allow Node.js >= v16. Also we can start using pnpm v7 (had some issues with pnpm v8 we can look into later).

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.


webpro:
Tracking https://github.com/microsoft/rushstack/issues/4033 for pnpm v8

ash-vd (approved):

eileenmguo (approved):

javadkh2:
There are some issues that arise when using pnpm 7. 
Perhaps it would be best to only update the node version to 16 at this stage

javadkh2:
There are some issues that arise when using pnpm 7. 
Perhaps it would be best to only update the node version to 16 at this stage

webpro:
Thanks, closing this in favor of #273.

PR #263, opened 2023-05-04
[Docs] Feat/docs add frontmatter

sstraatemans:
This story will add the MD meta data as props.
so we can add title and description etc.

because this runs as a plugin on the build it is NOT typescript.
## Check off the following:

- [ ] I have added tutorials.


eileenmguo (approved):

packages/apps/kadena-docs/cypress/e2e/layoutfull.cy.ts
```diff
@@ -41,11 +41,11 @@ describe('full layout', () => {
         .should('have.length', 1);
     });
 
-    it('deeplink on the page when an option of the aside menu is clicked', () => {
+    it.only('deeplink on the page when an option of the aside menu is clicked', () => {
```
ash-vd, line 44: Is this only here on purpose?

sstraatemans: it is def. not!

ash-vd (approved):

PR #264, opened 2023-05-04
Feat/notifications component

bearni95:
## Reason:

Closes [Notification Component Asana Ticket](https://app.asana.com/0/1204270938438108/1204455019525333)

## Check off the following:

- [x] I have added tutorials.


EnoF (approved):

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+  icon,
+  title,
+  description,
+  displayCloseButton = false,
+  actionButtonLabel,
+  actionButtonOnClick,
+  displayActionButton = false,
+  color = 'default',
+  simplified = false,
+  ...props
```
eileenmguo, line 41: We want to keep this library strict, so we usually want to avoid exposing the ability to pass anything to components. What other props do we want to pass on through this spread operator?

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const Icon = icon!;
+  const isSimple =
+    simplified || (!displayActionButton && description === undefined)
+      ? true
+      : false;
```
eileenmguo, line 47: nit: You don't need to do the ternary since it just evaluates to true or false

` const isSimple = simplified || (!displayActionButton && description === undefined)` should suffice

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+  title: string;
+  description?: string;
+  displayCloseButton?: boolean;
+  expand?: boolean;
+
+  actionButtonLabel?: string;
+  actionButtonOnClick?: () => void;
+  displayActionButton?: boolean;
+
+  color?: string;
```
eileenmguo, line 26: All variants should be typed with stitches variant type: https://stitches.dev/docs/typescript#extract-variants-from-a-component

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+    simplified || (!displayActionButton && description === undefined)
+      ? true
+      : false;
+
+  return (
+    <StyledNotification
+      {...props}
+      // eslint-disable-next-line
+      color={color as any}
+      type={isSimple ? 'simple' : 'full'}
```
eileenmguo, line 54: I would suggest just exposing this variant as a prop on the notifications component and giving it a default variant when defining the styled component. This way you can still just add `type="simple"` when it's necessary and it aligns more with the existing pattern we use for variants

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+    >
+      <AbsoluteButton position="left">
+        <Icon size="md" color={color} />
+      </AbsoluteButton>
+      <StyledHeading as="h6">{title}</StyledHeading>
+      {!isSimple && <StyledText as="p">{description}</StyledText>}
```
eileenmguo, line 60: I think you also need to also check if description exists?

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+          icon={SystemIcons.TrailingIcon}
+        >
+          {actionButtonLabel}
+        </StyledButton>
+      )}
+      <AbsoluteButton position="right">
+        {displayCloseButton && !isSimple && (
+          <StyledIconButton title="bla" icon={SystemIcons.Close!} />
+        )}
+      </AbsoluteButton>
```
eileenmguo, line 74: I think it would make more sense to wrap the `AbsoluteButton` component as well, otherwise it will sometimes just be an empty div

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const isSimple =
+    simplified || (!displayActionButton && description === undefined)
+      ? true
+      : false;
+
+  return (
+    <StyledNotification
+      {...props}
+      // eslint-disable-next-line
+      color={color as any}
```
ash-vd, line 53: Can we type this prop?

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,123 @@
...
+import { Heading, Text } from '../Typography';
+
+export const NotificationColors: string[] = [
+  'default',
+  'primary',
+  'secondary',
+  'positive',
+  'warning',
+  'negative',
+];
```
eileenmguo, line 13: Is there a reason why you need this? These are already defined in the styled component so that can be used for typing

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,123 @@
...
+      },
+      negative: {
+        backgroundColor: `$negativeSurface`,
+        border: '1px solid $negativeAccent',
+        color: '$negativeAccent',
+      },
+    },
+
+    expand: {
+      true: { width: '100%', height: '100%', maxWidth: '100%' },
```
eileenmguo, line 66: Should this expand in height as well? I would expect it to always just have the height of its contents, but expand in width

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,123 @@
...
+    },
+
+    type: {
+      simple: {
+        borderRadius: '$radii$md',
+        borderLeft: '1px solid $foreground',
+      },
+      full: {},
```
eileenmguo, line 74: If you aren't intending to add any styles for the full version, you could make the variant `simplified` with a `true` case like expand

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,123 @@
...
+    },
+  },
```
eileenmguo, line 76: I would recommend defining default variants here: https://stitches.dev/docs/variants#default-variants

packages/libs/react-components/src/components/Notification/Notification.tsx
eileenmguo, line 53: If color is defined using the stitches variant prop type you shouldn't have to cast the type here

eileenmguo (changes requested):

packages/libs/react-components/src/components/Notification/Notification.tsx
bearni95: Description is part of the value assignation to `isSimple`, and it should have a value when isSimple is false

packages/libs/react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+      },
+    },
+    color: {
+      options: [
+        'default',
+        'primary',
+        'secondary',
+        'positive',
+        'warning',
+        'negative',
```
eileenmguo, line 43: I would recommend making the variant a separate object where the styled component is being created and importing it here to get the keys so that if those variants are ever changed, there is one source of truth and every dependent usage will be updated. I believe you can see an example of this in IconButton

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,73 @@
...
+  title: string;
+  description?: string;
+  displayCloseButton?: boolean;
+  expand?: boolean;
+
+  actionButtonLabel?: string;
+  actionButtonOnClick?: () => void;
+  displayActionButton?: boolean;
+
+  color?: NotificationColor;
```
eileenmguo, line 27: I suggest using the variant typing for this

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  simple?: boolean;
```
eileenmguo, line 29: Also variant typing for this

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,119 @@
...
+  alignItems: 'flex-start',
+  padding: '$md calc(4 * $sm)',
+  position: 'relative',
+
+  variants: {
+    color: {
+      default: {
+        backgroundColor: `$neutral2`,
+        border: '1px solid $neutral6',
+        color: '$defaultContrast',
```
eileenmguo, line 29: I don't believe `$defaultContrast` is a color token in the theme

packages/libs/react-components/src/components/Notification/styles.ts
eileenmguo: Is there is reason why you are not setting a default variant?

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,57 @@
...
+  color,
+  simple,
+  expand,
+  buttons,
+}) => {
+  const Icon = icon!;
+  const isSimple = simple === 'true' || description === undefined;
+  return (
+    <StyledNotification color={color} expand={expand} simple={isSimple}>
+      <AbsoluteButton position="left">
```
sstraatemans, line 43: this is not a button? can we name the component something else?

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <Icon size="md" />
```
sstraatemans, line 44: the icon does not change color when the color changes? It does in the designs

packages/libs/react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+      },
+    },
+    description: {
+      control: {
+        type: 'text',
+      },
+    },
+    simple: {
+      options: Object.keys(simpleVariant) as (keyof typeof simpleVariant)[],
+      control: {
```
sstraatemans, line 34: simple is a boolean. can we make the control also a boolean like displayCloseButton option?

packages/libs/react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <Notification
+          simple={simple}
+          icon={Icon}
+          color={color}
+          title={title}
+          description={description}
+          displayCloseButton={displayCloseButton}
+          expand={expand}
+          buttons={
+            <Button title="Action label" icon={SystemIcons.TrailingIcon}>
```
sstraatemans, line 105: the icon of the button does not change color? is this a bug in the button component? if so can you create a bug story for this in Asana?

packages/libs/react-components/src/components/Notification/styles.ts
```diff
@@ -0,0 +1,117 @@
...
+export const simpleVariant = {
+  true: {
+    borderRadius: '$md',
+    borderLeftWidth: '1px',
+  },
+  false: {},
+};
+
+export const StyledNotification = styled('div', {
+  margin: '0px $sm',
```
sstraatemans, line 54: I dont know if we want a margin in here.
I think the positioning of a libary component should be set by the parent component in the application.

If we do want margins here: then the margins do not work well at the moment. in storybook i have a larger margin on the left than on the right

alber70g (approved):

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,64 @@
...
+  title: string;
+  description?: string;
+  displayCloseButton?: boolean;
+  expand?: VariantProps<typeof StyledNotification>['expand'];
+
+  color?: VariantProps<typeof StyledNotification>['color'];
+
+  simple?: VariantProps<typeof StyledNotification>['simple'];
+
+  buttons?: React.ReactNode;
```
eileenmguo, line 26: Since this can accept any ReactNode it might be good to name it something more generic like footerContent

eileenmguo (approved):
Looks good! The only minor comment is related to the buttons prop naming

sstraatemans (approved):

packages/libs/react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    expand,
+    color,
+    simple,
+  }) => {
+    const Icon = SystemIcons[selectIcon];
+    return (
+      <>
+        <Notification
+          simple={simple}
+          icon={Icon}
```
sstraatemans, line 96: The icons do not change color and stay black


image

packages/libs/react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+          color={color}
```
sstraatemans, line 97: The font color seems off. especially for the green. is this correct?
image

sstraatemans (changes requested):
after playing around I found some issues

sstraatemans (approved):

packages/libs/react-components/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <StyledNotification color={color} expand={expand} simple={isSimple}>
+      <StyledIconContainer position="left">
+        <Icon size="md" />
+      </StyledIconContainer>
+
+      <StyledHeading as="h6">{title}</StyledHeading>
+      {!isSimple && children}
+
+      {displayCloseButton === true && !isSimple && (
```
isaozler, line 57: Not something to pick up right away, but I do want to suggest having `header` (red) and `footer` (blue) slots for these components. 

image

This way, we can inject more functionalities into the components. In this case, we could have multiple header or footer actions.

isaozler (approved):

packages/libs/react-components/src/components/Notification/Notification.tsx
bearni95: Noted! Will add to the vanilla-extract rewrite PR

PR #267, opened 2023-05-08
[Chainweb Stream Client] Validate wire protocol client-server compatibility

Takadenoshi:
Added support for the new initial event that sends server config values.

The server-side and protocol changes are described in the [server PR description]() and the [new README section of that PR](https://github.com/kadena-io/chainweb-stream-server/blob/649972adbbe286c2bf44ba8c78b277d6015d3e00/README.md#validating-serverclient-configuration-compatibility)

The proposed client implementation is to refuse to proceed for most of the incompatibility conditions by firing an `error` event, logging an error to the console and disconnecting. The conditions that error out fatally:

- mismatched network
- incompatible confirmation depth
- incompatible wire protocol version (still TODO)
- mismatched type (account/event)
- mismatched ID (module/event name or account ID)

(the final two should never happen in production)

The exception is the heartbeat timeout config parameter. If the client heartbeat timeout is set to be smaller than the server, the client will automatically adapt its heartbeat timeout to `SERVER_HEARTBEAT_TIMEOUT + 2500ms`. The reasoning behind this is that changing this configuration value to match the server setting does not violate the developer expectations in any significant way, and it is preferable to (potentially) throwing a fatal error in production.

Still a draft as it is pending the version compatibility check, but the changes that are there already are good to review.

--

Regarding the wire protocol version [`(event:initial).config.v`, currently set to `0.0.2`] what should the behavior be when the client and server have a different version? 

I propose versioning the wire protocol by [semver 2.0](https://semver.org/) (e.g. this version change that modifies the `initial` event in a way that breaks backwards compatibility would be a major version bump, but 0.x.y is exempt as unstable) 

If there is a major version incompatibility, it should error our fatally as above.

Minor version incompatibility should just warn that some features may not be supported.

~Patch version incompatibility should just warn (or do nothing.)~ Patch versions are unlikely to be used as a backwards-compatible bugfix on the protocol level does not make much sense


packages/libs/chainweb-stream-client/src/index.ts
```diff
@@ -273,8 +372,12 @@ class ChainwebStream extends EventEmitter {
     if (this.limit !== undefined) {
       urlParamArgs.push(['limit', String(this.limit)]);
     }
+    // TODO This reconnection strategy of -3 from last max height
+    // guarrantees that we will not miss events, but it also means
+    // that confirmed transactions will be emitted more than once
+    // Discussion here: https://github.com/kadena-community/kadena.js/issues/275
     if (this._lastHeight !== undefined) {
-      urlParamArgs.push(['minHeight', String(this._lastHeight)]);
+      urlParamArgs.push(['minHeight', String(this._lastHeight - 3)]);
```
Takadenoshi, line 380: This also fixes a missed events edge case case when reconnectin.

Request for comments regarding what the behaviour should be regarding duplicates https://github.com/kadena-community/kadena.js/issues/275

alber70g (approved):

Randynamic (approved):

PR #270, opened 2023-05-09
Gradient Text component added

isaozler:
## Reason:

Closes [[issue link]](https://app.asana.com/0/1204526868335507/1204501321160463/f)

## Check off the following:

- [ ] I have added tutorials.

---
  - https://app.asana.com/0/0/1204501321160463


packages/libs/react-components/src/components/Typography/GradientText.tsx
```diff
@@ -0,0 +1,12 @@
+import { GradientText as StyledText } from './styles';
+
+import React, { FC } from 'react';
+
+export interface IGradientTextProps {
+  as?: 'span';
```
sstraatemans, line 6: do we need this, if there is only one option?

packages/libs/react-components/src/components/Typography/styles.ts
```diff
@@ -187,3 +187,15 @@ export const Text = styled(BaseText, {
...
   },
 });
+
+export const GradientText = styled('span', {
+  fontWeight: '$bold',
+  backgroundColor: 'inherit',
+  backgroundImage: 'linear-gradient(50deg, #ff00e9, #00c0ff 90%)',
+  backgroundSize: '100%',
+  color: 'white',
+  '-webkit-background-clip': 'text',
```
sstraatemans, line 197: wow..really thought that this would be done by the postprocessing.
But nope. 🚒

packages/libs/react-components/src/components/Typography/GradientText.tsx
isaozler: no, indeed we can remove that 👍

packages/libs/react-components/src/components/Typography/GradientText.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { GradientText } from './GradientText';
```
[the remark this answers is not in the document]
isaozler: that would also be my preference

sstraatemans (approved):

packages/libs/react-components/src/components/Typography/GradientText/GradientText.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+import React, { FC } from 'react';
+
+export interface IGradientTextProps {
+  children: React.ReactNode;
+}
+
+export const GradientText: FC<IGradientTextProps> = ({ children }) => {
+  return <StyledText>{children}</StyledText>;
+};
```
eileenmguo, line 11: You could simplify this further by just exporting the Styled component rather than rewriting it as a function component

isaozler: I was thinking that we might add a font weight property to this component, if you guys don't think that would be necessary that we can indeed just export the styled-component?

eileenmguo: even with a fontWeight property I don't think writing it as a function component would be necessary since stitches will type the variant for you

eileenmguo (approved):

PR #272, opened 2023-05-09
Kadena Transfer Repo Init

nil-amrutlal-dept:


KristinaSpasevska:
Regarding the folder structure, would it be nicer to have the business logic inside `services` folder inside `src`, instead of having it outside `src`? So that, we would have `/src/services/modules/list-modules.ts`  kind of structure

nil-amrutlal-dept:
> Regarding the folder structure, would it be nicer to have the business logic inside `services` folder inside `src`, instead of having it outside `src`? So that, we would have `/src/services/modules/list-modules.ts` kind of structure

Agreed! Will push changes

packages/apps/kadena-transfer/.eslintrc.js
```diff
@@ -0,0 +1,78 @@
+require('@rushstack/eslint-config/patch/modern-module-resolution');
+const fs = require('fs');
+
+const prettierOptions = JSON.parse(fs.readFileSync('./.prettierrc', 'utf8'));
+
```
alber70g, line 5: Can we remove everything from here and use the default config? You can use this one as example https://github.com/kadena-community/kadena.js/blob/17792c8301c9ba2f4e436b36483a0cc90abaa3c3/packages/apps/kadena-docs/.eslintrc.cjs

packages/apps/kadena-transfer/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 80,
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "all"
+}
```
alber70g, line 8: There is a .prettierrc in the root of the project. This one's redundant

packages/apps/kadena-transfer/README.md
```diff
@@ -0,0 +1,46 @@
+# Getting Started with Create React App
+
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
alber70g, line 3: I see it's bootstrapped with Create React App. We are only using nextjs for projects to reduce maintenance burden. With nextjs we can use so called "static exports" to generate a static site. See https://nextjs.org/docs/pages/building-your-application/deploying/static-exports for more info

packages/apps/kadena-transfer/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "kadena-transfer",
```
alber70g, line 2: Can we name this project `@kadena/transfer`?

packages/apps/kadena-transfer/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "web-vitals": "^2.1.4"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "lint": "eslint .",
+    "lint:fix": "eslint --fix ."
+  },
```
alber70g, line 25: Make sure to include the `_phase:...` scripts [example](https://github.com/kadena-community/kadena.js/blob/d7ea6eb7e343c79532abe791b25ade2d530a0863/packages/apps/kadena-docs/package.json#L6-L7)

packages/apps/kadena-transfer/package.json
```diff
@@ -0,0 +1,77 @@
+{
```
alber70g, line 1: Can we add the `lint-staged`

[example](https://github.com/kadena-community/kadena.js/blob/d7ea6eb7e343c79532abe791b25ade2d530a0863/packages/apps/kadena-docs/package.json#L22-L30)

packages/apps/kadena-transfer/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "jsx": "react-jsx"
+  },
+  "exclude": [
+    "node_modules",
+    "**/*.spec.ts",
+    "**/*.spec.tsx",
+    "**/*.test.ts",
+    "**/*.test.tsx"
+  ],
+  "include": ["src"]
```
alber70g, line 28: Are these all different from the `tsconfig-base.json`? If so, why? If not, let's remove the ones that aren't part of it. We probably need something similar to [this one](https://github.com/kadena-community/kadena.js/blob/a60b34845c079dfbcc4ccd0cab63780e13369162/packages/apps/kadena-docs/tsconfig.json)

We can also create a new profile for `nextjs` in `@kadena-dev/heft-rig/profiles/nextjs`. This would benefit other apps as well

PR #274, opened 2023-05-09
[Docs] MD compoment for tables

sstraatemans:
Creating the styles for MD tables
Closes https://app.asana.com/0/1202676524474548/1204549233615428/f

image
image

## Check off the following:

- [ ] I have added tutorials.


packages/apps/kadena-docs/src/components/Layout/components/Main/Main.tsx
```diff
@@ -35,7 +35,6 @@ export const Main: FC<IProps> = ({ children, frontmatter }) => {
       if (tree.length) {
         tree.map((item) => {
           // is the menu open?
```
EnoF, line 37: I'm aware this is not really part of the PR but, this function is relying on the side effect of mutating the item... It would be better to have the item returned in the map and used actually in a separate function. i.e.

```
tree.map(item => {
  return {
    ...item,
    isMenuOpen: pathname.startsWith(item.root),
    isActive: item.root === pathname,
 }
})// now use this tree to do something
```

EnoF (approved):

PR #276, opened 2023-05-10
[Kadena Transfer] Deployment parameters

MRVDH:


MRVDH:
Closed because it should go to `kadena-transfer` instead of `master`

PR #279, opened 2023-05-11
[pactjs-test-project] Make sure the post install is run for this test project before building

EnoF:
## Reason:

Initial build after cloning fails on this project as the `postinstall` script is not run with `rush install`

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/apps/graph-client/package.json
+++ b/packages/apps/graph-client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "build": "next build",
+    "build": "rushx postinstall && next build",
     "dev": "concurrently --kill-others \"npm:next:dev\" \"npm:start:graph\" \"sleep 3; rushx generate:sdk -w\"",
     "generate:sdk": "graphql-codegen --config codegen-sdk.yml",
     "postinstall": "npm run generate:sdk",


alber70g:
I don't think running postinstall on every build is the right solution. Maybe we can look at this? https://github.com/microsoft/rushstack/issues/862#issuecomment-942918104

packages/libs/pactjs-test-project/package.json
```diff
@@ -13,7 +13,7 @@
   "scripts": {
     "_phase:build": "npm run pactjs:generate:contract:file; tsc --noEmit",
     "_phase:test": "npm run test",
-    "build": "tsc --noEmit",
+    "build": "rushx postinstall && tsc --noEmit",
```
javadkh2, line 16: I suggest replacing the postinstall script with prebuild. Actually, running `rush install` or `rush update --purge` triggers all postinstall scripts. In the case of a fresh clone, it will fail because the "pactjs" dependency is not built yet
```JSON
 "build": "npm run postinstall && tsc --noEmit",
 "prebuild": "npm run pactjs:generate:contract:file && npm run pactjs:generate:template"
 ```

Comment (EnoF) [reply packages/libs/pactjs-test-project/package.json:16]: It's a bit awkward that we need to even think about this... The steps are I believe:

1. install deps
2. execute post install script of pactjs
3. execute post install script of pactjs-test-project (and any other depending projects)

However it seems that `pact-test-project` did not have `pactjs` available yet during post install, hence this project's post install fails due to a race condition. Are we not able to help rush understand that `pactjs-test-project` depends on the `pactjs` cli dep to be installed? (tbh I was expecting rush to have solved that)

File: packages/apps/graph-client/package.json [hunk 2]
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "build": "next build",
+    "build": "rushx postinstall && next build",
```
javadkh2, line 6: Do you know if this resolved the issue? 
The post-install script in this package does not depend on any other packages from the monorepo, so it should not fail due to missing builds. We should investigate why it failed during the post-install stage.

packages/libs/pactjs-test-project/package.json
javadkh2: Since this is the only package with a dependency on another package in the monorepo during the postinstall stage, I would replace it with a rebuild

Takadenoshi: `rush build` does not run `prebuild` scripts 🤐

[Issue open since September 2018](https://github.com/microsoft/rushstack/issues/830)

Takadenoshi:
> I don't think running postinstall on every build is the right solution. Maybe we can look at this? [microsoft/rushstack#862 (comment)](https://github.com/microsoft/rushstack/issues/862#issuecomment-942918104)

I tried this in a new branch and it still didn't build

packages/apps/graph-client/package.json
alber70g: It's not possible to run another `rush(x)` command during `rush build`. This will result in the error like "another rush command already running in this repository"

MRVDH:
@EnoF @Takadenoshi This PR/branch has become obsolete due to #294 right?

EnoF:
fixed via https://github.com/kadena-community/kadena.js/pull/294

PR #282, opened 2023-05-15
pactjs-cli unit test, windows compatibility

javadkh2:
## Reason:

pactjs-cli unit test is failing on Windows. because Windows use "\" for paths.

used nodejs path util to make paths compatible by OS

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/tools/pactjs-cli/src/retrieve-contract/tests/retrieve-contract.test.ts
+++ b/packages/tools/pactjs-cli/src/retrieve-contract/tests/retrieve-contract.test.ts
@@ -1,10 +1,10 @@
 jest.mock('fs');
 jest.mock('../../utils/retrieveContractFromChain');
-const path = require('path');
 
 import { retrieveContractFromChain } from '../../utils/retrieveContractFromChain';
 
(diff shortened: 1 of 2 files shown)


ash-vd (approved):

packages/tools/pactjs-cli/src/retrieve-contract/tests/retrieve-contract.test.ts
```diff
@@ -1,5 +1,6 @@
 jest.mock('fs');
 jest.mock('../../utils/retrieveContractFromChain');
+const path = require('path');
```
webpro, line 3: This looks like two different ways to imports things, is there a reason to deviate from the rest in this file?

webpro (approved):

packages/tools/pactjs-cli/src/retrieve-contract/tests/retrieve-contract.test.ts
javadkh2: Nice catch!
No, I've just updated it

PR #283, opened 2023-05-15
Feat/input component

rezaerami:
- add input
- add input group
- add text field
- add disabled, valid, and invalid statuses
- add stories

image
image

## Check off the following:

- [ ] I have added tutorials.


packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
...
+  },
+});
+
+export const StyledLabel = styled('label', {
+  fontWeight: '500',
+  fontSize: '14px',
+});
+
+export const StyledTag = styled('span', {
+  marginLeft: '16px',
```
isaozler, line 30: can you use the specified spacing and sizing units $1,2,3 etc here and for the rest of the props

isaozler:
the component looks great @rezaerami  👍

packages/libs/react-components/src/components/Input/styles.ts
rezaerami: yes definitly, that was actually one of my questions regard spacings
also I have another issue since in the design actual background color for input is #ffffff, but since I didn't see it's changing regarding light or dark them, I used neutral1 and neutral6 for background and foreground colors, which I hope it's ok

isaozler:
small detail
image

the inner borders should be 0
only the first input will have top-right-left border radius
and the last input will have bottom-right-left border radius

packages/libs/react-components/src/components/Input/InputGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  children,
+  status,
+  disabled,
+  className,
+  ...rest
+}) => {
+  return (
+    <StyledInputGroupWrapper
+      {...rest}
+      className={classnames({ disabled }, className)}
```
eileenmguo, line 41: I believe the functionality you are achieving by using className can alternatively be achieved by usage of stitches variants which is what we are doing in other components
https://stitches.dev/docs/variants

I think using this feature is preferable to adding the classnames package for the sake of consistency and having fewer dependencies

packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
...
+  paddingBottom: '8px',
+  '*': {
+    display: 'inline-block',
+  },
+});
+
+export const StyledLabel = styled('label', {
+  fontWeight: '500',
+  fontSize: '14px',
+});
```
eileenmguo, line 27: I think it would be beneficial to update the `Text` component to accept `label` as a variant and element and define this style there as I expect this styling will be reused in other places as well

packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const StyledInput = styled('input', {
+  background: 'none',
+  padding: '8px 0',
+  border: 'none',
+  flex: '1',
+  outline: 'none',
+  fontSize: 'inherit',
+  color: '$colors$neutral6',
+  '&.hasLeadingText': {
```
eileenmguo, line 126: nit: If you look in the `stitches.config.ts` file, you can see all the utils that are defined. So you could achieve the same with the following shorthand `px: $4`

packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
+/* eslint @kadena-dev/typedef-var: 0 */
+// TODO: Remove this when this issue is resolved: https://github.com/kadena-community/kadena.js/issues/201
+
+import { styled } from '../../styles';
+
+export const styleVariant = {} as const;
```
eileenmguo, line 6: I assume this just needed to be cleaned up, but I think a valuable variant could be status

packages/libs/react-components/src/components/Input/InputGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+    >
+      {!!(label || tag || info) && (
+        <StyledInputGroupHeader>
+          {!!label && <StyledLabel>{label}</StyledLabel>}
+          {!!tag && (
+            <StyledTag className={classnames({ disabled })}>{tag}</StyledTag>
+          )}
+          {!!info && (
+            <StyledInfo>
+              <span>{info}</span>
```
eileenmguo, line 51: Would the text component work for this?

eileenmguo (commented):
Looking good so far!

packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const StyledInputGroupWrapper = styled('div', {
+  color: '$colors$neutral6',
+  '&.disabled': {
+    color: '$colors$neutral3',
```
sstraatemans, line 11: you can just use `$neutral` instead of `$colors$neutral` when its about color values (so color, borderColor, background etc)

packages/libs/react-components/src/components/Input/InputGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  StyledInputGroupWrapper,
+  StyledLabel,
+  StyledTag,
+} from './styles';
+
+import classnames from 'classnames';
+import React, { FC } from 'react';
+
+export interface IInputGroupProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'as'> {
```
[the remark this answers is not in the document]
rezaerami: i would say for an interactive element we should keep it, like inputs or buttons, but for a div, indeed it's not needed

packages/libs/react-components/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,59 @@
...
+    'as' | 'disabled' | 'children'
+  > {
+  as?: 'input';
+  leftPanel?: typeof SystemIcons[keyof typeof SystemIcons] | string;
+  rightPanel?: typeof SystemIcons[keyof typeof SystemIcons];
+  disabled?: boolean;
+  status?: 'success' | 'error';
+}
+
+export const Input: FC<IInputProps> = ({
```
sstraatemans, line 25: the input field is missing a label and tag.
I see that it is available in the input group.
If this is what was discussed, than I still think the input field should have a label, for a11y sake.
also the optional and helpermessage are per input field, no?

packages/libs/react-components/src/components/Input/InputGroup.tsx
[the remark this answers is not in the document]
sstraatemans: I think we want to be very strict with our components.
if we do something like this, we also open the `style` prop, which we dont want to expose.
So I think we want to be explicit for every component

packages/libs/react-components/src/components/Input/Input.tsx
rezaerami: yes, following discussion with @eileenmguo and @isaozler we decided to put label, tag, info, and helper text in input group
although all together are also available as a text field

packages/libs/react-components/src/components/Input/styles.ts
```diff
@@ -0,0 +1,130 @@
...
+  height: '100%',
+});
+
+export const StyledIconWrapper = styled('span', {
+  padding: '8px 16px',
+  fontSize: '16px',
+  '> *': {
+    display: 'flex',
+    alignItems: 'center',
+  },
```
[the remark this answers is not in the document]
eileenmguo: Yeah, but I believe this selector is targeting the span that wraps the svg in the Icon component, so it's actually applying styling within the component itself

packages/libs/react-components/src/components/Input/InputGroup.tsx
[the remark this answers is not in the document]
eileenmguo: Yeah, based on our previous discussions about wanting to have a very strict component library, I think it makes sense to keep it strict by default and offer more flexibility only when we deem it necessary.

packages/libs/react-components/src/components/Input/InputGroup.tsx
rezaerami: yes, but the issue is it will regenerate styles every time, therefore in a page with many components, running it on browser causes lag, but with classnames you'll have style there, and it'll just changes it's look, so there is no style regeneration

packages/libs/react-components/src/components/Input/InputGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export interface IInputGroupProps {
+  label?: string;
+  tag?: string;
+  info?: string;
+  helper?: string;
+  status?: 'success' | 'error';
+  disabled?: boolean;
+  children: React.ReactNode;
+  className: string;
```
sstraatemans, line 25: what we need this classname for?
is this for extending / overwriting classes and variants?

I think this is something we dont want, as we want the styles to be strict

packages/libs/react-components/src/components/Input/InputGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const InputGroup: FC<IInputGroupProps> = ({
+  label,
+  tag,
+  info,
+  helper,
+  children,
+  status,
+  disabled,
+  className,
+  ...rest
```
sstraatemans, line 37: I dont think we need rest props right? Because we have specified which props are allowed

packages/libs/react-components/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,59 @@
...
+  StyledLeadingText,
+} from './styles';
+
+import classnames from 'classnames';
+import React, { FC } from 'react';
+
+export interface IInputProps
+  extends Omit<
+    React.HTMLAttributes<HTMLInputElement>,
+    'as' | 'disabled' | 'children'
```
sstraatemans, line 16: I think we want to omit the classname here as well. we want to be strict in styling

rezaerami: by omitting classnames we won't be able to extend a component in styles, cause all styles are applied by classname under the hood of stitches, therefore if we omit it, and then extend it as a styled component, we won't be able to overwrite any styles

packages/libs/react-components/src/components/Input/InputGroup.tsx
rezaerami: yes exactly, it's for extending, but i don't think it's a good idea to set a strict rule like this, cause in the end some styles for different usages can be different, or at least there should be room for modification of styles if we don't allow styled component extension, the consumer will wrap it in another element to overwrite it, which messes up the accessibility so I would say having extension ability would be beneficial

sstraatemans: hm..yeah i understand. 
I do have the opinion that this package, at the moment is for Kadena UI alone, so keep it as strict until we really need it.
If we open up the classname a developer can do what they want, where maybe the style system does not allow it (lets say, make the text green)
If the UI team dont give the classname option, and the developer wants something that is not possible, they need to go talk to the UI team. if this can be implemented. where, in the case of the green text, they probably say no.

I think a design system should not be overwritten.
But we can discuss with Eileen.

rezaerami: remove classnames

packages/libs/react-components/src/components/Input/InputGroup.tsx
eileenmguo: Yeah, I understand your rationale and I think it's a good approach, but it doesn't align with the standard we already have. In this circumstance where we will not develop this package further, I would prefer to keep things consistent

I think this might be a good approach for the next component library though!

sstraatemans (approved):

PR #284, opened 2023-05-15
chore(cka): replace shell script with zx script

javadkh2:
## Reason:
Since shell scripts are not natively executable in Windows, rush test fails on that platform

installed zx and replace the shell script with zx script

## Check off the following:

- [ ] I have added tutorials.


Takadenoshi (approved):

Takadenoshi:
Should we also clean up the test directories after execution? The previous script did not do this but it would be a worthwhile improvement. The paths are not in gitignore so they are left as untracked files.

webpro:
Looking at the contents of the cross-OS script maybe we could also consider not using a dependency at all?

The thing is, `zx` is great but it does not prevent anyone from using commands that are not cross-OS. While Node.js itself has fine primitives and built-ins to do what we seem to need, forcing ourselves not to break stuff in another environment than our own.

packages/tools/create-kadena-app/scripts/generate-test.mjs
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env zx
+
+import 'zx/globals';
+
+// This script is used to generate projects for every supported template and test them.
+const generateAndTest = (template) =>
+  within(async () => {
+    const directory = `test-${template}-project`;
+    // echo "Generating the project"
+    await $`node lib/index.js generate-project --name ${directory} --template ${template}`;
```
Takadenoshi, line 10: does the (forward) slash work in windows? I think we might want `path.join()` here instead

Takadenoshi (changes requested):

webpro:
> Not sure what you are recommending in lieu of zx by "not using a dependency". pure node and child_process.spawn() ?

Pure Node.js, since afaics spawning seems unnecessary as we do `node script.js` so we could instead `import` that other script?

But yeah, when we really need to execute external binaries like `git` that we can't call from npm/npx scripts we can use `zx` (or [execa](https://github.com/sindresorhus/execa) which is more light-weight).

packages/tools/create-kadena-app/scripts/generate-test.mjs
Takadenoshi: agreed that we must be aware to be explicitly os agnostic but in my view this point is perpendicular to the tool being used. With regards to this, `zx` is the same as execa or spawning a process with node as far as I can tell

Takadenoshi:
For this particular script yes, we could use import, but generally it would be nice to have a bash scripting replacement that is able to quickly glue things together, even if we are limited to a very small set of commands that would be available in vanilla windows environments. I'd rather have the conveniences and wrappers that zx provides (`await $...`) than use spawn or write our own wrappers. I didn't know about execa so I don't mind this or that - it looks like the friendly `$` interface is similarly available there.

@javadkh2 We have another dev script [here](https://github.com/kadena-community/kadena.js/blob/chore/zx-script-for-cka/packages/apps/graph/src/utils/dev-database/clean-csv-script.sh) but, since it is not invoked from anywhere I think it might be fine to leave it as-is

packages/tools/create-kadena-app/scripts/generate-test.mjs
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env zx
+// This script is used to generate projects for every supported template and test them.
+
+import 'zx/globals';
+
+import { rm } from 'fs/promises';
+
+// the default log uses stderr for write output of commands, even for commands that do not generate errors or warnings.
+// This can cause Rush to treat everything as a warning message. This function changes this behavior.
+$.log = (entry) => {
```
javadkh2, line 10: I an unexpected behaviour in the default log function. 
I have created an issue on the repo. if they change the behaviour we can remove this function
https://github.com/google/zx/issues/623

webpro:
> For this particular script yes, we could use import, but generally it would be nice to have a bash scripting replacement that is able to quickly glue things together, even if we are limited to a very small set of commands that would be available in vanilla windows environments. I'd rather have the conveniences and wrappers that zx provides (`await $...`) than use spawn or write our own wrappers. I didn't know about execa so I don't mind this or that - it looks like the friendly `$` interface is similarly available there.

Agreed, I was responding to what I'm seeing in this PR. We shouldn't write our own wrappers or anything. Yet as both the monorepo and the number of people working on it are growing I think some governance/policies to keep dependencies manageable will help.

packages/tools/create-kadena-app/scripts/generate-test.mjs
javadkh2: I have updated the code.
@webpro that was an important tip, I will make sure everything is working as expected on Mac, Windows and WSL

javadkh2:
@Takadenoshi I will have a look at those scripts

alber70g (approved):

ggobugi27 (approved):

jermaine150 (approved):

alber70g (approved):

Takadenoshi (approved):

Takadenoshi:
We decided not to use git hooks so I will close this

PR #293, opened 2023-05-17
Introduce Knip to clean up unused files, dependencies and exports

webpro:
**WIP - DO NOT MERGE**

The idea is that each package in the monorepo can be looked after separately, or all together. The nice thing about [Knip](https://github.com/webpro/knip) is that, even though I created it and it's far from perfect, it gives starting points to keep the monorepo clean & tidy. Otherwise it's "mopping with the tap open" now that we're growing and keep building new features. Automation for the win!

## TODOs

- [ ] The issue I'm facing is that it's not yet possible to run Knip from root with Rush, which is required. When running for a single package, it doesn't know about the other packages in the workspace.
- [ ] I hope it's possible to have Rush simply pass through any CLI arguments to the underlying tool (otherwise we'd have to add each and every argument to `rush.json`?)
- [ ] Add configuration and fine-tune Knip

## Example

Here's an example of how I would intend to use it:

```
❯ knip --workspace packages/libs/kadena.js
Unused dependencies (4)
bignumber.js  packages/libs/kadena.js/package.json
blakejs       packages/libs/kadena.js/package.json
node-fetch    packages/libs/kadena.js/package.json
tweetnacl     packages/libs/kadena.js/package.json
Unused devDependencies (8)
@kadena-dev/eslint-plugin  packages/libs/kadena.js/package.json
@kadena-dev/heft-rig       packages/libs/kadena.js/package.json
@rushstack/eslint-config   packages/libs/kadena.js/package.json
@types/heft-jest           packages/libs/kadena.js/package.json
@types/node-fetch          packages/libs/kadena.js/package.json
serve                      packages/libs/kadena.js/package.json
ts-loader                  packages/libs/kadena.js/package.json
webpack-cli                packages/libs/kadena.js/package.json
Unlisted binaries (1)
python  packages/libs/kadena.js/package.json
Unused exports (3)
keyPair             packages/libs/kadena.js/src/api/tests/mockdata/contCommand.ts
payload             packages/libs/kadena.js/src/api/tests/mockdata/contCommand.ts
stringifiedPayload  packages/libs/kadena.js/src/api/tests/mockdata/contCommand.ts
```

Some dependencies may actually be used, but that's usually either a bug in Knip or implicit/magic from whatever tool we're using (we can configure stuff to be ignored).


alber70g:
Autoinstallers or rush plugins are the likely way to integrate this. 

https://rushjs.io/pages/maintainer/autoinstallers/
https://rushjs.io/pages/maintainer/using_rush_plugins/

webpro:
Giving up on this one, mostly because Rush and Knip aren't really "compatible":

- Rush has no `package.json` in the root of the project, while Knip requires one (or a `pnpm-workspaces.yaml`) in the root of the project (at least to know about all workspaces).
- Rush has more levels of indirection around rig stuff (https://rushstack.io/pages/heft/rig_packages/), forcing more config on Knip to ignore certain `devDependencies` in most packages.

Knip can still be used by running it in a package directory (either globally installed or as a package `devDependency`), but the idea of this PR to run it integrated within Rush standards seems not really feasible/worthwhile.

I'll look into cleaning up packages separately and open PRs along the road.

PR #297, opened 2023-05-17
Rewrite pre-commit hook in zx

Takadenoshi:
In order to support vanilla/native Windows environments we need to remove bash scripts in favor of a portable scripting solution.

For the pre-commit git hook the implementation is not obvious as it is not in a package/project scope as far as rush is concerned, and as such can't import the `zx` module.

As a workaround I am introducing an "internal" project under `common/scripts/pre-commit` that houses the core of the pre-commit script. The actual script in .git/hooks/pre-commit is just a "symlink" import() statement.

I don't like it but I am not sure how to best approach this.

While the pre-commit hook is just an invocation of a node script, in this case we cannot (safely) invoke it with an import() directly as the rush API recommends against third party tools importing the rush-cli package


alber70g (approved):

javadkh2 (approved):

common/scripts/pre-commit/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "common-scripts-precommit",
+  "version": "0.0.0",
+  "scripts": {
+    "build": "echo OK",
+    "test": "echo \"Error: no test specified\" && exit 1"
```
alber70g, line 6: I don't think they're needed. If they are, we might also need the _phase:* commands

rush.json
```diff
@@ -588,7 +588,14 @@
...
       "versionPolicyName": "alpha"
-    }
+    },
+
+
+    {
+      "packageName": "common-scripts-precommit",
+      "projectFolder": "common/scripts/pre-commit",
+      "shouldPublish": false
+    },
```
alber70g, line 598: Can we rename this to `@kadena-dev/precommit-scripts` or something alike and sort inside tools by packageName?

alber70g (changes requested):

common/scripts/pre-commit/package.json
Takadenoshi: Both are required - otherwise building the repo errors out with `ERROR: The project 'common-scripts-precommit' does not define a 'build' command in the 'scripts' section of its package.json` (same for test)

Takadenoshi:
@alber70g happy with these changes?

webpro:
I think this one be closed now that we're no longer using Git hooks?

PR #301, opened 2023-05-22
Fix usage of `repository` and `npx` in package.json

webpro:
The `repository` field did originally not support `directories`, but now we can add them, which is in this PR (also see https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository).

Since package managers should add `bin` directory to the path automatically, there's no need to put `npx` in between when running scripts.

### DIFF
--- a/common/changes/@kadena-dev/eslint-config/feat-fix-packagejson-fields_2023-05-22-10-49.json
+++ b/common/changes/@kadena-dev/eslint-config/feat-fix-packagejson-fields_2023-05-22-10-49.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena-dev/eslint-config",
+      "comment": "Fix usage of `repository` and `npx` in package.json",
+      "type": "none"
+    }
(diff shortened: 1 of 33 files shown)


alber70g (approved):

Randynamic (approved):

ash-vd (approved):

webpro:
So this PR requires a lot of approvals, since it touches a lot of packages. Unfortunately github does not provide an efficient/optimal list of approvals to unblock this PR, but I think it's this:

```
/packages/libs/chainweb-stream-client           @Takadenoshi
/packages/tools/create-kadena-app               @jermaine150
```

By the way, I think packages should have at least two owners, but that'll probably be resolved once we install teams on github.

Takadenoshi (approved):

webpro:
Oh the joy of changing something in almost each and every package, I missed one:

```
/packages/libs/react-components                 @eileenmguo @sstraatemans @isaozler
```

sstraatemans (approved):

PR #302, opened 2023-05-22
[Docs] add analytics

sstraatemans:
## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.


packages/apps/kadena-docs/src/components/Analytics/Analytics.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    gtag('config', TRACKING_ID, {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      send_page_view: false, //manually send page views to have full control
+    });
+    gtag('event', 'page_view', {
+      page_path: window.location.pathname,
+      send_to: TRACKING_ID,
+    });
+  }, []);
```
alber70g, line 22: Do we look for DoNotTrack property/header/whatever? Or is this something we don't care about?

packages/apps/kadena-docs/src/types/ApiResponse.ts
alber70g, line 1: It might be better to put generic types like this in a `d.ts`

packages/apps/kadena-docs/tsconfig.json
```diff
@@ -1,14 +1,9 @@
 {
   "extends": "./node_modules/@kadena-dev/heft-rig/profiles/default/tsconfig-base.json",
   "compilerOptions": {
-    "types": [
-      "heft-jest",
-      "node"
-    ],
+    "types": ["heft-jest", "node", "@types/gtag.js"],
```
alber70g, line 4: `@types/gtag.js` shouldn't be needed as all `@types/*` is included by typescript in the `typeRoots` compiler option

packages/apps/kadena-docs/src/types/ApiResponse.ts
alber70g: As discussed, better to move the type to the source of the concrete value. So move [runReport](https://github.com/kadena-community/kadena.js/blob/d62f0290b654c0b65e2135a324ac96c9b7afaecd/packages/apps/kadena-docs/src/pages/index.tsx#L136] also to a different file. Something like `analytics-client`

ash-vd (approved):

PR #305, opened 2023-05-22
Bootstrapping `dev-cli` from a different repository

EnoF:
- tests to be part of the src and fix build moved from heft to tsc
 - ava test to exclude d.ts files

## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.


alber70g:
Few points to start with:
- can we integrate this with the existing pactjs-cli (we can rename `@kadena/pactjs-cli` to `@kadena/cli`)
- is there anything specific that you need to use `ink` that cannot be done with `inquirer`. Relying on `react` has shown to be troublesome as specific versions have to be used, or can it be a peer-dependency to prevent issues?
- can you update the tsconfig to be more specific and inherit from the tsconfig-base (the version below should work)
- can we move the typescript 5.x upgrade to a separate PR (merge that one before this one)
- upgrade the generic packages in a separate PR 
   I see these two have been changed     
   - "@rushstack/eslint-config": "~3.3.0", 
   - "rushstack/heft": "~0.50.6",
- I see `packages/libs/chainwebjs/etc/chainwebjs.api.md` has changed. Is that an actual change or something related to the typescript 5 upgrade?
- `packages/tools/kda-cli/tests/questions.test.tsx` is using tabs instead of spaces, is prettier ran correctly?

```json
{
  "compilerOptions": {
    "types": ["node"],
    "module": "Node16",
    "moduleResolution": "node16",
    "skipLibCheck": true
  },
  "ts-node": {
    "esm": true
  }
}
```

PR #319, opened 2023-05-24
Various minor fixes and cleanups for the election dApp tut

webpro:
## Reason:

Closes [issue link]

## Check off the following:

- [x] I have added tutorials.

### DIFF
--- a/packages/tutorials/election-dapp/pact/election.pact
+++ b/packages/tutorials/election-dapp/pact/election.pact
@@ -129,7 +129,7 @@
     "Get the votes count by cid"
 
     ;; Read the row using cid as key and select only the `votes` column
-    (at 'votes (read candidates cid ['votes]))
+    (at 'votes (read candidates cid ["votes"]))
   )
 
   (defun insert-candidate (candidate)
--- a/packages/tutorials/election-dapp/pact/election.pact
+++ b/packages/tutorials/election-dapp/pact/election.pact
@@ -129,7 +129,7 @@
     "Get the votes count by cid"
 
     ;; Read the row using cid as key and select only the `votes` column
-    (at 'votes (read candidates cid ["votes"]))
+    (at 'votes (read candidates cid ['votes]))
   )
 
   (defun insert-candidate (candidate)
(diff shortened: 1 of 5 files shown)


ash-vd (approved):

packages/tutorials/election-dapp/pact/election.pact
```diff
@@ -129,7 +129,7 @@
     "Get the votes count by cid"
 
     ;; Read the row using cid as key and select only the `votes` column
-    (at 'votes (read candidates cid ['votes]))
+    (at 'votes (read candidates cid ["votes"]))
```
EnoF, line 132: > Symbols are string literals representing some unique item in the runtime, like a function or a table name. Their representation internally is simply a string literal so their usage is idiomatic.
>Symbols are created with a preceding tick, thus they do not support whitespace nor multiline syntax.

Though on most projects I've seen, usually they use a symbol where possible, probably because it's more strict and therefore conveys the idea better? So I'm not entirely sure if we want to change from from `'` to `"`

alber70g (approved):

jermaine150 (approved):

packages/tutorials/election-dapp/pact/election.pact
webpro: The single quote breaks syntax highlighting

trendzetter: shouldn't be the case. the atom pact-lang plugin does take it into account and I submitted a patch for the vscode plugin that can be merged too: https://github.com/kadena-community/pact-vscode/pull/3

trendzetter: > The single quote breaks syntax highlighting

shouldn't be the case. the atom pact-lang plugin does take it into account and I submitted a patch for the vscode plugin that can be merged too: https://github.com/kadena-community/pact-vscode/pull/3

Randynamic (approved):

packages/tutorials/election-dapp/pact/election.pact
webpro: I think in general not closing such strings is confusing, but not gonna argue that here. I reverted this change as it is not a fix like I thought initially.

PR #320, opened 2023-05-24
feat(pg): pact parser

javadkh2:
## Reason:
We parse pact contracts to generate type definition files for contracts that users later import from the @kadena/client package. However, the existing pact parser lacks flexibility and maintainability, making it difficult to add new features.

This pull request introduces a new approach to parsing pact contracts. It includes general utilities located in the parser-utilities folder, which are used to compose the pact grammar. You can find the grammar itself in the pactGrammar.ts file. Additionally, I have updated the existing CLI tool to generate types based on both the old and new parsers.

## Check off the following:

- [ ] I have added tutorials.


packages/libs/pactjs-generator/src/contract/parsing/pactParser.ts
```diff
@@ -0,0 +1,410 @@
...
+      return storage;
+    },
+  };
+};
+
+interface IModuleWithPointer extends IModule {
+  pointer?: IPointer;
+}
+
+// parse a module and its dependencies
```
alber70g, line 231: If you want to add specific docs, you can add it by using `/** */`. That will be part of the generated reference docs as well. You can also add semver indicators. See also https://api-extractor.com/pages/overview/intro/ 
```suggestion
/**
 * parse a module and its dependencies
 * @alpha
 */
```

packages/libs/pactjs-generator/src/contract/parsing/pactParser.ts
```diff
@@ -0,0 +1,410 @@
...
+async function loadModuleDependencies(
+  fullName: string,
+  getModule: (name: string) => Promise<IModuleWithPointer | undefined>,
+): Promise<Array<IModuleWithPointer>> {
```
alber70g, line 235: If you want to add specific docs, you can add it by using `/** */`. That will be part of the generated reference docs as well. See also https://api-extractor.com/pages/overview/intro/ 

```suggestion
/**
 * parse a module and its dependencies
 * @alpha
 */
async function loadModuleDependencies(
  fullName: string,
  getModule: (name: string) => Promise<IModuleWithPointer | undefined>,
): Promise<Array<IModuleWithPointer>> {
```

Applies to below occurrences as well

packages/libs/pactjs-generator/src/contract/parsing/pactParser.ts
```diff
@@ -0,0 +1,410 @@
...
+  if (bodyPointer !== undefined) {
+    pointer.reset(bodyPointer);
+    const blockPointer = getBlockPointer(pointer);
+    const data = functionCalls(fnList, availableExternals)(blockPointer);
+    if (data !== FAILED) {
+      return unwrapData(data);
+    }
+  }
+
+  return {};
```
alber70g, line 278: Can you invert the if statement so it'll be an early return?

```suggestion
  if (bodyPointer === undefined) {
    return {};
  }
  
  pointer.reset(bodyPointer);
  const blockPointer = getBlockPointer(pointer);
  const data = functionCalls(fnList, availableExternals)(blockPointer);
  if (data !== FAILED) {
    return unwrapData(data);
  }
```

packages/libs/pactjs-generator/src/contract/parsing/utils/dataWrapper.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable @kadena-dev/no-eslint-disable */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export interface IWrappedData<
+  T extends any = any,
+  N extends string | undefined = string | undefined,
+> {
+  name: N;
+  inspect: true;
+  data: T;
+}
```
alber70g, line 10: It could be nice to have a little more descriptive type argument names. E.g. `TData` and `TName`, what do you think?

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/README.md
alber70g, line 1: Can you run prettier on this?

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/inspect.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const $: IInspector = (one: string | IParser, second?: IParser) =>
+  rule((pointer) => {
+    const name = typeof one === 'string' ? one : undefined;
+    if (name === 'inspect') throw new Error('inspect is a reserved name');
+    const parser = second || (one as IParser);
+    const result = parser(pointer);
+    if (result === FAILED) return FAILED;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return wrapData(result, name) as any;
```
alber70g, line 29: You could use `unknown` here, that would satisfy the rule, and as it's recast by the interface it'll be fine for the caller

```suggestion
    return wrapData(result, name) as unknown;
```

Maybe this is a solution for other places as well

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/repeat.ts
```diff
@@ -0,0 +1,79 @@
...
+  <T extends Array<IParser>>(...parsers: T): IParser<
+    IWrappedData<
+      Partial<
+        UnionToIntersection<
+          MakeArr<UnwrappedObjects<IsWrappedData<RuleReturnType<T[number]>>>>
+        >
+      >,
+      undefined
+    >
+  >;
```
alber70g, line 27: 😍

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/rule.ts
```diff
@@ -0,0 +1,33 @@
+import { ExWrappedData } from '../dataWrapper';
+import { IPointer } from '../getPointer';
+
+// eslint-disable-next-line @rushstack/typedef-var
+export const FAILED = Symbol('FAILED');
```
alber70g, line 5:
```suggestion
export const FAILED: symbol = Symbol('FAILED');
```
?

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/tests/asString.test.ts
```diff
@@ -0,0 +1,27 @@
+import { getPointer } from '../../getPointer';
+import { asString } from '../asString';
+import { block } from '../block';
+import { id } from '../id';
+import { FAILED } from '../rule';
+import { seq } from '../seq';
+import { skipTheRest } from '../skip';
+
+describe('asString parser', () => {
+  it('should return result as string if rule is matched', () => {
```
alber70g, line 10: we can make the `it`s without `should`.
```suggestion
  it('returns result as string if rule is matched', () => {
```

Idem for the rest

packages/libs/pactjs-generator/src/index.ts
```diff
@@ -18,6 +18,11 @@ export * from './contract/generation/IContractDefinition';
  */
 export * from './contract/generation/generator';
 
+/**
+ * @alpha
+ */
+export * from './contract/generation/generatorV2';
+
```
alber70g, line 25: I'm not sure if it's needed to put the semver annotation here, if they're already at the function itself.

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -1,13 +1,14 @@
...
-  IContractDefinition,
+  generateDts2,
+  pactParser,
   StringContractDefinition,
 } from '@kadena/pactjs-generator';
 
 import { retrieveContractFromChain } from '../utils/retrieveContractFromChain';
 
-import { ContractGenerateOptions } from './';
+import { IContractGenerateOptions } from './';
```
alber70g, line 11: Can we import from the file instead of the barrel file? Otherwise we inadvertently import everything from the barrel file

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -47,145 +48,213 @@ function verifyTsconfigTypings(
...
+      );
+      return content ?? '';
     }
+    console.log(`
+      the generator tries to fetch ${name} from the blockchain but the api data is not presented.
+      this happen because ${name} mentioned via --contracts directly or it is a dependency of a module.
+      the scrip skips this module
+      `);
+    return '';
+  };
```
alber70g, line 131: Do we want to log with the indents? This log means that it'll be written like so

```sh
$ some command
      the generator tries to fetch ${name} from the blockchain but the api data is not presented.
      this happen because ${name} mentioned via --contracts directly or it is a dependency of a module.
      the scrip skips this module
```

Otherwise we can do this:
```suggestion
    console.log(`
the generator tries to fetch ${name} from the blockchain but the api data is not presented.
this happen because ${name} mentioned via --contracts directly or it is a dependency of a module.
the scrip skips this module\n`);
    return '';
  };
```

alber70g (approved):

packages/libs/pactjs-generator/src/contract/parsing/utils/parser-utilities/rule.ts
javadkh2: I want a unique symbol. but `:symbol` just mention its a symbol not a specific symbol and then it breaks all of the places that work with FAILED symbol.
So basically 
```typescript
const A = Symbol('A');
const B = Symbol('B');
// This comparison appears to be unintentional because the types 'typeof A' and 'typeof B' have no overlap.
if (A === B) {
}
```
but this one does not throw the ts error
```typescript
const A:symbol = Symbol('A');
const B:symbol = Symbol('B');
// type script cant understand these symbols are dufferent
if (A === B) {
}
```

javadkh2: And what is the point of adding a type for primitive types at all. I believe its redundant and sometimes just does not let typescript to refer types completely .
```typescript
// typescript narrows the type to "HELLO" ; much better type for a const
const A = "HELLO"

// this is just a string
const B:string = "HELLO"

File: packages/libs/client/etc/client.api.md [hunk 12]
```diff
@@ -24,6 +24,12 @@ export function buildUnsignedTransaction(parts: string[], holes: string[], args:
 // @alpha (undocumented)
 export function createPactCommandFromTemplate(tpl: IPactCommand): PactCommand;
 
+// @alpha (undocumented)
+export interface ICapV2 {
+    // (undocumented)
+    addCap(cap: string, signer: string, ...args: PactValue[]): this;
```
Randynamic, line 30: I would love to have a more verbose API like `addCapabilities` as kt is more clear and more inline with our renewed domain model

PR #326, opened 2023-05-25
Remove tutorials (in favor of kadena-community/voting-dapp)

webpro:
Some things in this monorepo complicate versioning and/or dependency management. More importantly it has no benefit for users, it's much clearer from the tutorial to have a simple and clean repo to refer to and clone, etc.


alber70g:
I agree with the idea that we need to lower the barrier to use this tutorial.

However, now that it's moved out of the monorepo, how do we ensure that any updates to the libs will only be merged and released once the tutorials is fixed?

webpro:
We don't. The tutorial works with a certain version of some our libs. Saying that we can only release libs once our tutorial is fixed is the same as saying that we should ensure all apps of all consumers of our libs must keep working. Consumers, just like our own tutorials, upgrade at their own pace. Eventually tooling is just a means to an end, let's overcomplicate things.

alber70g:
With the libraries in an unstable state at this time, it would not be beneficial to have the tutorials not be up-to-date with them. I agree with your statement, if we had stable versions of all the libraries, but that is not the case now.

alber70g (approved):

alber70g:
When we move this to the other repo let's pin the versions of the @kadena packages to a single version.

jermaine150 (approved):

PR #340, opened 2023-05-31
[@kadena/transfer] Enable safe transfer

KristinaSpasevska:
## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/apps/transfer/src/services/transfer/coin-transfer.ts
+++ b/packages/apps/transfer/src/services/transfer/coin-transfer.ts
@@ -77,3 +77,94 @@ export async function transferCreate(
   await pactCommand.send(generateApiHost(networkId, chainId));
   return pactCommand;
 }
+
+export async function safeTransferCreate(
+  fromAccount: string,
+  toAccount: string,
+  amount: string,
+  fromPrivateKey: string,
+  toPrivateKey: string,
+  chainId: ChainId,
+  networkId: ChainwebNetworkId,
+): Promise<PactCommand> {
+  if (isNaN(Number(amount))) {
+    throw new Error('Amount must be a number');
+  }
+
+  // sender need to send extra coins to the receiver which will be returned by the receiver
+  const amountWithExtra = Number(amount) + 0.00001;
+
+  const pactCommand = new PactCommand();
+  pactCommand.code = `(coin.transfer-create "${fromAccount}" "${toAccount}" (read-keyset "ks") ${convertDecimal(
+    amountWithExtra.toString(),
+  )})(coin.transfer "${toAccount}" "${fromAccount}" ${convertDecimal(
+    '0.00001',
+  )})`;
+
+  pactCommand
+    .addCap('coin.GAS', onlyKey(fromAccount))
+    .addCap(
+      'coin.TRANSFER',
+      onlyKey(fromAccount),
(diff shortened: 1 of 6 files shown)


nil-amrutlal-dept:
I've made some changes to the existing variable names and also to the structure of the code on my cross-chain transfer branch. Since I am not able to complete the transfer yet, I'd say we can merge this one when ready and then I can refine everything (there will probably be conflicts).

nil-amrutlal-dept (approved):

packages/apps/transfer/src/services/transfer/coin-transfer.ts
```diff
@@ -77,3 +77,94 @@ export async function transferCreate(
...
+
+  // sender need to send extra coins to the receiver which will be returned by the receiver
+  const amountWithExtra = Number(amount) + 0.000000000001;
+
+  const pactCommand = new PactCommand();
+  pactCommand.code = `(coin.transfer-create "${fromAccount}" "${toAccount}" (read-keyset "ks") ${convertDecimal(
+    amountWithExtra.toString(),
+  )})(coin.transfer "${toAccount}" "${fromAccount}" ${convertDecimal(
+    '0.000000000001',
+  )})`;
```
alber70g, line 102: In the implementation of `convertDecimal` it'd be also better to use either `{ decimal: "10.0" }` or PactNumber or BigNumber

packages/apps/transfer/src/services/transfer/coin-transfer.ts
```diff
@@ -77,3 +77,94 @@ export async function transferCreate(
...
+
+  pactCommand
+    .addCap('coin.GAS', onlyKey(fromAccount))
+    .addCap(
```
alber70g, line 106: Is it on purpose that only `x:y` type accounts are allowed? There can be named accounts that don't have the prefix `x:` (`x` being anything)

packages/apps/transfer/src/services/transfer/coin-transfer.ts
```diff
@@ -77,3 +77,94 @@ export async function transferCreate(
...
+      'coin.TRANSFER',
+      onlyKey(fromAccount),
+      fromAccount,
+      toAccount,
+      Number(amountWithExtra),
```
alber70g, line 111: Idem

packages/apps/transfer/src/services/transfer/coin-transfer.ts
```diff
@@ -77,3 +77,94 @@ export async function transferCreate(
...
+    secretKey: toPrivateKey,
+    publicKey: onlyKey(toAccount),
+  });
+
+  if (
+    signatureSender.sig === undefined ||
+    signatureReceiver.sig === undefined
+  ) {
+    throw new Error('Failed to sign transaction');
+  }
```
alber70g, line 153: The sender and receiver could be signed in a separate process where the transaction from `createCommand()` could be sent to the other person via an async process. 
Does the feature allow for that use case?

alber70g (changes requested):
Some comments

MRVDH:
Note: PR on hold since it was decided not to do transfers in the dev tooling website yet.

PR #291, opened 2023-05-16
[@kadena/react-ui]: Adding new component library using vanilla extract 

eileenmguo:
## Reason:

Due to discussions within the team about the implications of runtime CSS solutions like Stitches, we have decided to switch to vanilla-extract/css (VE). This PR sets up the foundation for the new component library so it behaves similarly to the previous one (with tokens aligned with our design system) with the added features of VE like the sprinkles utility classes.

Things to note: 
- There were difficulties with Next.js was using the cjs version of the package for server builds and the esm version for browser builds which caused a classname misalignment. Similar to this [issue](https://github.com/vanilla-extract-css/vanilla-extract/issues/940). This was solved by adding @kadena/react-ui to the `transpileModules` configuration in Next.js which requires Next.js version >= 13.1.0
- The button component has been updated to look like the designs, but for the sake of not adding more to this PR, I have not added Icon support or different style variations (for example unfilled). These will be added in the future
- We will likely need to use [recipes](https://vanilla-extract.style/documentation/packages/recipes/) for styling in the future, but this is not included in this PR

## Check off the following:

- [x] I have reviewed my changes and run the appropriate tests.
- [x] I have have run `rush change` to add the appropriate change logs.
- [x] I have added/edited docs.
- [x] I have added tutorials.
- [x] I have double checked and DEFINITELY added docs.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204541876909328

PR #342, opened 2023-06-01
Format generated test file

webpro:
See https://github.com/kadena-community/kadena.js/pull/291#discussion_r1211732301

This is a quick fix so the generated file is reformatted according to our Prettier config. Ideally we would do it directly in code but that would require more changes and adding the dependency, but this is a test project so I didn't bother.

It's kinda weird that we have to take into account that Rush does not execute `npm run pre[task]` or `npm run post[task]` but in this case we can take advantage of it.

### DIFF
--- a/packages/apps/kadena-docs/next-env.d.ts
+++ b/packages/apps/kadena-docs/next-env.d.ts
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference types="next/navigation-types/compat/navigation" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/basic-features/typescript for more information.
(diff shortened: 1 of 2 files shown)


alber70g:
I thought we removed it from being formatted and added it to `.prettierignore`. Maybe an idea to not format generated code in general?

webpro:
> I thought we removed it from being formatted and added it to `.prettierignore`. Maybe an idea to not format generated code in general?

Yes, totally agree. I assumed that the problem had to do with prettier not picking up the root ignore list when executed from a package dir, or something else more complicated (since nobody fixed it). But everything seems OK afaic, so I just pushed a change so this PR results in just an unformatted generated file.

alber70g (approved):

Randynamic (approved):

ash-vd (approved):

PR #347, opened 2023-06-01
Improve and make issue templates more consistent

webpro:
We can have fancy forms in our issue templates. This PRs makes three consistent ones:

- Bug report
- Feature request
- Documentation suggestion

The idea is that we don't need to document extensively how to file bug reports elsewhere, so we'll need to guide users properly within the GitHub bug reports.

Please let me know anything that's missing or could be improved.

---
  - https://app.asana.com/0/0/1204735333828871


.github/ISSUE_TEMPLATE/001-bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+    attributes:
+      label: What operating system are you using?
+      placeholder: Mac, Windows, Linux
+    validations:
+      required: false
+
+  - type: input
+    id: browser
+    attributes:
+      label: What browser are you using?
```
ash-vd, line 26: Might be good to also ask for the version

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,5 @@
 blank_issues_enabled: false
 contact_links:
-  - name: Kadena Discord Kadena.js Support (Kadena.js channel)
+  - name: Kadena Discord
     url: https://discord.io/kadena
-    about: Contact us on discord if you're having trouble with the library.
+    about: Contact us on Discord if you're having trouble with the library.
```
ash-vd, line 5:
```suggestion
    about: Contact us on Discord if you're having trouble with the library. We have a specific channel for Kadena.js. "#kadena-js" can be found in the "Buidler Zone".
```

ash-vd (approved):

.github/ISSUE_TEMPLATE/001-bug_report.yml
```diff
@@ -0,0 +1,64 @@
...
+description: Report an issue with Kadena
+labels:
+  - bug
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to file a bug report. Please fill out this form as completely as possible.
+
+        ✅ I have read the [Code of Conduct](https://github.com/kadena-community/kadena.js/blob/master/CODE_OF_CONDUCT.md)
```
MRVDH, line 11: Shouldn't this be:

- [ ] I have read the [Code of Conduct](https://github.com/kadena-community/kadena.js/blob/master/CODE_OF_CONDUCT.md)
- [ ] The bug was not reported before
- [ ] I am using the **latest version** of the relevant package(s)

Instead of emojis

Also applicable in other places

webpro: That's a great idea, it will enforce people to read and confirm; updated.

.github/ISSUE_TEMPLATE/001-bug_report.yml
```diff
@@ -0,0 +1,75 @@
...
+
+  - type: textarea
+    id: description
+    attributes:
+      label: Describe the bug
+      description: A clear and concise description of the bug
+    validations:
+      required: true
+
+  - type: input
```
lkononenko, line 50: Probably it should be a textarea if we expect the steps, also in order to have a possibility to add attachments for screenshots/videos/gifs?

alber70g:
Is it an idea to have specific bug report templates for **specific packages/topics** in the repo?
That way we can add the prefix for each issue `[@kadena/client] ...` and add questions specific to the package or package type.
Maybe too much work for now, but something to think of.
Akin https://github.com/microsoft/rushstack/issues/new/choose

webpro:
> Is it an idea to have specific bug report templates for **specific packages/topics** in the repo? That way we can add the prefix for each issue `[@kadena/client] ...` and add questions specific to the package or package type. Maybe too much work for now, but something to think of. Akin https://github.com/microsoft/rushstack/issues/new/choose

Yes good idea. For now I've added a dropdown to the existing template. If we feel the need we can extend to your idea later.

.github/ISSUE_TEMPLATE/001-bug_report.yml
webpro: Oh that was too much copy-pasting from my end. Fixed a few more instances, thanks!

Randynamic (approved):

PR #349, opened 2023-06-01
Recreation of the Card component using Vanilla Extract

bearni95:
## Reason:

Closes ["Recreate the Surface component"](https://app.asana.com/0/1204649083736950/1204713993965161)

## Changes made:
- Recreated the card component under `packages/libs/react-ui` using vanilla-extract
- Added `sized.0` to theme.css.ts
- Added `width['max-content']` to sprinkles.css.ts

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/styles/themes.css.ts
```diff
@@ -47,6 +47,7 @@ export const vars = createGlobalTheme(':root', {
     1: `0px 1px 2px 0 $colors$neutral3`,
   },
   sizes: {
+    0: '0',
```
eileenmguo, line 50: 👍

packages/libs/react-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Body: FC<ICardChildComponentProps>;
+  Footer: FC<ICardChildComponentProps>;
+}
+
+export const Card: ICard = {
+  Container,
+  Heading,
+  Body,
+  Footer,
+};
```
eileenmguo, line 63: This was mentioned as a comment on the last PR, but this component doesn't need subcomponents because the card component should just be a generic component that accepts any children as props. So, you should only need to export `Card`

packages/libs/react-ui/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // TODO: implement with heading tag
+  return <h4>{children}</h4>;
+};
+
+export const Body: FC<ICardChildComponentProps> = ({ children }) => {
+  return <div className={StyledCardBody}>{children}</div>;
+};
+export const Footer: FC<ICardChildComponentProps> = ({ children }) => {
+  return <div className={StyledCardFooter}>{children}</div>;
+};
```
eileenmguo, line 26: These can be removed

packages/libs/react-ui/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <>
+        <Card.Container stack={stack} fullWidth={fullWidth}>
+          <Card.Heading>Getting Started is Simple</Card.Heading>
+          <Card.Body>
+            Learn Kadena&apos;s core concepts & tools for development in 15
+            minutes
+          </Card.Body>
+          <Card.Footer>
+            {/*}
+            <Button title={'Button'} icon={SystemIcons.Information}>
```
eileenmguo, line 49: cleanup

packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,82 @@
...
+            marginY: 'md',
+            borderRadius: 'sm',
+            border: 'none',
+          },
+    ),
+    {
+      selectors: stack ? selectors : {},
+    },
+  ];
+});
```
eileenmguo, line 61: I'm thinking that with vanilla extract the use of variants for booleans seems messier. I would recommend using classnames and styling the default (false styles) as part of the main "style" and with the additional ones optionally added to override the defaults. You can use this library as a utility to do this  https://www.npmjs.com/package/classnames

```
<div className={classNames(base, { stack, fullWidth })}>
```

This way the base styles + the false versions of the stack and fullWidth variants will be applied by default and if the props are true, the other styles will override the default

eileenmguo (commented):
looks good! The only thing is the unexpected package update 👍

common/config/rush/pnpm-lock.yaml
```diff
@@ -6756,7 +6758,7 @@ packages:
     dependencies:
       '@storybook/client-logger': 7.0.0-rc.8
       '@storybook/core-events': 7.0.0-rc.8
-      '@storybook/csf': 0.0.2-next.11
+      '@storybook/csf': 0.1.1-next.0
```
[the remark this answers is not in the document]
bearni95: The changes happened after running `rush-update` to install the classnames package. The fix works like a charm, will keep the code snippet handy for the other PRs that have similar changes!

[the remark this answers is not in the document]
eileenmguo: Nice! glad it worked 👍

packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,43 @@
...
+    flexDirection: 'column',
+    alignItems: 'flex-start',
+    paddingX: 'lg',
+    paddingY: 'md',
+    borderRadius: 'sm',
+    marginY: 'md',
+    border: 'none',
+  }),
+  {
+    width: 'max-content',
```
eileenmguo, line 19: could you add this to the sprinkles now that you've added it to the sprinkles file?

packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,43 @@
...
+  },
+]);
+
+export const fullWidthClass = style([{ width: '100%' }]);
```
eileenmguo, line 23: Just merged a PR that adds this to sprinkles! So you can use the utility class for this too

eileenmguo (commented):
The only thing left is cleaning up the pnpm-lock.yaml file of package upgrades and this should be ready to merge!

packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,41 @@
...
+  }),
+]);
+
+export const fullWidthClass = style([sprinkles({ width: '100%' })]);
+
+export const stackClass = style([
+  {
+    marginTop: 0,
+    marginBottom: 0,
+  },
```
eileenmguo, line 27: these can also be sprinkles

eileenmguo (approved):
This is good to merge after the stack styles are moved to sprinkles 👍

eileenmguo:
Did a `rush update` locally with the `pnpm-lock.yaml` file from master, and it seems like those packages are still getting updated for me. I do think it's pretty isolated so it shouldn't cause issues elsewhere so I think this is safe to merge

common/config/rush/pnpm-lock.yaml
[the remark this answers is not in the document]
alber70g: You can update the packages. I just talked with @eileenmguo to see whether we want to include pre-releases in the installed packages. 
This might be because we identify a prerelease package as the minimal version of the package, signalling we want to have pre-release packages ("^7.0.0-rc.3"). If we change that to not have `-rc.3` it'll only take stable releases.

alber70g:
Copy of my comment: https://github.com/kadena-community/kadena.js/pull/349/files#r1222688519

You can update the packages. I just talked with @eileenmguo to see whether we want to include pre-releases in the installed packages.
This might be because we identify a prerelease package as the minimal version of the package, signalling we want to have pre-release packages ("^7.0.0-rc.3"). If we change that to not have -rc.3 it'll only take stable releases.

PR #358, opened 2023-06-05
[@kadena/react-ui]: Adding icon components

eileenmguo:
## Reason:

Adds the icon components to the new component library 
https://app.asana.com/0/1204649083736950/1204549680288781

NOTE: the icons now share consistent fixed sizing. The foreground color is also applied to both, but it only actually gets applied on the System component due to the current color setting

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Icons/IconWrapper.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const IconWrapper = (
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  Component: React.FC<SVGProps<SVGSVGElement>>,
+): React.FC<IconType> => {
+  const WrappedIcon: React.FC<SVGProps<SVGSVGElement> & IIconProps> = ({
+    size = 'md',
+    ...props
+  }) => (
+    <div className={sizeVariants[size]}>
```
sstraatemans, line 19: this used to be a span.
And it looks like it is still a display:'block'

eileenmguo: good catch

PR #360, opened 2023-06-05
Recreate Grid component using vanilla-extract

bearni95:
## Reason:

Closes ["Recreate Grid component"](https://app.asana.com/0/1204649083736950/1204541876909331/)

- Migrated the grid component (container and item) into `vanilla-extract`
- Added corresponding export options

## Check off the following:

- [x] I have added tutorials.

TODO: Investigate alternative ways to using the style attribute


packages/libs/react-ui/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+    area === undefined
+      ? {
+          gridColumnStart: colStart,
+          gridColumnEnd: colEnd,
+          gridRowStart: rowStart,
+          gridRowEnd: rowEnd,
+        }
+      : {
+          gridArea: area,
+        };
```
eileenmguo, line 34: Is there a reason why you opted to change the way this is implemented from the previous grid component?

packages/libs/react-ui/src/components/Grid/GridContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const styles = {
+    gridTemplateAreas: templateAreas,
+    gridTemplateRows: templateRows,
+    gridTemplateColumns: templateColumns,
+  };
+
+  const classList = classNames(GridContainerClass, gapVariants[spacing]);
+
+  return (
+    <div className={classList} style={styles}>
```
eileenmguo, line 30: I feel like using inline styles is generally looked down upon and I'm also not sure if there would be implications with server components when using this. I'm not sure if there is a better solution though. Do you have thoughts on this @sstraatemans ?

sstraatemans: yeah I would say, not use the style prop and keep it strict

eileenmguo: But, I can't think of an alternative? We don't have the flexibility of dynamic styling with VE since the styles are compiled to static style sheets. I believe other solutions like https://github.com/wayfair/rainbow-sprinkles do use inline styles so maybe this is the best option for this

sstraatemans: ok..as long as there is no style prop it is fine

common/config/rush/pnpm-lock.yaml
```diff
@@ -7763,7 +7765,7 @@ packages:
       '@storybook/channels': 7.0.0-rc.8
       '@storybook/client-logger': 7.0.0-rc.8
       '@storybook/core-events': 7.0.0-rc.8
-      '@storybook/csf': 0.0.2-next.11
+      '@storybook/csf': 0.1.1-next.0
```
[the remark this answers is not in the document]
bearni95: The changes happened after running `rush update` when installing classnames, will keep the extra changes off the commits from here on!

packages/libs/react-ui/src/components/Grid/GridItem.tsx
bearni95: The existing implementation was causing issues for when one set of variables was defined (col and row starts/ends) but the other wasn't (gridarea).

sstraatemans (approved):

packages/libs/react-ui/src/components/Grid/Grid.css.ts
```diff
@@ -0,0 +1,39 @@
+import { vars } from '../../styles';
+
+import { style, styleVariants } from '@vanilla-extract/css';
+
+export const GridContainerClass = style([
```
eileenmguo, line 5: I'm noticing that you are capitalizing the classNames. I don't think this is normal convention. Is there a reason why you decided to do this?

bearni95: Back in stitches, similar objects contained the actual components and went capitalized. I see it makes sense to change to camel-case for the classNames

packages/libs/react-ui/src/components/Grid/Grid.css.ts
```diff
@@ -0,0 +1,40 @@
...
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    '3xl': '3xl',
+  } as Record<string, keyof typeof vars.sizes>,
+  (gap) => {
+    return [
+      gridContainerClass,
+      {
+        gridGap: vars.sizes[gap],
+      },
+    ];
```
eileenmguo, line 38: sprinkles! also doing this inline will likely make this more readable I think (you wont have to write comments and the value is directly next to the style attribute it is being applied to)

packages/libs/react-ui/src/components/Grid/stories.css.ts
```diff
@@ -0,0 +1,19 @@
...
+    padding: 2,
+    color: 'neutral6',
+    display: 'flex',
+  }),
+  {
+    width: '100%',
+    height: '100%',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
```
eileenmguo, line 18: all of these can be applied via sprinkles

alber70g (approved):

packages/libs/react-components/src/components/Input/Input.tsx
```diff
@@ -16,8 +16,8 @@ export interface IInputProps
   > {
   as?: 'input';
   leadingText?: string;
-  leftPanel?: typeof SystemIcons[keyof typeof SystemIcons];
-  rightPanel?: typeof SystemIcons[keyof typeof SystemIcons];
+  leftPanel?: (typeof SystemIcons)[keyof typeof SystemIcons];
+  rightPanel?: (typeof SystemIcons)[keyof typeof SystemIcons];
```
eileenmguo, line 20: Were you able to figure out what was applying this? I've seen other PRs auto-formatting this back

packages/libs/react-ui/package.json
```diff
@@ -71,6 +71,7 @@
     "@types/node": "^16.0.0",
     "@types/react": "^18.0.30",
     "@vanilla-extract/webpack-plugin": "~2.2.0",
+    "classnames": "^2.3.1",
```
eileenmguo, line 74: This is a normal dependency so it can be removed from dev dependencies

packages/libs/react-ui/src/components/Grid/GridContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+  templateColumns,
+}) => {
+  const styles = {
+    gridTemplateAreas: templateAreas,
+    gridTemplateRows: templateRows,
+    gridTemplateColumns: templateColumns,
+  };
+
+  return (
+    <div className={gapVariants[spacing]} style={styles}>
```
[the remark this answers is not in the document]
eileenmguo: Could you still add a comment in the code for this? Just so it's clear that we intend to come back and take a look at this when there is a bit of time to investigate

packages/libs/react-components/src/components/Input/Input.tsx
bearni95: I think I've finally found it! Tried removing my global install of prettier and the change gets reverted upon committing the affected files 🥳🥳 will revert these in the currently open PRs that also got affected by t

eileenmguo (approved):

eileenmguo (approved):

PR #362, opened 2023-06-05
Recreate Stack component unsing vanilla-extract

bearni95:
## Reason:

Closes ["Recreate Stack component"](https://app.asana.com/0/1204649083736950/1204549673395923/f)

- Recreated the stack component using `vanilla-extract`

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Stack/Stack.css.ts
```diff
@@ -0,0 +1,111 @@
...
+export const container = style([
+  sprinkles({
+    display: 'flex',
+    gap: 'md',
+    flexDirection: 'row',
+    justifyContent: 'flex-start',
+    alignItems: 'flex-start',
+    flexWrap: 'nowrap',
+  }),
+  {},
```
eileenmguo, line 14: cleanup

packages/libs/react-ui/src/components/Stack/Stack.css.ts
```diff
@@ -0,0 +1,111 @@
...
+  '2xl': '2xl',
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  '3xl': '3xl',
+};
+
+export const spacingClass = styleVariants(spacingVariants, (gap) => {
+  return [
+    container,
+    {
+      gap: vars.sizes[gap as keyof typeof vars.sizes],
```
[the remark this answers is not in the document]
eileenmguo: Also, I think this mapping is a great tool, but it doesn't need to be used in all situations. For example, the variant implementations in this file are quite simple and may even be more readable and simpler if you were to explicitly write them out (I think in these case it would result in  fewer lines of code). What do you think about the difference in this situation?

eileenmguo (changes requested):

packages/libs/react-ui/src/components/Stack/Stack.css.ts
```diff
@@ -0,0 +1,111 @@
...
+export const ItemSizeClass = styleVariants(vars.sizes, (size) => {
+  return [
+    Item,
+    sprinkles({}),
+    {
+      width: size,
+      height: size,
+    },
+  ];
+});
```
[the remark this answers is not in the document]
bearni95: I felt the same way, and initially placed in in the stories file, but vanilla-extract complains:

```
Styles were unable to be assigned to a file. This is generally caused by one of the following:


- You may have created styles outside of a '.css.ts' context
- You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
```

Would it make sense to define those styles in the stories file and feed them straight via `style` dom parameter?

packages/libs/react-ui/src/components/Stack/Stack.css.ts
```diff
@@ -0,0 +1,103 @@
...
+    'space-between': 'space-between',
+    'space-around': 'space-around',
+  },
+  (justify) => {
+    return [
+      container,
+      {
+        justifyContent: justify,
+      },
+    ];
```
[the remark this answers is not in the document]
bearni95: Sprinkles seems to dislike being fed `justifyContent`, `alignItems`, `flexWrap` nor `flexDirection`. The last two also require the type to be defined, otherwise even without sprinkles ts complains :(

alber70g (approved):

packages/libs/react-ui/src/components/Stack/Stack.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+        justifyContentClass,
+      ) as (keyof typeof justifyContentClass)[],
+      control: { type: 'radio' },
+    },
+    alignItems: {
+      options: Object.keys(alignItemsClass) as (keyof typeof alignItemsClass)[],
+      control: { type: 'radio' },
+    },
+    direction: {
+      control: { type: 'boolean' },
```
eileenmguo, line 32: should have options here

packages/libs/react-ui/src/components/Stack/Stack.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+    },
+    flexWrap: {
```
eileenmguo, line 34: should have options here

eileenmguo (commented):
Missed some storybook updates on my previous review. Once that is done, this would be ready to merge!

eileenmguo (approved):

packages/libs/react-ui/package.json
```diff
@@ -71,6 +71,7 @@
     "@types/node": "^16.0.0",
     "@types/react": "^18.0.30",
     "@vanilla-extract/webpack-plugin": "~2.2.0",
+    "classnames": "^2.3.1",
```
eileenmguo, line 74: Didn't see this before! This is already a normal dependency so it can be removed here

PR #350, opened 2023-06-02
Add code of conduct

webpro:
Add `CODE_OF_CONDUCT.md` - this inital version is a copy of https://www.contributor-covenant.org

[View CODE_OF_CONDUCT.md](https://github.com/kadena-community/kadena.js/blob/d8c4f528c25303a33d8504ba914258020dfc81f1/CODE_OF_CONDUCT.md)

Still needs a contact method (e.g. email address) for complaints, etc.

PR #372, opened 2023-06-08
Add existing coc

webpro:
Supersedes #350. I'm not seeing significant missing elements in comparison, except for a generic contact option (see below). So I'll close that one.

This is a copy of the draft at https://docs.google.com/document/d/1SU9OKkobYpq0R5vDK1MeYxiz8RpXxYmaceF3UMtzjXQ

- [x] Do we need HR or legal to review?
- [x] This misses contact options (i.e. email address). The original draft targeted Discord, where you can PM a moderator. GitHub (and other platforms) don't have that option.

### DIFF
--- a/packages/apps/kadena-docs/src/data/menu.mjs
+++ b/packages/apps/kadena-docs/src/data/menu.mjs
@@ -719,6 +719,18 @@ export const menuData = [
         "isActive": false
       },
       {
+        "children": [],
+        "root": "/docs/contribute/code-of-conduct",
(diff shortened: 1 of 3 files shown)


Randynamic (approved):

Randynamic:
For the contact info, I think best to keep it on Discord

webpro:
> For the contact info, I think best to keep it on Discord

There should be a private way to contact. I know Discord has DMs, but not everyone is on Discord.

PR #374, opened 2023-06-08
[@kadena/client] Cont Command Client Support 

nil-amrutlal-dept:
## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.


packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,477 @@
...
+
+/**
+ * Used to build Command objects modularly by adding pact code, environment data, capabilities, and sigs
+ * as necessary.
+ * Once the command is complete and ready for use, you can either call `createCommand` to get the ICommand object,
+ * or you can call `local` or `send` to send the command to /send or /local endpoints of a node with the provided URL.
+ *
+ * @alpha
+ */
+export class ContCommand
```
javadkh2, line 95: This class has a lot in common with the PactCommand class, maybe you can extend the class instead of copying those functions

javadkh2:
I think you also need to update pactCreator  to use ContCommand class as well if the user use `cont` for calling the function then it should create a ContCommand object. 
it seems for exec command you can use the PactCommand class. 
so the api would be like this
```javascript
Pact.module.coin['transfer-crosschain'].exec("from" "to" 10) // this should return PactCommand instance with (transfer-crosschain "from" "to" 10) as code

Pact.module.coin['transfer-crosschain'].const(pactId, step, rollback, proof) // this should return ContCommand instance
```

packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,477 @@
...
+  public cmd: string | undefined;
+  public nonce: NonceType | undefined;
+  public proof: string;
+  public step: number;
+  public pactId: string;
+  public rollback: boolean;
+  public status: TransactionStatus;
+  public requestKey: string | undefined;
+
+  public constructor() {
```
javadkh2, line 125: I think its better to pass (proof, step, pactId, rollback) to the constructor, then you don't need those related setters.
and the API would be more cleaner because we need to set those properties in one step

packages/libs/client/src/pact.ts
```diff
@@ -488,10 +491,17 @@ const pactCreator = (): IPact => {
       that: unknown,
       args: Array<string | number | boolean>,
     ) {
+      if (code.endsWith('.cont')) {
+        const transaction: ContCommand = new ContCommand(
```
javadkh2, line 495: you need to return the transaction here

packages/libs/client/src/contPact.ts
alber70g: Or pull stuff to an abstract class and let both extend from it

packages/libs/client/src/contPact.ts
alber70g, line 1: Not sure if `contPact` is the right name. I'd name it `contBuilder`, or something more descriptive.

packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Create a command that's compatible with the blockchain
+   * @returns a command that can be send to the blockchain
+   * (see https://api.chainweb.com/openapi/pact.html#tag/endpoint-send/paths/~1send/post)
+   */
+  public createCommand(): IUnsignedCommand {
+    const dateInMs: number = Date.now();
+
+    // convert to IUnsignedTransactionCommand
+    const unsignedTransactionCommand: ICommandPayload = {
```
alber70g, line 64: Does the comment match the code?

packages/libs/client/src/pact.ts
```diff
@@ -17,10 +17,12 @@ import {
   ISignatureJson,
   IUnsignedCommand,
   PactValue,
+  Type,
 } from '@kadena/types';
 
 import { IPactCommand } from './interfaces/IPactCommand';
 import { parseType } from './utils/parseType';
+import { ContCommand } from '.';
```
alber70g, line 25: As we're in the root, can we import directly from the file itself instead of from the barrel file?

packages/libs/client/src/pact.ts
```diff
@@ -168,6 +170,7 @@ export class PactCommand
     this.status = 'malleable';
     this.requestKey = undefined;
     this.nonce = undefined;
+    this.type = 'exec';
```
alber70g, line 173: On both line 149/151 and 173 it's set to 'exec'. Is this one redundant?

alber70g (commented):
General comment.
I see the use of `@kadena/types` which is used in other places, prior to this. 
I'm becoming less a fan of this. I'd rather have the source describe the types and know where it comes from.

It's either from a function (function/class/etc) that requires a certain datastructure, OR it's an API that describes this. Therefore the goal is to make the different api endpoints a wrapper in the repo with it's types. Preferably generated from a spec (openapi/swagger/...). 

For this PR I'd like to have the types described and exported from the place where they're required and referenced from that place when they are required somewhere else.

alber70g (commented):
General comment.
I see the use of `@kadena/types` which is used in other places, prior to this. 
I'm becoming less a fan of this. I'd rather have the source describe the types and know where it comes from.

It's either from a function (function/class/etc) that requires a certain datastructure, OR it's an API that describes this. Therefore the goal is to make the different api endpoints a wrapper in the repo with it's types. Preferably generated from a spec (openapi/swagger/...). 

For this PR I'd like to have the types described and exported from the place where they're required and referenced from that place when they are required somewhere else.

@javadkh2, what's your opinion?

packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,99 @@
...
+import { log } from 'debug';
+
+/**
+ * @alpha
+ */
+export interface IContCommandBuilder<
+  TCaps extends Record<string, TArgs>,
+  TArgs extends Array<TCaps[keyof TCaps]> = TCaps[keyof TCaps],
+> {
+  createCommand(): IUnsignedCommand;
```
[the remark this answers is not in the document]
javadkh2: based on the API (https://api.chainweb.com/openapi/pact.html#tag/model-payload) tit seems that the only distinction between "cont" and "exec" is the payload. so I think it also accept capabilities.

packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,197 @@
...
+    },
+  });
+
+  if (proofResponse === undefined) {
+    throw new Error('Unable to obtain SPV Proof');
+  }
+
+  const proof = await proofResponse.text();
+
+  const contCommand = new ContCommand(proof, 1, requestKey, false);
```
javadkh2, line 194: I think the step could be something different, if the defpact has more than 2 steps , so maybe its better to accept it as an input

packages/libs/types/src/PactCommand.ts
```diff
@@ -163,6 +163,10 @@ export type ChainId =
   | '17'
   | '18'
   | '19';
+/**
+ * @alpha
+ */
+export type Type = 'cont' | 'exec';
```
alber70g, line 169: Can we move this to the source where it's required? I think it should be in packages/libs/client/src/pact.ts

Ultimately, I want to get rid of @kadean/types altogether. The types should belong to where they're required, and describe an interface for a function or class. Not just be somewhere for general use

nil-amrutlal-dept: Fixed this on:
https://github.com/kadena-community/kadena.js/pull/413

nil-amrutlal-dept:
Closing PR: another branch containing the same functionality was merged https://github.com/kadena-community/kadena.js/pull/413

PR #375, opened 2023-06-08
 Recreate IconButton component using vanilla-extract

bearni95:
## Reason:
Closes ["Recreate Icon Button component"](https://app.asana.com/0/1204649083736950/1204549680288787/f)

- Migrated the IconButton component into vanilla-extract

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+  sprinkles({
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    borderRadius: 'lg',
+    cursor: 'pointer',
+    size: 11,
+  }),
+  {
+    backgroundColor: fallbackVar(surfaceColor, 'transparent'),
```
eileenmguo, line 24: I like this fallbackVar feature!

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+    color: fallbackVar(highContrastColor, vars.colors.neutral5),
+    border: 0,
```
eileenmguo, line 26: I believe this can be a sprinkle

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+    transition: 'opacity .2s ease',
+    selectors: {
+      '&:hover': {
+        opacity: '.6',
+      },
+      '&:focus-visible': {
+        outlineOffset: '2px',
+        outline: '2px solid $$svgColor',
```
eileenmguo, line 34: I think this is leftover from stitches?

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+    positive: 'positive',
+    warning: 'warning',
+    negative: 'negative',
+  },
+  (color) => {
+    return [
+      container,
+      {
+        vars: {
+          [surfaceColor]: vars.colors[`${color as ColorType}Surface`],
```
eileenmguo, line 59: I don't think this works for inverted?

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+          [highContrastColor]:
+            color === 'inverted'
+              ? vars.colors.neutral3
+              : vars.colors[`${color as ColorType}HighContrast`],
```
eileenmguo, line 63: I like the fallbackVar feature, but I think that the usage of vars in general should be used only when necessary. So surface color could be applied as just a sprinkle (if I understand correctly) since it doesn't require any selectors. This is because when you assign a new value to a `var`, VE is required to create a new class to do that.
https://vanilla-extract.style/documentation/api/create-var/#setting-the-variable

In this instance you will need to generate custom styles anyways for the selector so the result is negligible, but the usage of vars should be a conscious decision since we want to avoid adding additional classes to compile where necessary. It's not necessary from a results point of view to change this here, but I would recommend changing it anyway to help practice thinking about this concept and as a reference for others who look at this code!

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
bearni95: Sprinkles doesn't seem to like the variables created via `createVar` for backgroundColor nor color.

Border was good to move into a sprinkle

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,68 @@
...
+        outlineOffset: '2px',
+        outline: '2px solid $$svgColor',
+      },
+    },
+  },
+]);
+
+globalStyle(`${container} > span > svg`, {
+  color: fallbackVar(highContrastColor, vars.colors.neutral5),
+});
```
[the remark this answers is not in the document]
bearni95: Vanilla Extract doesn't allow to style with selectors from the parent! If styling from the `container` var, the following error message throws:

```
Error: Invalid selector: & > span > svg

Style selectors must target the '&' character (along with any modifiers), e.g. `${parent} &` or `${parent} &:hover`.

This is to ensure that each style block only affects the styling of a single class.

If your selector is targeting another class, you should move it to the style definition for that class, e.g. given we have styles for 'parent' and 'child' elements, instead of adding a selector of `& ${child}`) to 'parent', you should add `${parent} &` to 'child').

If your selector is targeting something global, use the 'globalStyle' function instead, e.g. if you wanted to write `& h1`, you should instead write 'globalStyle(`${parent} h1`, { ... })'
```

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,67 @@
...
+    borderRadius: '$lg',
+    cursor: 'pointer',
+    size: '$11',
+    border: 'none',
+  }),
+  {
+    transition: 'opacity .2s ease',
+    color: fallbackVar(contrastColor, vars.colors.$neutral5),
+    backgroundColor: fallbackVar(surfaceColor, 'transparent'),
+    selectors: {
```
eileenmguo: Should move this to sprinkles

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -0,0 +1,67 @@
...
+  return [
+    container,
+    {
+      vars: {
+        [contrastColor]: vars.colors[`$${color}Contrast`],
+        [surfaceColor]: vars.colors[`$${color}Surface`],
+      },
+    },
+  ];
+});
```
eileenmguo: We should be able to add the default and inverted variants here as well

packages/libs/react-ui/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+      className={className}
+      onClick={onClick}
+      aria-label={ariaLabel}
+    >
+      <Icon size="md" />
+    </button>
+  ) : (
+    <a className={className} href={href} aria-label={ariaLabel}>
+      <Icon size="md" />
+    </a>
```
eileenmguo: To keep the jsx cleaner, I would generally try to avoid larger ternaries like this if there is another option

eileenmguo (approved):

PR #379, opened 2023-06-09
[@kadena/react-ui]: Recreating Heading and Text components with VE

eileenmguo:
## Reason:

Recreates the Heading and Text components in the @kadena/react-ui

- Adds VE recipes to implement the styles. Also tested this in a consuming app with RSC (react server components) and the runtime involved with finding the correct classname to apply the styles does not interfere with usage in server components

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,54 @@
+import {
```
sstraatemans, line 1: missing the gradientText component for Typography

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
```diff
@@ -0,0 +1,131 @@
+/* eslint @typescript-eslint/naming-convention: 0, @kadena-dev/typedef-var: 0 */
```
sstraatemans, line 1: missing an index file

packages/libs/react-ui/src/index.ts
```diff
@@ -4,9 +4,17 @@ export {
   IIconProps,
   SystemIcon,
   ProductIcon,
+  Text,
```
sstraatemans, line 7: missing the HEader export

sstraatemans (approved):

isaozler (approved):

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
alber70g: Why do we need this? I prefer not globally disabling things for a whole file, but using `eslint-disable-next-line` or something similar. If this is a structural problem, we should modify the `@kadena-dev/eslint-config`

alber70g (approved):

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
eileenmguo: I just opened another PR to update the @kadena-dev/eslint-config for typedef-var and am going to look into addressing the naming-convention one today 👍

PR #382, opened 2023-06-09
[@kadena/react-ui]: Importing global styles

eileenmguo:
## Reason:

We need to be able to import global styles to the consuming application. The common way of importing it via a path to the file within the package wasn't working in this case so I thought of the following options:
- Add creating a global styles file as part of the setup instructions on the readme
- Importing the global styles file in the index file

Went the the later approach although I don't know if there are any repercussions to this strategy. I think this is fine for now, but let me know if you have any other ideas!

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/libs/react-ui/.storybook/preview.ts
+++ b/packages/libs/react-ui/.storybook/preview.ts
@@ -1,7 +1,6 @@
 import { darkThemeClass } from '../src/styles';
 import type { Preview } from '@storybook/react';
 import { themes } from '@storybook/theming';
-import '../src/styles/global.css';
 
 const preview: Preview = {
   parameters: {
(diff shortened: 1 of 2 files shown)


sstraatemans:
so I just need to import the globalstyles sheet? 
versus the old UI package, where I create one.
Can I still create my own globalstyles?

eileenmguo:
> so I just need to import the globalstyles sheet? versus the old UI package, where I create one. Can I still create my own globalstyles?

So you actually don't need to import the global.css file yourself since it's imported in index. I believe if you use anything exported from that index file, global styles will be applied. If you want to apply your own global styles you can just great your own css file and use vanilla extract's global styles function just like they are used within the library

PR #384, opened 2023-06-12
Add remark plugins

webpro:
As Markdown usage in our repo increases, I'd like to install more utilities to help writing and generating it. Using Prettier is great for formatting, but we can do more.

The [Unified](https://unifiedjs.com)/[remark](https://remark.js.org) ecosystem is what powers many websites that felt the same need, such as the Prettier and Node.js websites.

For starters, we will need a Markdown parser and plugins that support YAML front matter and GFM (GitHub Flavored Markdown). And a plugin that uses our Prettier configuration and formats it like we are used to. And then the fun begins with more plugins:

- `remark-reference-links` converts `[](https://...)` links to `[][0]` refs, this is great for readability.
- `remark-order-reference-links` orders the link refs from the previous plugin, so you don't have to.

This PR also adds a custom plugin:

- `handleCommentMarkers` to make it easy to add functions to replace whatever you want between comment markers (that look like `<!--myPlugin [...] -->`).

And the first of them is a `packageTable` function to generate the packages table in the root `README.md`. This PR also updates that table.

Next step is to make it easier/automate formatting Markdown, but for now we can do with the command that's documented in the `README.md` of this new `remark-plugins` in `packages/tools`.

This tool uses "JSDoc-style TypeScript" so it doesn't require a build step (and most of that ecosystem does the same).

---
  - https://app.asana.com/0/0/1204741076546821


sstraatemans:
already using `GFM` package, so good to have it general.
wasnt a fan of the `handleCommentMarkers` because then it could be anything and that is hard to give a specific style I think (maybe using it wrong).

alber70g (commented):
Is there a requirement for `remark-plugins` to have the `index.js` in the root and not have a `src` directory? Otherwised I'd like to make it uniform with the other packages and have that.

Also, would it be possible to have it in Typescript?

webpro:
> already using `GFM` package, so good to have it general. wasnt a fan of the `handleCommentMarkers` because then it could be anything and that is hard to give a specific style I think (maybe using it wrong).

Yet not sure I understand, The "anything" is quite specific: it must not have a space and the name must be an exported function and on the `commentMarkers` object (`<!--myPlugin`). Or can you elaborate?

> But maybe that is just specific to docs.

Yeah this is actually initially meant for everything outside the docs, although it might be a good idea to use a subset or alternative set of plugins maybe.

webpro:
> Is there a requirement for `remark-plugins` to have the `index.js` in the root and not have a `src` directory? Otherwised I'd like to make it uniform with the other packages and have that.
> 
> Also, would it be possible to have it in Typescript?

They're related things. Not using a build step is nice and it's still typed (although with JSDoc). Eventually it's just internal tooling. Any chance we could iterate fast? Because I also didn't want to jump through the Rush hoops for this one just yet.

webpro:
Not sure what would be faster: discuss or do, so added TypeScript + ESLint

Randynamic (approved):

alber70g (approved):

packages/tools/remark-plugins/tsconfig.json
alber70g, line 1: Can we use the base tsconfig to extend from?

PR #391, opened 2023-06-12
Add first version of "supported environments" ADR

webpro:
Would be good to record the environments we support. I found it a bit difficult to scope and word it properly.

For instance, can we keep it concise and simply include supported browsers in this document as well, does that make sense?

So please feel free to comment if something is not clear, or you need more context of something, etc.

### DIFF
--- a/docs/decisions/0002-supported-environments.md
+++ b/docs/decisions/0002-supported-environments.md
@@ -1,0 +1,51 @@
+# Supported Environments
+
+Date: 2023-06-12
+
+Status: proposed
+
+## Context
+
+The tools and code **in this monorepo** run on various operating systems and
+environments, including Node.js and browsers.
+
+It's not feasible to support every operating system and every version of every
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-supported-environments.md
```diff
@@ -0,0 +1,51 @@
...
+work, but bugs or feature requests will have low priority.
+
+### GitHub Actions
+
+We run tests in GitHub Actions. We run tests in `ubuntu-latest`, since that
+covers most use cases and is fast. It's also the most-used distro for WSL.
+
+### Browsers
+
+The documentation website, tools and tutorials should work in... TODO "latest 2
```
Randynamic, line 43: - Firefox latest 2
- Chrome latest 2
- Safari (iOS) latest 2
- Edge latest 2
- Brave latest 2

Randynamic (approved):

PR #396, opened 2023-06-12
[@kadena/react-ui]: Changing token naming convention and lint cleanup

eileenmguo:
## Reason:

This was prompted by the need to disable several lint errors. This includes: 
- Lint error cleanup for @kadena-dev/typevar-def
- And changing tokens to be prefixed with $ so that they don't trigger the typescript naming convention lint rule

Prefixing tokens with $
Advantages:
- It makes it more clear when you are assigning a custom variable
Disadvantages:
- There are still instances in which we need to disable the lint errors. For example when creating variants for a components spacing, we don't want to have to prefix the associated options with $ as well. 

Alternative: Disabling the `@typescript-eslint/naming-convention` lint rule in React projects

Curious to hear your thoughts and open to alternative suggestions!

## Check off the following:

- [x] I have added tutorials.

### DIFF
--- a/packages/libs/react-ui/src/components/Icons/IconWrapper.css.ts
+++ b/packages/libs/react-ui/src/components/Icons/IconWrapper.css.ts
@@ -4,12 +4,12 @@ import { style, styleVariants } from '@vanilla-extract/css';
 
 export const iconContainer = style([
   sprinkles({
-    color: 'foreground',
+    color: '$foreground',
   }),
 ]);
 
 export const sizeVariants = styleVariants({
-  sm: [iconContainer, sprinkles({ size: 4 })],
+  sm: [iconContainer, sprinkles({ size: '$4' })],
   md: [iconContainer, sprinkles({ size: 6 })],
   lg: [iconContainer, sprinkles({ size: 8 })],
   xl: [iconContainer, sprinkles({ size: 10 })],
--- a/packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
+++ b/packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
@@ -13,68 +13,68 @@ export const elementVariants = {
   h1: [
     sprinkles({
       fontSize: {
-        xs: '5xl',
-        md: '7xl',
-        lg: '9xl',
-        xl: '10xl',
-        '2xl': '12xl',
+        xs: '$5xl',
+        md: '$7xl',
+        lg: '$9xl',
+        xl: '$10xl',
+        xxl: '$12xl',
       },
     }),
   ],
   h2: [
     sprinkles({
-      fontWeight: 'bold',
+      fontWeight: '$bold',
       fontSize: {
-        xs: '4xl',
-        lg: '6xl',
-        xl: '8xl',
-        '2xl': '9xl',
+        xs: '$4xl',
+        lg: '$6xl',
+        xl: '$8xl',
+        xxl: '$9xl',
       },
     }),
   ],
   h3: [
     sprinkles({
       fontSize: {
-        xs: '2xl',
-        lg: '4xl',
-        xl: '5xl',
-        '2xl': '6xl',
+        xs: '$2xl',
+        lg: '$4xl',
+        xl: '$5xl',
+        xxl: '$6xl',
       },
     }),
   ],
   h4: [
     sprinkles({
       fontSize: {
-        xs: 'xl',
-        lg: '2xl',
-        xl: '3xl',
-        '2xl': '4xl',
+        xs: '$xl',
+        lg: '$2xl',
+        xl: '$3xl',
+        xxl: '$4xl',
       },
     }),
   ],
   h5: [
     sprinkles({
       fontSize: {
-        xs: 'lg',
-        xl: 'xl',
-        '2xl': '4xl',
+        xs: '$lg',
+        xl: '$xl',
+        xxl: '$4xl',
       },
     }),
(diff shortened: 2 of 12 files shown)


packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
```diff
@@ -13,68 +13,68 @@ export const elementVariants = {
...
   h2: [
     sprinkles({
-      fontWeight: 'bold',
+      fontWeight: '$bold',
       fontSize: {
-        xs: '4xl',
-        lg: '6xl',
-        xl: '8xl',
-        '2xl': '9xl',
+        xs: '$4xl',
```
EnoF, line 28: why do we have `$4` as well as `$4xl`?

EnoF: when to use `$4` vs `$md`?

packages/libs/react-ui/src/components/Icons/IconWrapper.css.ts
```diff
@@ -4,12 +4,12 @@ import { style, styleVariants } from '@vanilla-extract/css';
...
 export const iconContainer = style([
   sprinkles({
-    color: 'foreground',
+    color: '$foreground',
   }),
 ]);
 
 export const sizeVariants = styleVariants({
-  sm: [iconContainer, sprinkles({ size: 4 })],
+  sm: [iconContainer, sprinkles({ size: '$4' })],
```
webpro, line 12: Maybe forgot dollars below?

packages/libs/react-ui/src/components/Typography/typography.css.ts
```diff
@@ -1,20 +1,20 @@
...
   uppercase: [sprinkles({ textTransform: 'uppercase' })],
   lowercase: [sprinkles({ textTransform: 'lowercase' })],
   capitalize: [sprinkles({ textTransform: 'capitalize' })],
   none: [sprinkles({ textTransform: 'none' })],
 };
 
-export const colorVariants = {
-  default: [sprinkles({ color: 'neutral4' })],
-  emphasize: [sprinkles({ color: 'neutral6' })],
+export const colorVariants: Record<string | number, ComplexStyleRule> = {
```
webpro, line 17: Can you give an example of when type `number` is need?

webpro (approved):

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
eileenmguo: So spacing tokens have the following naming conventions:
- Numeric - the number represents the number * 4 (our design system is built on units of 4). This size is fixed 
- xs, sm, md, etc.. - which are spacing variables that change depending on the breakpoints. These are the values that are used in layout components, but can be used anywhere

Fontsizes have a relative size based naming convention so they would have the following result assuming that the default browser fontsize (16px) is not changed:
$sx = 12px
$sm = 14px
$base = 16px
$md = 18px
$lg  = 20px
...etc 
$12xl = 84px

I can see how this is a bit confusing; this is just the naming convention that I felt worked the best. Does this make sense or do you have any other suggestions?

packages/libs/react-ui/src/components/Icons/IconWrapper.css.ts
eileenmguo: yes! good catch

packages/libs/react-ui/src/components/Typography/typography.css.ts
eileenmguo: I just copied over the libraries prop type, but there is no need for the number in this case. I will remove it 👍

packages/libs/react-ui/src/components/Icons/IconWrapper.css.ts
eileenmguo: yes! I somehow didn't save them before. good catch 👍

sstraatemans (approved):

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
EnoF: I understand the numeric and md/sm etc, thanks. I still wonder what `$4xl` falls into though.

EnoF (approved):

packages/libs/react-ui/src/components/Typography/Heading/Heading.css.ts
eileenmguo: `$4xl` would fall into the fontSizes category since they are relative to the base. The options would be 
```
  fontSizes: {
    xs: '0.75rem', // 12px
    sm: '0.875rem', // 14px
    base: '1rem', // 16px
    md: '1.123rem', // 18px
    lg: '1.25rem', // 20px
    xl: '1.5rem', // 24px
    '2xl': '1.75rem', // 28px
    '3xl': '2rem', // 32px
    '4xl': '2.25rem', // 36px
    '5xl': '2.5rem', // 40px
    '6xl': '2.75rem', // 44px
    '7xl': '3rem', // 48px
    '8xl': '3.25rem', // 52px
    '9xl': '3.75rem', // 60px
    '10xl': '4.5rem', // 72px
    '11xl': '5rem', // 80px
    '12xl': '5.25rem', // 84px
  },
```

Maybe in the future if we notice we are only using certain ones, we can decrease the options

PR #355, opened 2023-06-02
Notification component vanilla extract rewrite [old]

bearni95:
## Reason:

Closes ["Recreate the Notifications Component"](https://app.asana.com/0/1204649083736950/1204713993965158/)

## Changes made (preferably with images/screenshots):
- [x] Rewrite the core of the notification component using `vanilla-extract`
- [x] Pass down "footer" as parameter (blue)
- [ ] Pass down "header" as parameter (red)
img
- [ ] Support for notification-icon and close-icon
 
## Check off the following:

- [ ] I have reviewed my changes and run the appropriate tests.
- [ ] I have have run `rush change` to add the appropriate change logs.
- [ ] I have added/edited docs.
- [ ] I have added tutorials.
- [ ] I have double checked and DEFINITELY added docs.

PR #398, opened 2023-06-12
 Notification component vanilla extract rewrite

bearni95:
## Reason:
Some strange changes leaked into [the original PR for this issue](https://github.com/kadena-community/kadena.js/pull/355). 
Closes ["Recreate the Notifications Component"](https://app.asana.com/0/1204649083736950/1204713993965158/)

- [x] Rewrite the core of the notification component using `vanilla-extract`
- [x] Pass down "footer" as parameter (blue)
- [x] Pass down "header" as parameter (red)
img
- [x] Support for notification-icon and close-icon
 
## Check off the following:

- [x] I have added tutorials.

## Comments 
- [x] [Remove unrelated changes from PR](https://github.com/kadena-community/kadena.js/pull/355#discussion_r1223992421)
- [x] General code cleanup
- [x] [Turn block into Sprinkles](https://github.com/kadena-community/kadena.js/pull/355#discussion_r1224004280)
- [x] [Prevent type casting](https://github.com/kadena-community/kadena.js/pull/355#discussion_r1224006372)
- [x] [Move marginLeft block into sprinkles](https://github.com/kadena-community/kadena.js/pull/355#discussion_r1224007000)
- [x] [Turn header and footer into subcomponents](https://github.com/kadena-community/kadena.js/pull/355#discussion_r1224011060)


packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,141 @@
...
+              : vars.colors[`${color as ColorType}Contrast`],
+          [surfaceColor]:
+            color === 'default'
+              ? vars.colors.neutral2
+              : vars.colors[`${color as ColorType}Surface`],
+        },
+      },
+    ];
+  },
+);
```
eileenmguo, line 100: Reminders: 
- Use sprinkles where you can! So for color and background color you can use sprinkles
- Type the colors when creating the object so that you don't have to cast them to a specific type later https://dev.to/ylerjen/typescript-cast-is-a-type-breaker-1jbh

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,141 @@
...
+export const iconContainerClass = style([
+  sprinkles({
+    marginX: 'md',
+    marginTop: 'sm',
+    display: 'flex',
+    right: 0,
+  }),
+  {
+    position: 'relative',
+    // marginLeft:'auto'
```
eileenmguo, line 111: Clean up comments throughout the PR. Also I believe position should be a sprinkle already, but if not, it should be added

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,141 @@
...
+  },
+]);
+
+export const iconContainerFullWidthClass = style([
+  sprinkles({
+    marginTop: 'sm',
+    marginRight: 'md',
+  }),
+  { marginLeft: 'auto' },
```
eileenmguo, line 120: we should probably also move "auto" to sprinkles for all margin applications

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,141 @@
...
+    padding: 0,
+    width: 'min-content',
+    display: 'flex',
+    justifyContent: 'flex-end',
+    alignItems: 'center',
+    marginRight: 'md',
+  }),
+  {
+    position: 'absolute',
+    right: '100%',
```
eileenmguo, line 135: I think both of these should be moved to sprinkles

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,141 @@
...
+  },
+]);
+
+globalStyle(`${headerContainerClass} > span`, {
```
eileenmguo, line 139: Should this just be applied using padding on the headerContainerClass?

eileenmguo (commented):
Not sure what things are leftover from the previous PR, but I would suggest going over the following checklist:
- Move styles to sprinkles whenever possible. If you're using a styling value that is very common and there isn't an existing sprinkle, think about whether to create a sprinkle for it
- Check that you are not using type casting
- Check for commented out lines that can be cleaned up

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,127 @@
...
+  }),
+]);
+
+export const expandClass = style([
+  sprinkles({
+    width: '100%',
+  }),
+  {
+    width: '100% !important',
+  },
```
eileenmguo, line 43: Is this necessary?

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,127 @@
...
+    justifyContent: 'flex-start',
+    alignItems: 'flex-start',
+    width: 'max-content',
+    height: 'min-content',
+    position: 'relative',
+  }),
+  {
+    border: `1px solid ${fallbackVar(contrastColor, vars.colors.$neutral6)}`,
+    color: fallbackVar(contrastColor, vars.colors.$neutral6),
+    backgroundColor: fallbackVar(surfaceColor, vars.colors.$neutral2),
```
eileenmguo, line 26: Both of these you can still move to sprinkles, this way you won't have to use var!

packages/libs/react-ui/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,100 @@
...
+  simple,
+  expand = false,
+}) => {
+  const isSimple = (simple as boolean) || children === undefined;
+  const Icon = icon || SystemIcon.HelpCircle;
+
+  return (
+    <div
+      className={classNames(
+        [color === 'default' ? containerClass : colorVariants[color!]],
```
eileenmguo, line 45: What is the purpose of the ! in colorVariants[color!]?

packages/libs/react-ui/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {
+          [expandClass]: expand,
+          [simpleClass]: isSimple,
+        },
+      )}
+    >
+      <div className={iconContainerClass}>
+        <Icon size={'md'} />
+      </div>
```
eileenmguo, line 54: Let's take a look together on how we could improve the structure of styling! I think the structure of this component could be simplified by some adjustments

packages/libs/react-ui/src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,100 @@
...
+      )}
+    </div>
+  );
+};
+
+export interface INotificationFooterProps {
+  children: React.ReactNode;
+}
+
+export const NotificationFooter: FC<INotificationFooterProps> = ({
```
eileenmguo, line 83: As discussed in a meeting with the UI team, we want to use the sub component naming convention. So for example, when using these together, we would reference the components as follows: Notification, Notification.Footer, Notification.Header

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+    expand: false,
+    color: undefined,
+  },
+  render: ({
+    selectIcon,
+    title,
+    displayCloseButton: displayCloseButton,
+    expand,
+    color,
+    simple,
```
eileenmguo, line 87: Let's try to come up with naming convention for booleans so that it is clear. I was thinking the "comparative" form of the word: so similar to `disabled`

In this case it would be "expanded" and "simplified". What do you think?

packages/libs/react-ui/src/components/Notification/Notification.css.ts
eileenmguo: Discussed online: Can use variants for this so we don't need to use important

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -0,0 +1,146 @@
...
+]);
+
+export const expandVariants = styleVariants(
+  {
+    true: true,
+    false: false,
+  },
+  (expand) => {
+    return [
+      containerClass,
```
eileenmguo, line 39: You only need to add this class if it isn't already being applied, but in this case it is getting applied twice because it is in both variants. How about we come up with a convention that we don't include them in variants moving forward is it is very clear what classes we are applying to elements in the jsx?

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+        icon={icon}
+        simplified={simple}
+        expanded={expand}
+        color={color}
+        title={title}
+        displayCloseButton={displayCloseButton}
+      >
+        <Notification.Header>
+          <SystemIcon.Link size="md" />
+        </Notification.Header>
```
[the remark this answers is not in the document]
eileenmguo: @isaozler Will there be specific header actions for all notifications, for example will there only be 3 specific options? The reason i ask is because it might make sense to move the close icon to the HeaderActions instead of having it separate and having a prop for actions if it doesn't need to be that dynamic

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+    expanded: expand,
+    color,
+    simplified: simple,
+  }) => {
+    const icon = SystemIcon[selectIcon];
+    return (
+      <Notification.Container
+        icon={icon}
+        simplified={simple}
+        expanded={expand}
```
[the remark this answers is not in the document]
bearni95: The renaming function on VSCode reassigned the variables in the lines above:

https://github.com/kadena-community/kadena.js/blob/c8e97915b0691144190cc48823150833adeeb5d0/packages/libs/react-ui/src/components/Notification/Notification.stories.tsx#L133

While it does work with the reassignation (`expanded: expand`) getting rid of the assignation will make code prettier

packages/libs/react-ui/src/components/Notification/Notification.css.ts
bearni95: I agree, it makes more sense to apply the `containerClass` directly into the component, when used in multiple variants

eileenmguo:
Just checked the story and it seems that the contents of the notification can't be changed so I can't test the behavior when the content is really long. I would assume we want a maximum width when it is not full screen to keep things readable. I thinking we can put a max character width on it? Also the notifications component should be within the component section in storybook!

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <div className={classList}>
+      <div className={iconContainerClass}>
+        <Icon size={'md'} />
+      </div>
+
+      <div>
+        <p className={contentClass}>
+          <h4 className={cardTitleClass}>{title}</h4>
```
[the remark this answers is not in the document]
eileenmguo: I see that the visual behavior is actually what we would expect, but actually not rendering this will keep the html cleaner!

packages/libs/react-ui/src/components/Notification/NotificationButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const NotificationButton: FC<INotificationButtonProps> = ({
+  icon,
+  color,
+  children,
+}) => {
+  const Icon = icon;
+  return (
+    <button className={colorVariants[color]}>
+      {children}
+      <div className={iconContainerClass}>
```
eileenmguo, line 21: When wrapping svgs, you should use inline elements like spans to make the html more semantically correct. Divs are meant to be used as block elements to group more items together, but this icon is meant to be inline with the text so a span makes more sense in this case and in most cases

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const classList = classNames(
+    containerClass,
+    colorVariants[color],
+    expandVariants[expanded ? 'true' : 'false'],
+  );
+
+  return (
+    <div className={classList}>
+      <div className={iconContainerClass}>
```
eileenmguo, line 45: same here

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <p>{children}</p>
+        </p>
+      </div>
+
+      <span
+        onClick={onClose}
+        className={expanded ? iconContainerExpandedClass : iconContainerClass}
+      >
+        {hasCloseButton && <SystemIcon.Close size={'md'} />}
+      </span>
```
eileenmguo, line 61: I'm confused by this. Should this be a button?

bearni95: It does make sense to make it a button, although it won't be aligned with the other icon, nor the title:
notification-icon-button

should we go with it?

bearni95: I actually mean it's alignment to the top and right ends of the card, as the `IconButton` component defines their own sizes/margins

eileenmguo: Ahh yes, in this case I also think it makes sense to give it custom styles rather than reusing the IconButton. The main point is that it needs to be a `button` element so for accessibility. Right now you can tab to the close button to close it because it just uses an onclick handler. It does look like the color is also off so that would be a good thing to fix as well

bearni95: I see!

Ended up re-using the same button component, `NotificationButton`, as the one used for the bottom actions, with the possibility to not include text. Looks alright and aligns properly 🥳

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -0,0 +1,71 @@
...
+        className={expanded ? iconContainerExpandedClass : iconContainerClass}
+      >
+        {hasCloseButton && (
+          <NotificationButton
+            onClick={onClose}
+            icon={SystemIcon.Close}
+            color={color}
+          ></NotificationButton>
+        )}
+      </span>
```
eileenmguo, line 68: Given the way that this is written, there can be an empty span. Why don't you create a custom closeButtonClass and just render the button using the ternary?

eileenmguo (approved):

PR #412, opened 2023-06-14
chore(monorepo): update eslint-config-next to match with new Next version

alber70g:


lkononenko:
Now `transfer` app needs to be updated as well

lkononenko (approved):

PR #413, opened 2023-06-14
[@kadena/client] ContCommand Class 

nil-amrutlal-dept:
## Reason:

Closes [issue link]

## Check off the following:

- [ ] I have added tutorials.


packages/libs/client/etc/client.api.md
```diff
@@ -21,9 +21,26 @@ export function buildCommandFromTemplate(parts: string[], holes: string[], args:
 // @internal (undocumented)
 export function buildUnsignedTransaction(parts: string[], holes: string[], args: Record<string, string>): IPactCommand & ICommandBuilder<{}>;
 
+// @alpha
+export class ContCommand extends PactCommand implements IContCommand, IContCommandBuilder<Record<string, unknown>> {
+    constructor(proof: string, step: number, pactId: string, rollback: boolean);
```
nil-amrutlal-dept, line 26: Javad
_I think its better to pass (proof, step, pactId, rollback) to the constructor, then you don't need those related setters.
and the API would be more cleaner because we want to set those properties in one step anyway_

nil-amrutlal-dept: Changed made

nil-amrutlal-dept (commented):
Albert
_General comment.
I see the use of @kadena/types which is used in other places, prior to this.
I'm becoming less a fan of this. I'd rather have the source describe the types and know where it comes from.

It's either from a function (function/class/etc) that requires a certain datastructure, OR it's an API that describes this. Therefore the goal is to make the different api endpoints a wrapper in the repo with it's types. Preferably generated from a spec (openapi/swagger/...).

For this PR I'd like to have the types described and exported from the place where they're required and referenced from that place when they are required somewhere else_

packages/libs/client/src/contPact.ts
```diff
@@ -0,0 +1,200 @@
...
+import { PactCommand } from './pact';
+
+/**
+ * @alpha
+ */
+export interface IContCommandBuilder<
+  TCaps extends Record<string, TArgs>,
+  TArgs extends Array<TCaps[keyof TCaps]> = TCaps[keyof TCaps],
+> {
+  createCommand(): IUnsignedCommand;
```
nil-amrutlal-dept, line 14: Albert

> Do conts have `caps`? I don't think so, so we can leave that out.

> You cannot add `code `to a `cont`, so even for that case it's irrelevant

nil-amrutlal-dept: Javad 

> based on the API (https://api.chainweb.com/openapi/pact.html#tag/model-payload) tit seems that the only distinction between "cont" and "exec" is the payload. so I think it also accept capabilities.

javadkh2:
I think its good enough for now

javadkh2 (approved):

MRVDH (approved):
Looking good!

packages/libs/client/src/pactCreator.ts
```diff
@@ -0,0 +1,52 @@
+import { createExp } from '@kadena/pactjs';
+
+import { parseType } from './utils/parseType';
+import { ContCommand } from './contPact';
+import { IPact, PactCommand } from './pact';
+
+import { log } from 'debug';
```
javadkh2, line 7: this is not correct the [log function](https://github.com/debug-js/debug/blob/master/src/node.js#L193) is the pure stderr.write without any checking, you need to use the setup function instead.
```suggestion
import debug, { Debugger } from 'debug';

const log: Debugger = debug('pactjs:proxy');
```

common/changes/@kadena/types/feat-contCommand-class-implementation_2023-06-14-13-44.json
alber70g, line 1: This one can be removed right?

packages/libs/client/src/tests/contPact.test.ts
```diff
@@ -0,0 +1,115 @@
...
+beforeEach(() => {
+  targetChainId = '1';
+  requestKey = 'requestKey';
+  apiHost = 'apiHost';
+  step = 1;
+  rollback = false;
+});
+describe('ContCommand', () => {
+  describe('createCommand', () => {
+    test('should return a valid command', () => {
```
alber70g, line 20: can we use `it`s and omit `should`. (ie `returns a valid command`)

packages/libs/client/src/tests/utils.test.ts
```diff
@@ -0,0 +1,40 @@
+import { parseType } from '../utils/parseType'; // Replace with the actual module path
+
+describe('parseType', () => {
+  test('should return the decimal value when input is a PactLiteral object with decimal property', () => {
```
alber70g, line 4: idem

packages/libs/pactjs-test-project/src/example-templates/generated-tx-templates/index.ts
```diff
@@ -11,17 +11,17 @@ export default {
...
     const parts = [
-      'code: |-\r\n  (coin.transfer "',
+      'code: |-\n  (coin.transfer "',
       '" "',
       '" ',
-      ')\r\ndata:\r\npublicMeta:\r\n  chainId: "',
-      '"\r\n  sender: ',
-      '\r\n  gasLimit: 2500\r\n  gasPrice: 1.0e-8\r\n  ttl: 600\r\nnetworkId: ',
-      '\r\nsigners:\r\n  - pubKey: ',
-      '\r\n    caps:\r\n      - name: "coin.TRANSFER"\r\n        args: [',
```
alber70g, line 21: Should we use snapshot testing here to prevent issues with line endings? Similar to https://github.com/kadena-community/kadena.js/blob/8b5603f453722e50751c95c96fc483f1a2efa880/packages/libs/pactjs-generator/src/template/generation/tests/generateTemplates.test.ts#L51

alber70g (changes requested):

packages/libs/pactjs-test-project/src/example-templates/generated-tx-templates/index.ts
alber70g: My bad, I though this was a test. Leave it for now. We might want to remove this one from the repo

alber70g (approved):

PR #422, opened 2023-06-15
fix build

rezaerami:
- add condition for page extensions since on dev regex won't work
- add cross env to pass inline env variable in commands

## Reason: fix build

Closes [issue link]

- add condition for page extensions since on dev regex won't work
- add cross env to pass inline env variable in commands

## Check off the following:

- [ ] I have added tutorials.

### DIFF
--- a/packages/apps/transfer/next.config.js
+++ b/packages/apps/transfer/next.config.js
@@ -3,14 +3,17 @@ const nextTranslate = require('next-translate-plugin');
 
 module.exports = nextTranslate({
   reactStrictMode: true,
-  pageExtensions: ['^((?!(test|spec)).)*.tsx$'],
-  webpack: (config, { dev }) => {
+  pageExtensions: process.env.NODE_ENV === "production" ? ['(?<!(spec|test).)tsx'] : ['tsx'],
+  webpack: (config, { isServer }) => {
     config.module.rules.push(
       {
         test: /\.(spec|test).*$/,
         loader: 'ignore-loader'
       }
     );
+    if (!isServer) {
+      config.resolve.fallback.fs = false;
+    }
     return config;
-  }
+  },
 });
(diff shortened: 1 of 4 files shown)


MRVDH (approved):

timoheddes (approved):

packages/apps/transfer/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "_phase:build": "npm run build",
     "_phase:test": "npm run test",
-    "build": "next build",
+    "build": "cross-env NODE_ENV=production next build",
```
ash-vd, line 8: Do we really need this? To my knowledge `next build` automatically sets this (https://github.com/vercel/next.js/discussions/13410#discussioncomment-18760)

rezaerami: yep, this will distinguish between dev or prod in next.config file later to pass regex or simple file extension, otherwise it'll always hit the TSX extension and try to build test files which will later throw error during build

packages/apps/transfer/next.config.js
```diff
@@ -3,14 +3,17 @@ const nextTranslate = require('next-translate-plugin');
...
+  pageExtensions: process.env.NODE_ENV === "production" ? ['(?<!(spec|test).)tsx'] : ['tsx'],
+  webpack: (config, { isServer }) => {
     config.module.rules.push(
       {
         test: /\.(spec|test).*$/,
         loader: 'ignore-loader'
       }
     );
+    if (!isServer) {
+      config.resolve.fallback.fs = false;
```
sanderlooijenga, line 15: Perhaps add a comment here indicating why this is done? 😁

sanderlooijenga (approved):

packages/apps/transfer/next.config.js
rezaerami: this is regarding the erorr of "module fs not found" cause by default it'll put fs codes on client (if we use fs in components) which will be executed on server, but not the client, so this is for later issues that might happen by using fs module in app directory

PR #424, opened 2023-06-15
Simplify CODEOWNERS

webpro:
Not so long ago we had a simple `CODEOWNERS` file, it looked something like this:

```
* @alber70g @Randynamic @ggobugi27 
```

That wasn't great when their availability is limited, they don't have the in-depth knowledge to review/approve something, etc. So a few weeks ago I went ahead and created basically what we currently in the master branch. It seems to work fine, but some of us think we can do with less targeted ownership. This PR depicts this idea, and comes with the message that we can take responsibility and request reviews and approvals from the people we think are best suited to do so.

With this new setup, lots of files do no longer have code ownership. There are still folders/packages that have one owner. This means that changes touching them require their approval. Overall, it should be easier to get PRs through.

The required number of approvals is always 2. It is up to you to chase more if that's appropriate for the changes made.

If you think you should own certain folder(s) or file(s), then you can tell us now; or later in another PR.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,33 +1,23 @@
 # About code owners: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-*                                               @alber70g @randynamic @ggobugi27
+/*.*                                            @alber70g
+/*.md                                           @webpro
+/common                                         @alber70g
+/packages/apps/graph                            @alber70g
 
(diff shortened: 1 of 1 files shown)


Takadenoshi (approved):

EnoF (approved):

ash-vd (approved):

alber70g (approved):

Randynamic (approved):

MRVDH (approved):
yes PLEASE :D

PR #426, opened 2023-06-15
Table component vanilla extract rewrite

bearni95:
## Reason:

Closes ["Recreate table component"](https://app.asana.com/0/1204649083736950/1204755978331326/f)

- recreated the table component, according to the latest figma design (included in the Asana ticket)
- added functionality to add a url to each `<tr>` row

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+    rowCount: 3,
+    columnCount: 5,
+  },
+  render: ({ rowCount, columnCount }) => {
+    return (
+      <Table>
+        <Table.Head>
+          <Table.Tr>
+            {Array.from(Array(columnCount)).map((id, tdIdx) => {
+              return <Table.Tr.Th key={`td${tdIdx}`}>test {tdIdx}</Table.Tr.Th>;
```
eileenmguo, line 49: I think keeping subcomponents at one level is best. So `Table.Th` should be sufficient

packages/libs/react-ui/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+        return child;
+      })}
+    </table>
+  );
+};
+
+Table.Body = TBody;
+Table.Head = THead;
+Table.Tr = Tr;
```
eileenmguo, line 38: What is the purpose of this?

packages/libs/react-ui/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children?: CompoundType<typeof TBody> | CompoundType<typeof THead>;
+}
+
+interface ITableComposition extends FC<ITable> {
+  Body: typeof TBody;
+  Head: typeof THead;
+  Tr: typeof Tr;
+}
+
+// eslint-disable-next-line react/prop-types
```
eileenmguo, line 19: What is this reason for this? We want to avoid disabling lint errors when possible

eileenmguo (commented):
Probably need to have a conversation with Isa about how link should work on this table

packages/libs/react-ui/src/components/Table/Table.tsx
bearni95: Was part of the original implementation for the table, simplifying it now

packages/libs/react-ui/src/components/Table/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+interface ITable extends FC<ITableProps> {
+  Body: FC<ITBodyProps>;
+  Head: FC<ITHeadProps>;
+  Tr: FC<ITrProps>;
+  Th: FC<IThProps>;
+  Td: FC<ITdProps>;
+}
+
+export const Table: ITable = TableContainer as ITable;
```
eileenmguo, line 20: Can you achieve this without a type cast?

eileenmguo (commented):
Are the colors, active, and focus states correct? 
image
@isaozler @bearni95

packages/libs/react-ui/src/components/Table/index.ts
bearni95: As we discussed, replaced it with a container variable, exporting the `<Table>` component as `Table.Root`

packages/libs/react-ui/src/styles/sprinkles.css.ts
```diff
@@ -13,6 +13,7 @@ export const breakpoints: Record<string, string> = {
 const systemProperties = defineProperties({
   properties: {
     fontFamily: vars.fonts,
+    borderSpacing: vars.sizes,
```
eileenmguo, line 16: Looks like border-spacing is a table specific css property so it likely won't be used much. I would remove this from sprinkles

eileenmguo (approved):

PR #428, opened 2023-06-16
fix: environment variables

rezaerami:
- add env file
- expose envs to client
- fix services to use envs
- add ace builds packages

## Reason: fix environment variables 

Closes [issue link]

- add env file
- expose envs to client
- fix services to use envs
- add ace builds packages

## Check off the following:

- [ ] I have added tutorials.


packages/apps/transfer/next.config.js
```diff
@@ -4,6 +4,15 @@ const nextTranslate = require('next-translate-plugin');
...
   pageExtensions: process.env.NODE_ENV === "production" ? ['(?<!(spec|test).)tsx'] : ['tsx'],
+  env: {
+    KADENA_API_TTIL: JSON.stringify(process.env.KADENA_API_TTIL),
+    KADENA_MAINNET_API: JSON.stringify(process.env.KADENA_MAINNET_API),
+    KADENA_MAINNET_NETWORKS: JSON.stringify(process.env.KADENA_MAINNET_NETWORKS),
+    KADENA_TESTNET_API: JSON.stringify(process.env.KADENA_TESTNET_API),
+    KADENA_TESTNET_NETWORKS: JSON.stringify(process.env.KADENA_TESTNET_NETWORKS),
+    GAS_PRICE: JSON.stringify(process.env.GAS_PRICE),
+    GAS_LIMIT: JSON.stringify(process.env.GAS_LIMIT),
+  },
```
alber70g, line 15: What's te reason to `JSON.stringify` everything? If I recall correctly, everything from `process.env` is already a `string`

packages/apps/transfer/src/constants/kadena.ts
```diff
@@ -0,0 +1,35 @@
...
+      MAINNET01: 'mainnet01',
+    },
+    API_HOST: ({
+      networkId,
+      chainId,
+    }: {
+      networkId: string;
+      chainId: string;
+    }) =>
+      `${process.env.KADENA_MAINNET_API}/chainweb/0.0/${networkId}/chain/${chainId}/pact` ??
```
alber70g, line 14: I usually use a util function that retrieves all the env vars, adds typings and does parsing where needed. This also allows for easy tracking of all env-vars.

E.g. https://github.com/kadena-community/kadena.js/blob/master/packages/apps/graph/src/utils/dotenv.ts

alber70g: Oh, nvm :) I see this is that file 😅

packages/apps/transfer/.env.example
```diff
@@ -0,0 +1,10 @@
+NODE_ENV=development
+
+# Kadena Networks
+KADENA_API_TTIL=60000
+KADENA_MAINNET_API=api.chainweb.com
+KADENA_TESTNET_API=api.testnet.chainweb.com
+
+# Gas Setting
+GAS_PRICE="0.00000001"
+GAS_LIMIT="750"
```
alber70g, line 10: Not sure what the use cases are that should be covered by default with these gas prices, but isn't this a bit low?
Maybe add a comment what these defaults should represent

packages/apps/transfer/.env.example
```diff
@@ -0,0 +1,10 @@
+NODE_ENV=development
+
+# Kadena Networks
+KADENA_API_TTIL=60000
```
alber70g, line 4: A TTL for a transaction of 60 seconds is fairly short (I don't know whether this is ms or seconds). Also good to mention what this default is used for
Depending on the use case different TTLs apply. For a transaction I'd say at least 3 but maybe more block heights, (3 being 90 sec).

packages/apps/transfer/next.config.js
rezaerami: it's not required, it's just to ensure that the value received is a string on the different environments

packages/apps/transfer/.env.example
Randynamic: IMHO `60000*60` would be a nice default.

packages/apps/transfer/src/constants/kadena.ts
rezaerami: added a new layer for env with an interface and a function to get envs 
the only issue is it can't be dynamically following the docs from nextjs they replace process.env.* with their value so it's not a normal javascript object so it can't be called dynamically like process.env[key] so I had to map it to an actual object and then get what I wanted

packages/apps/transfer/.env.example
rezaerami: removed gas price since it's adjustable by user, but we should keep gas limit, and number is coming from old transfer website
let me know if I should change it

packages/apps/transfer/.env.example
rezaerami: changed it to 10 minutes

alber70g (approved):

PR #429, opened 2023-06-16
chore(monorepo): prettier runs with global config

alber70g:


ash-vd (approved):

sanderlooijenga (approved):

packages/libs/react-ui/package.json
lkononenko, line 1: react-ui doesn't have `lint-staged` command in the `scripts`

sstraatemans (approved):

packages/libs/react-ui/package.json
alber70g: Thanks! Adding it now

PR #432, opened 2023-06-16
Breadcrumb component vanilla extract rewrite

bearni95:
## Reason:

Closes ["Recreate Breadcrumb Component"](https://app.asana.com/0/1204649083736950/1204755978331334/f)

- recreated the breadcrumbs component using `vanilla-extract`
- 
## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children?: FunctionComponentElement<IBreadcrumbItemProps>[];
+  icon?: typeof ProductIcon[keyof typeof ProductIcon];
+}
+
+export const BreadcrumbsContainer: FC<IBreadcrumbsProps> = ({
+  children,
+  icon,
+}) => {
+  return (
+    <div className={containerClass}>
```
Randynamic, line 18: So breadcrumbs is used for navigation correct? Why aren't we using the nav item here?

bearni95: It looks like [the way to go](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav)!

Instead of divs I've used the structure from the original breadcrumbs implementation (`<ul>`/`<li>`), and wrapped both in a `<nav>` tag

Randynamic: Yes, and because we are using the Compound Components pattern it's ok to have a tight coupling of the `ul` and `li` structures here.

Randynamic (approved):

packages/libs/react-ui/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import { ProductIcon } from '../Icons';
+
+import { Breadcrumbs, IBreadcrumbsProps } from '.';
+
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+const ItemArray: string[] = [
+  'He-man',
```
Randynamic, line 9: 🤣

packages/libs/react-ui/src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children,
+  icon,
+  href,
+}) => {
+  const Icon = icon;
+
+  return (
+    <li className={itemClass}>
+      {Icon && (
+        <div className={iconContainer}>
```
Randynamic, line 28: shouldn't we use an inline element here instead of a block-level element.
Do we have rules about this stated somewhere?

eileenmguo (commented):
Almost there! The component just needs to be styled more closely to the design 
image

bearni95:
> Almost there! The component just needs to be styled more closely to the design image

breadcrumbs

What browser are you using? It displays fine on both Firefox and Chrome on my end!

bearni95:
> > Almost there! The component just needs to be styled more closely to the design image
> 
> breadcrumbs
> 
> What browser are you using? It displays fine on both Firefox and Chrome on my end!

Seems similar to the `gridGap` issue on [the grid fixes PR](https://github.com/kadena-community/kadena.js/pull/456#issuecomment-1603286745), as in, the font seems to be loading rather slow, and interfering with the styling

eileenmguo:
> > > Almost there! The component just needs to be styled more closely to the design image
> > 
> > 
> > breadcrumbs
> > What browser are you using? It displays fine on both Firefox and Chrome on my end!
> 
> Seems similar to the `gridGap` issue on [the grid fixes PR](https://github.com/kadena-community/kadena.js/pull/456#issuecomment-1603286745), as in, the font seems to be loading rather slow, and interfering with the styling

So it did have the same issue where the sizing wasn't being applied, however also in this case it doesn't make sense for the spacing the change between the icon and the letters depending on breakpoints so I just changed it to a fixed size which works for this PR and the dynamic spacing variables will be fixed in the grid PR!

eileenmguo (approved):

PR #433, opened 2023-06-16
Box Component

bearni95:
## Reason:

Closes ["Box Component"](https://app.asana.com/0/1204649083736950/1204549680288790/f)

- Primitive layout component that gives access to design tokens for margin

## Check off the following:

- [x] I have added tutorials.


packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -21,7 +21,7 @@ export const container = style([
 export const fullWidthClass = style([sprinkles({ width: '100%' })]);
 
 export const stackClass = style([
-  sprinkles({ marginY: 0 }),
+  sprinkles({ marginY: '$0' }),
```
eileenmguo, line 24: I don't believe 0 needs the `$` prefix since it is a not a custom value

eileenmguo (changes requested):
I think there may have been some miscommunication about this component. Where are you getting the styles for this component from?

bearni95:
As discussed, tried integrating a similar pattern to the one vanilla-extract uses for their own `<Box>` component. 

PR reverted to draft due to the experimental status of the current implementation

eileenmguo (approved):
This is super cool! I think it works great :)

packages/libs/react-ui/src/components/Box/Box.tsx
```diff
@@ -0,0 +1,44 @@
...
+  marginX = undefined,
+  marginY = undefined,
+  marginTop = undefined,
+  marginBottom = undefined,
+  marginLeft = undefined,
+  marginRight = undefined,
+  top = undefined,
+  bottom = undefined,
+  left = undefined,
+  right = undefined,
```
eileenmguo, line 22: What is the reason for adding these?

packages/libs/react-ui/src/components/Box/Box.stories.tsx
```diff
@@ -0,0 +1,106 @@
...
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+const selectOptions: (keyof typeof vars.sizes | undefined)[] = [
+  undefined,
+  ...(Object.keys(vars.sizes) as (keyof typeof vars.sizes)[]),
+];
+
+const meta: Meta<IBoxProps> = {
+  title: 'Layout/Vanilla Extract Box',
```
eileenmguo, line 15: I would also change this to Layout/Box

packages/libs/react-ui/src/components/Box/Box.tsx
```diff
@@ -0,0 +1,44 @@
+import { Sprinkles, sprinkles } from '../../styles';
+
+import React, { createElement, ElementType } from 'react';
+
+export interface IBoxProps extends Sprinkles {
```
eileenmguo, line 5: Can we try making the type more strict? So there is a typescript error if anyone tries to pass in anything other than the margin props

packages/libs/react-ui/src/components/Box/Box.tsx
bearni95: Leftovers from testing, removed now

eileenmguo (approved):
Super excited about this one!

PR #434, opened 2023-06-19
Move PR guidelines to comment, allow freeform description

webpro:
The current pull request template isn't really working, imho. Either we just tick all the boxes, or we don't even bother at all. How can we improve this? Make sure PRs all complete including descriptions that make sense?

This PR introduces an even simpler message, since I think "freeform" results in a lower threshold to fill out the description with relevant things, but happy to hear your thoughts.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,10 +1,10 @@
 <!--
 Thanks for contributing to this project!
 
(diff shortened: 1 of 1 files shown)


timoheddes:
Perhaps instead of deleting the boxes entirely, we can keep the list of them as reminders. E.g. "Please check that you've done the following (as relevant/appropriate to your changes): ...."

alber70g (approved):

alber70g (approved):

PR #437, opened 2023-06-19
Remove git hooks & fix formatting

webpro:
First off, apologies for this large PR. Had to fix all the formatting issues everywhere in order to get this PR itself green. Probably best to review commits one by one.

tl/dr;

- Remove Git hooks, including everything related to `husky` + `lint-staged`
- Add `format` script to packages
- Add `--check` argument so CI fails on formatting issues
- Format _all the things_ to get green lights from CI
- Remove `sort-package-json`
  - We weren't really using it anyway, and is problematic with `format --check` in CI.
  - Sort(ed) them all using `npx sort-package-json "packages/*/*/package.json"`
- Add shell script to check values of script in every `package.json` (to easily find missing or oddities in e.g. `lint` or `format` scripts)

**NOTE** You can now remove the Git hook if you want. Then you are responsible to lint and format your code according to the configuration files (global `.prettierrc` and local `.eslintrc.js`). Use your IDE for this or the CLI: you can use commands like `rush lint --only <name>` from root or `rushx format` in project dir. CI will complain if you forget, but such checks are a lot more efficient from your own machine.


webpro:
Current output of that shell script:

Screenshot 2023-06-19 at 5 25 22 PM

Screenshot 2023-06-19 at 5 25 46 PM

packages/libs/pactjs-generator/package.json
```diff
@@ -21,23 +21,11 @@
     "build:grammar:watch": "chokidar src/lexer.js src/grammar.ne src/tests/test.contract.pact -c \"rushx build:grammar\"",
     "test:grammar:watch": "heft test -w -t grammar --disable-code-coverage",
     "util:lexer-grammar:watch": "concurrently --kill-others npm:build:*:watch # npm:test:grammar:watch",
+    "format": "prettier config src --write -u --config ../../../.prettierrc --ignore-path ../../../.prettierignore",
```
EnoF, line 24: Can we provide the `.prettierrc` via a `@kadena-dev` package? Similar to eslint-config? Or are there other reasons to have it exposed as this?

EnoF (approved):

packages/libs/pactjs-generator/package.json
webpro: Good point. For now I just repeated what we already did. Not sure what's the best way here. Ideally, packages are fully self-contained so I guess we could copy that file around. Or put it in another package, but I don't like the magic here (e.g. symlinks or copying files before running scripts, etc). Could be something to discuss and apply in another PR. Current solution is ugly but 100% concise + effective.

ash-vd (approved):

.prettierignore
```diff
@@ -1,2 +1,5 @@
 CODEOWNERS
+packages/**/__generated__/**
+packages/apps/kadena-docs/src/data/menu.mjs
```
Randynamic, line 3: why are these added just out of curiosity?

webpro: Because with this PR we're formatting more code. The previous setup wasn't complete. And still isn't tbqh, but we'll get to that soon enough.

Randynamic (approved):

Takadenoshi (approved):

alber70g (approved):

webpro:
Had to add a merge commit, it was unable to rebase.

PR #439, opened 2023-06-20
Feat/reactui modal

sstraatemans:


packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -18,7 +18,9 @@ export const container = style([
   }),
 ]);
 
-export const fullWidthClass = style([sprinkles({ width: '100%' })]);
+export const fullWidthClass = style({
```
sstraatemans, line 21: if I dont, the width is overwritten by the "max-content" set earlier

alber70g (approved):

Randynamic:
When opening the modal does the body get an `overflow: hidden;` to stop scrolling when the modal is visible?

Randynamic (approved):

packages/libs/react-ui/src/components/Card/Card.css.ts
eileenmguo: I think i'm missing some context. Why was this switched from a sprinkle to custom styling?

packages/libs/react-ui/src/styles/sprinkles.css.ts
```diff
@@ -73,10 +74,25 @@ const responsiveProperties = defineProperties({
...
+      $screen75: '75vw',
+      $screen100: '100vw',
+    },
+    height: {
+      ...vars.sizes,
+      '100%': '100%',
+      $screen25: '25vh',
+      $screen50: '50vh',
+      $screen75: '75vh',
+      $screen100: '100vh',
```
eileenmguo, line 94: Are these values that we would use often? For me, these seem like style values that would only be used once or twice in an app

packages/libs/react-ui/src/styles/sprinkles.css.ts
```diff
@@ -73,10 +74,25 @@ const responsiveProperties = defineProperties({
     paddingRight: vars.sizes,
     marginTop: vars.sizes,
     marginBottom: vars.sizes,
-    marginLeft: vars.sizes,
-    marginRight: vars.sizes,
-    width: { ...vars.sizes, '100%': '100%', 'max-content': 'max-content' },
-    height: { ...vars.sizes, '100%': '100%' },
+    marginLeft: { ...vars.sizes, $auto: 'auto' },
```
eileenmguo, line 77: So this is a bit confusing, but we are not prefixing all values with `$`. It is only custom values. So in this case all of our custom sizes would be prefixed with `$` but since auto is one of the existing options for marginLeft, we would just name it `auto`. Also this should be added to the shorthand version `marginX`

eileenmguo:
image

Is this styled correctly? The close button looks strange

packages/libs/react-ui/src/components/Card/Card.css.ts
sstraatemans: for some reason when using the sprinkels its specificity was to low.
removing it, it changes in the current classname. overwriting the old value

packages/libs/react-ui/src/components/Modal/Modal.css.ts
```diff
@@ -0,0 +1,72 @@
...
+export const closeButton = style([
+  sprinkles({
+    display: 'flex',
+    alignItems: 'center',
+    backgroundColor: 'transparent',
+    fontWeight: '$normal',
+    borderStyles: 'none',
+  }),
+  {
+    border: '0',
```
eileenmguo, line 70: I think this can be moved to sprinkles as border: "none"

eileenmguo (approved):

PR #442, opened 2023-06-20
Allow warnings in test and lint commands

webpro:
Warnings are warnings, they should not make CI runs fail. A few notes:

- External tooling might use `console.warn` that's out of our control.
- ESLint warnings from rule violations that we consider errors should be configured as such.
- We shouldn't need to work around this by piping stuff (`stdout`/`stderr`), etc.

I wonder for what reason exactly we need this, and how we might mitigate that issue in a different way?

### DIFF
--- a/common/config/rush/command-line.json
+++ b/common/config/rush/command-line.json
@@ -20,7 +20,7 @@
       "enableParallelism": true,
       "ignoreDependencyOrder": false,
       "ignoreMissingScript": false,
-      "allowWarningsInSuccessfulBuild": false,
+      "allowWarningsInSuccessfulBuild": true,
       // default: false. Changed to true as it should not need to run when nothing has changed
       "incremental": true,
       "watchForChanges": false,
@@ -37,7 +37,7 @@
     //   "enableParallelism": true,
     //   "ignoreDependencyOrder": false,
     //   "ignoreMissingScript": false,
-    //   "allowWarningsInSuccessfulBuild": false,
+    //   "allowWarningsInSuccessfulBuild": true,
     //   // default: false. Changed to true as it should not need to run when nothing has changed
     //   "incremental": true,
     //   "watchForChanges": true,
@@ -100,7 +100,7 @@
       "ignoreDependencyOrder": false,
       // default: false. this command is just for convenience
       "ignoreMissingScript": true,
-      "allowWarningsInSuccessfulBuild": false,
+      "allowWarningsInSuccessfulBuild": true,
       "incremental": false,
       "watchForChanges": false,
       "disableBuildCache": false
@@ -197,7 +197,7 @@
     //    *
     //    * Note: The default value is false. In Rush 5.7.x and earlier, the default value was true.
     //    */
-    //   "allowWarningsInSuccessfulBuild": false,
+    //   "allowWarningsInSuccessfulBuild": true,
     //
     //   /**
     //    * If true then this command will be incremental like the built-in "build" command


Takadenoshi:
Agree about all your points. Especially stderr during testing has surfaced a few separate times recently.

Regarding eslint I am not sure when we would intentionally choose to use warnings vs errors, but if we do, then we can configure it to exit non-zero for any warning. (which would make them... errors?)

Another question: do we want non-build-failing WARNs in our eslint config?

Takadenoshi:
Our current WARN rules:

import/no-duplicates

@typescript-eslint/strict-boolean-expressions

@kadena-dev/typedef-var

```        '@typescript-eslint/typedef': [
          'warn',
          {
            arrayDestructuring: false,
            arrowParameter: false,
            memberVariableDeclaration: true,
            objectDestructuring: false,
            parameter: false,
            propertyDeclaration: true,
            variableDeclaration: false,
            variableDeclarationIgnoreFunction: true,
          },
        ],
```

alber70g:
Making warnings fail the build is two fold:

 - we want to allow things to be non-blocking during (local) development, but they should go wrong in the build
 - having accumulated warnings will lead to no one looking seriously at warnings anymore. That one important warning is drowned in clutter

Some more reasoning (from https://rushjs.io/pages/maintainer/add_to_repo/): If the command writes anything to the stderr stream, Rush will interpret this to mean that at least one error or warning was reported. This will break the build. (This is by design -- if you allow people to merge PRs that "cry wolf", pretty soon you will find that so many warnings have accumulated that nobody even reads them any more.) Some tooling libraries (e.g. Jest) write to stderr as part of their normal operation; you will need to redirect their output.


I'd say all warnings should be able to be suppressed properly, individually and consciously, however, some sources of warnings do not have the ability to suppress a warning in that manner. 

Let's 

1. find a way that the things will be blocking when they should be (in the ci) and are allowed in local development, and
2. have a process that allows to reflect on warnings that should be addressed

webpro:
> Regarding eslint I am not sure when we would intentionally choose to use warnings vs errors, but if we do, then we can configure it to exit non-zero for any warning. (which would make them... errors?)

Warnings should not result in non-zero exits. And it shouldn't be configuration magic people don't immediately grasp: as we can see people are confused as to why CI runs unnecessarily fail now, just from warnings.

> Another question: do we want non-build-failing WARNs in our eslint config?

Personally I think it's normal for e.g. the naming convention stuff to be warnings, definitely not errors that must be "fixed" or suppressed. Developers see the warning during development and decide how to act on it (and there's still PR reviews, and that's that). So there are some rules we don't need to be so vigorous about, imho.

By the way, ESLint has an option to suppress warnings. We could use that in CI.

webpro:
> Making warnings fail the build is two fold:
> 
> * we want to allow things to be non-blocking during (local) development, but they should go wrong in the build

Not sure I agree with this. I think things should behave the same as much as possible in dev vs production (and also in CI). Fail early, fail fast. Short feedback loop. I also like to refer to this old-school but still relevant https://12factor.net/dev-prod-parity. Or did you mean something else?

> * having accumulated warnings will lead to no one looking seriously at warnings anymore. That one important warning is drowned in clutter

If it's important enough to be an error, it should be an error.

> Some more reasoning (from https://rushjs.io/pages/maintainer/add_to_repo/): If the command writes anything to the stderr stream, Rush will interpret this to mean that at least one error or warning was reported. This will break the build. (This is by design -- if you allow people to merge PRs that "cry wolf", pretty soon you will find that so many warnings have accumulated that nobody even reads them any more.) Some tooling libraries (e.g. Jest) write to stderr as part of their normal operation; you will need to redirect their output.

Then I'd say we should be looking into fixing or working around that Jest issue, not let everything else suffer from this.

> I'd say all warnings should be able to be suppressed properly, individually and consciously, however, some sources of warnings do not have the ability to suppress a warning in that manner.
> 
> Let's
> 
> 1. find a way that the things will be blocking when they should be (in the ci) and are allowed in local development, and

Make them behave the same as much as possible.

> 2. have a process that allows to reflect on warnings that should be addressed

Development + pull requests. Like you said, we don't want to drown in warnings, so they should imho always be addressed right away. But do we really have that issue? Let's see how "bad" it actually is first?

webpro:
What baffles me is that Rush marks `stderr` output from scripts that **exit with code zero** (`0`) as _warnings_ (and only then). And with the default `"allowWarningsInSuccessfulBuild": false` setting we are using it then proceeds to **exit with a non-zero code itself**.

This is... unexpected. First you need to understand what errors and warnings are to Rush, and then you need to understand that only in CI this results in failing jobs (and not during local development).

I think what Rush is missing is that scripts don't know upfront that they will eventually exit with zero or not. So they log to `stderr` as they see fit. Error logs, status information, diagnostics are written to `stderr` since it would interfere with output. We can't and don't want to control that. The only thing that matters is the final exit code, simple as that. I will die on this hill :)

Also, just checked and other tools like Nx and Turborepo don't even have/need this option.

Last but not least, this fragment from the Rush code docs strikes me as pedantic. Any project has its own responsibility to take care of this (not to be decided by a tool that sits in front of everything else):

> In an active monorepo, we've found that if you allow any warnings in your main branch, it inadvertently
> teaches developers to ignore warnings, which quickly leads to a situation where so many "expected" warnings
> have accumulated that warnings no longer serve any useful purpose.

/rant

webpro:
Glad to see this is sorted. For anyone reading this and using Rush: do not rely on `allowWarningsInSuccessfulBuild: true`, it's better to configure warning, errors and http exit codes properly.

PR #444, opened 2023-06-20
[@kadena/transfer] X-chain transfer finisher page

KristinaSpasevska:


packages/apps/transfer/src/services/utils/utils.ts
```diff
@@ -1,9 +1,17 @@
...
+): string => {
+  return `https://${server}/chainweb/0.0/${networkId}/chain/${chainId}/pact`;
 };
 
 export const convertIntToChainId = (value: number): ChainId => {
   return value.toString() as ChainId;
 };
+
+export const validateRequestKey = (key: string): boolean => {
+  return key.length === 43;
```
MRVDH, line 16: Do we know this 100% for sure?

packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
```diff
@@ -0,0 +1,82 @@
...
+        chainId: targetChain,
+        sender: gasPayer,
+        gasLimit,
+        gasPrice,
+      },
+      network,
+    );
+
+    // remove double quotes from stringified proof
+    contCommand.proof = contCommand.proof.slice(1, -1);
```
MRVDH, line 62: I don't know the proof format but wouldn't it better to do `.replaceAll('"', '')` just in case the api response ever changes to not include quotes?

packages/apps/transfer/src/pages/transfer/module-explorer/index.tsx
```diff
@@ -33,7 +35,22 @@ const GetCode: FC = () => {
   const [moduleChain, setModuleChain] = useState<number>(1);
   const [results, setResults] = useState<ModuleResult>({});
 
-  const networkdId = 'testnet04';
+  const { network } = useAppContext();
+
+  const chainNetwork: {
```
MRVDH, line 40: I see this code more often, we should probably make this into a constant

packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
nil-amrutlal-dept: I think this should be changed on the client itself. I will make the change. I also ran into the same issue and I don't think its sensible to have to slice or replace this manually.

packages/apps/transfer/src/pages/transfer/module-explorer/index.tsx
nil-amrutlal-dept: That would be great

packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
```diff
@@ -0,0 +1,82 @@
...
+  error?: string;
+}
+
+export interface TransferResult {
+  requestKey?: string;
+  status?: string;
+}
+
+const gasLimit: number = 850;
+const gasPrice: number = 0.00000001;
```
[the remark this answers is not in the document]
KristinaSpasevska: The gas limit already is an env var, but the price it think it may change regarding the speed. We may need to give the user an option to increase it if they want better speed (like the option in Chainweaver).

packages/apps/transfer/src/services/utils/utils.ts
KristinaSpasevska: I will check this with someone from Kadena

KristinaSpasevska: The same validation is used in old transfer, so I guess that's the correct format/length

packages/apps/transfer/src/pages/transfer/module-explorer/index.tsx
MRVDH: I created a separate task for this, no need to do it in this PR.

packages/apps/kadena-docs/src/data/menu.mjs
```diff
@@ -945,15 +945,5 @@ export const menuData = [
     "icon": "PactLanguage",
     "isMenuOpen": false,
     "isActive": false
-  },
```
MRVDH, line 948: docs? incorrect merge?

MRVDH (approved):
Approved with one comment 😄

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-finisher/index.tsx
alber70g, line 1: Does the transfer app use the generic eslint config? I think not as I see a bunch of things that are different compared to what it should be

alber70g (approved):

packages/apps/transfer/src/services/cross-chain-transfer-finish/get-transfer-data.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  try {
+    const chainInfoPromises = Array.from(new Array(20)).map((item, chainId) => {
+      const host = generateApiHost(
+        server,
+        networkId,
+        convertIntToChainId(chainId),
+      );
+      return pactCommand.poll(host);
+    });
```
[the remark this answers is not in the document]
KristinaSpasevska: I couldn't find any better solution to get receiver chainID & transfer data by only having the request key as available user data. If there is a better way, let me know.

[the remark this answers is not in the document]
alber70g: That's correct, with only the `requestKey` it isn't possible. You could add a chain as an optional input field for chain-id with a button next to it saying "Check all chains".
Especially when we're moving to 50-100 or even more chains, this becomes a lengthy operation

packages/apps/transfer/src/services/cross-chain-transfer-finish/get-transfer-data.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (validatedRequestKey === undefined) {
+    return { error: t('Invalid length of request key') };
+  }
+
+  const pactCommand = new PactCommand();
+
+  pactCommand.requestKey = requestKey;
+
+  try {
+    const chainInfoPromises = Array.from(new Array(20)).map((item, chainId) => {
```
alber70g, line 54: I'd rather have a constant with an array with all the separate chains spelled out. There's no reason chains will be always just a number. Maybe we can fix this in a separate PR across this project

packages/apps/transfer/src/components/Common/Layout/MainLayout.spec.tsx
```diff
@@ -7,6 +7,7 @@ jest.mock('next/router', () => ({
 import MainLayout from './MainLayout';
 
 import { render } from '@testing-library/react';
+// @ts-ignore
 import React from 'react';
 
 describe('MainLayout', () => {
```
alber70g, line 13: can we change everything to `describe`-`it`?

packages/apps/transfer/src/components/Common/Layout/MainLayout.spec.tsx
```diff
@@ -7,6 +7,7 @@ jest.mock('next/router', () => ({
 import MainLayout from './MainLayout';
 
 import { render } from '@testing-library/react';
+// @ts-ignore
 import React from 'react';
```
alber70g, line 11: Is this necessary?

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-finisher/index.tsx
```diff
@@ -1,128 +1,308 @@
...
-    },
-    Testnet: {
-      server: kadenaConstants.TESTNET.API,
-      network: kadenaConstants.TESTNET.NETWORKS.TESTNET04,
-    },
-  };
-
   const [requestKey, setRequestKey] = useState<string>('');
   const [kadenaXChainGas, setKadenaXChainGas] =
     useState<string>('kadena-xchain-gas');
```
alber70g, line 54: Setting the gas station as the sender should be a conscious choice, not the default behaviour. This is to have a minor barrier in using the gas station. Also, when the gas station is empty (which is possible) there should be a way to have a custom sender

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-finisher/index.tsx
```diff
@@ -1,128 +1,308 @@
...
+      setTxError(contCommand.error);
+    }
+
+    if (contCommand instanceof ContCommand) {
+      try {
+        const pollResult = await contCommand.pollUntil(host, {
+          onPoll: async (transaction, pollRequest): Promise<void> => {
+            console.log(
+              `Polling ${requestKey}.\nStatus: ${transaction.status}`,
+            );
```
alber70g, line 134: I see various `console.log`s. I think it's nice to have a debuggable flow in the browser, whenever there's issues popping up. This could be done using the [`debug`](https://www.npmjs.com/package/debug) library. We could add relevant log statements across the xchain flow.  What do you think?

packages/apps/transfer/src/components/Common/Layout/index.spec.tsx
```diff
@@ -1,6 +1,7 @@
 import Layout from './index';
 
 import { render } from '@testing-library/react';
+// @ts-ignore
```
alber70g, line 4: Idem

packages/apps/transfer/src/services/cross-chain-transfer-finish/get-transfer-data.ts
```diff
@@ -0,0 +1,101 @@
...
+      return pactCommand.poll(host);
+    });
+    const chainInfos = await Promise.all(chainInfoPromises);
+
+    let found: { chainId: number; tx: ICommandResult } | undefined;
+    chainInfos.map(async (pactInfo, chainId) => {
+      if (pactInfo[validatedRequestKey] !== undefined) {
+        found = { chainId: chainId, tx: pactInfo[validatedRequestKey] };
+      }
+    });
```
alber70g, line 69: We can use `forEach` here or reduce, and return one item or undefined

PR #449, opened 2023-06-21
[@kadena/transfer] Account Transactions

MRVDH:


packages/apps/transfer/src/services/accounts/get-transactions.ts
```diff
@@ -0,0 +1,36 @@
...
+export async function getTransactions(options: {
+  network: string;
+  chain: string;
+  account: string;
+}): Promise<Transaction[]> {
+  const { network, chain, account } = options;
+
+  try {
+    const result: Transaction[] = await fetch(
+      `https://estats${
```
KristinaSpasevska, line 25: It would be nice to have this whole string in a separate const

MRVDH: Agreed. I checked constants/kadena.ts but it's heavily focussed on the chainweb API. Would be better maybe to rewrite this in another task to also be able to make it focussed on multiple APIs, like estats. I will create a separate task for this on the board.

nil-amrutlal-dept:
Just from the UI point, these two inputs are not aligned vertically. Not sure if its supposed to be this way, if so, please ignore
image

KristinaSpasevska (approved):

PR #452, opened 2023-06-21
[@kadena/client] Fix stringified issue with proof

nil-amrutlal-dept:


KristinaSpasevska (approved):

MRVDH (approved):

MRVDH:
Note: possibly a breaking change for dApps implementing this. If they used a slice method on it, then this will break for them.

alber70g (approved):

PR #453, opened 2023-06-21
implement persistor

rezaerami:
- add js cookie
- add helper methods to persist with js cookie via prefix
- use persistor to get or set user's selected network

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -302,6 +302,7 @@ importers:
       ignore-loader: ~0.1.2
       jest: ~29.5.0
       jest-environment-jsdom: ~29.5.0
+      js-cookie: ~3.0.5
       lint-staged: ~13.2.0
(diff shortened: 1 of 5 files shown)


MRVDH (approved):

packages/apps/kadena-docs/src/data/menu.mjs
alber70g, line 1: What's this change? I see it in all the PRs but it's never going through it seems... @sstraatemans?

alber70g (approved):

packages/apps/kadena-docs/src/data/menu.mjs
sstraatemans: weird that it is in all PRs.
it is because the docs build has run.
but that already should be in master.

sstraatemans: this is fine.
I'm working on build scripts and after that, this file will be ignored and created in the build

PR #454, opened 2023-06-21
Feat/chainweb stream deduplicate transactions

Takadenoshi:


packages/libs/chainweb-stream-client/src/sliding-cache.ts
```diff
@@ -0,0 +1,176 @@
...
+    a.meta?.id === b.meta?.id &&
+    a.meta?.confirmations === b.meta?.confirmations,
+};
+
+export default class SlidingCache<Shape extends ITransaction>
+  implements SlidingCacheConfig<Shape>
+{
+  public minCacheValue: number = Infinity;
+  public maxCacheValue: number = -Infinity;
+  public cache: Shape[] = [];
```
alber70g, line 50: Depending on the `cacheSpan` this can get a very expensive. It could be more efficient to use the identifier as a key in a Map (opposed to a obj, as you're more `getting` from the Map/obj to compare, than `setting` to the map/obj)

alber70g (commented):
Comments/ideas

packages/libs/chainweb-stream-client/src/sliding-cache.ts
Takadenoshi: I do agree with this but I think there is a tradeoff to be made for the first release

For now I would opt for a correct implementation without over-complicating the data structures

In theory we don't need to store the entire txns either, we could keep 3 things per element: identity value (txn id), equality value (txn id + confirmation depth) and the caching value (height)

So maybe something like 2 maps, one with `{ [height]: [txn_id1, txn_id2, txn_id3] }` that would be necessary to evict quickly, and another with `{ [txn_id1]: [txn1_confirmation_depth] }` to track the most recent version of each transaction 

But that may also have some overhead as we will be creating new objects, whereas with this pattern we should just be referencing the existing txn in memory. 

I will test how this works with the chattiest module I can think (`coin`) and see if it releasable in this state or not.

PR #456, opened 2023-06-21
Fix grid stylings

bearni95:
The variant class for the `Grid` component wasn't inheriting the base class, causing a visual bug. 
- Concatenated them using `classNames`


packages/libs/react-ui/src/components/Grid/Grid.css.ts
```diff
@@ -16,15 +16,12 @@ export const gridItemClass = style([
...
-  // eslint-disable-next-line @typescript-eslint/naming-convention
-  '3xl': [sprinkles({ gridGap: '$3xl' })],
+  $2xs: [sprinkles({ gridGap: '$2xs' })],
+  $xs: [sprinkles({ gridGap: '$xs' })],
+  $sm: [sprinkles({ gridGap: '$sm' })],
+  $md: [sprinkles({ gridGap: '$md' })],
+  $lg: [sprinkles({ gridGap: '$lg' })],
+  $xl: [sprinkles({ gridGap: '$xl' })],
+  $2xl: [sprinkles({ gridGap: '$2xl' })],
+  $3xl: [sprinkles({ gridGap: '$3xl' })],
```
eileenmguo, line 26: These were intentionally left without the "$" prefix because they are not style values, but prop options. The idea was to keep the "$" only in the css as a way to use custom style values

packages/libs/react-ui/src/components/Grid/Grid.stories.tsx
```diff
@@ -28,14 +28,14 @@ type Story = StoryObj<typeof Grid.Container>;
...
   name: 'Grid',
   args: {
-    spacing: 'xl',
+    spacing: '$xl',
   },
   render: ({ spacing }) => (
     <>
       <Grid.Container spacing={spacing}>
-        {Array.from(new Array(12)).map((i) => (
+        {Array.from(new Array(12)).map((empty, i) => (
```
eileenmguo, line 36: can you use _ in the place of empty? Might look a bit cleaner

bearni95: I agree! sadly, it raises a warning with the linter... should we use `_` and add an ignore line?

packages/libs/react-ui/src/components/Grid/Grid.css.ts
bearni95: Trying to get rid of the ts-ignore-next line, but makes sense to revert for consistency

eileenmguo:
Spacing isn't being applied 
image

ash-vd (approved):

eileenmguo:
It was because the global.css file wasn't getting imported into storybook so the sm, md, lg, etc spacing variables weren't getting set! I just committed the change

eileenmguo (approved):

PR #468, opened 2023-06-22
[@kadena/react-ui]: Add input component

eileenmguo:


packages/libs/react-ui/src/components/Input/Input.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  },
+  render: ({ leftIcon, rightIcon, outlined, leadingText, onChange }) => (
+    <Input
+      id="inlineInputStory"
+      leftIcon={SystemIcon[leftIcon]}
+      rightIcon={SystemIcon[rightIcon]}
+      onChange={onChange}
+      placeholder="This is a placeholder"
+      leadingText={leadingText}
+      outlined={outlined}
```
timoheddes, line 78: I think you may be missing the disabled property?

eileenmguo: Yes, I missed that. Added 👍

sstraatemans (approved):

PR #471, opened 2023-06-23
Card design update

bearni95:
Adjusted the card component to reflect the latest design updates:
- Border around the card
- Lighter background color
- Shorter inter-card line when `stacked`

### DIFF
--- a/packages/libs/react-ui/src/components/Card/Card.css.ts
+++ b/packages/libs/react-ui/src/components/Card/Card.css.ts
@@ -4,7 +4,7 @@ import { style } from '@vanilla-extract/css';
 
 export const container = style([
   sprinkles({
-    backgroundColor: '$neutral2',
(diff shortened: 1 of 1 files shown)


packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -25,15 +29,24 @@ export const fullWidthClass = style({
 export const stackClass = style([
   sprinkles({ marginY: 0 }),
   {
+    margin: 0,
```
eileenmguo, line 32: can this be moved to sprinkles?

bearni95: it was actually redundant, removed it for simplicity on the last commit

eileenmguo (approved):

Issue #139, opened 2022-12-21
[@kadena/client] add support for pacts (defpact) (for crosschain transfers)


alber70g:
add support for pacts (for crosschain transfers)

alber70g:
An idea for implementing this. It is a product of discussions we had previously.

```ts
Pact.modules.coin.transfer-crosschain.exec("from" "to" 10)
Pact.modules.coin.transfer-crosschain.cont(pactId, step, rollback, proof)
```

Another part is the rewrite of the API to support three main phases:

1. expression creation,
2. transaction builder
3. finalisation

1. Build expression(s)
   a. const transferExp = Pact.modules.coin.transfer("from", "to", 10) 
4. Create builder
   a. const builder = createTransactionBuilder(transferExp)
   b. builder.addCap()
   c. builder.setMeta()
   d. builder.setData()
   e. builder.setNonceCreator()
   f. const unsignedTx = builder.createTransaction()
5. Finalise by signing/sending/polling
   a. Sign (optional) 
   ---i. const signedTx = sign(unsignedTx)
   b. Send
   ---i. const { send } = getClient(apiHostTemplate)
   ---ii. const response = await send(signedTx)

PR #475, opened 2023-06-26
Feat/3 phase tr

javadkh2:


packages/libs/client/src/v2/client/utils/request.ts
```diff
@@ -0,0 +1,139 @@
...
+  results.reduce((acc, data) => ({ ...acc, ...data }), {} as T);
+
+export type IPollRequestPromise<T> = Promise<Record<string, T>> & {
+  requests: Record<string, Promise<T>>;
+};
+
+export const mergeAllPollRequestPromises = <T extends object>(
+  results: Array<IPollRequestPromise<T>>,
+): IPollRequestPromise<T> => {
+  return Object.assign(Promise.all(results).then(mergeAll), {
```
EnoF, line 125: What does the `Object.assign` do on a promise?

packages/libs/client/src/v2/client/client.ts
```diff
@@ -0,0 +1,245 @@
...
+  ) {
+    const results = groupByHost(
+      typeof requestKeys === 'string' ? [requestKeys] : requestKeys,
+      options,
+    ).map(([hostUrl, requestKeys]) =>
+      pollStatus(hostUrl, requestKeys, options),
+    );
+
+    // merge all of the result in one object
+    const mergedResults = mergeAllPollRequestPromises(results);
```
EnoF, line 164: How does this work? I was kinda expecting a `await` before we actually get `mergedResults`? 

I was actually expecting something along the lines of `const mergedResults = await mergeAllPollRequestsPromises(results)`

packages/libs/client/src/v2/client/client.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+  function pollStatusFunction(
+    requestKeys?: string[] | string,
+    options?: IPollOptions & INetworkOptions,
+  ) {
+    const results = groupByHost(
+      typeof requestKeys === 'string' ? [requestKeys] : requestKeys,
+      options,
+    ).map(([hostUrl, requestKeys]) =>
+      pollStatus(hostUrl, requestKeys, options),
```
EnoF, line 160: I know we have a poll function, but do we not rather want to use the `listen`?

EnoF:
The folder example is miss spelled

packages/libs/client/src/v2/exmaple/simple-api.ts
```diff
@@ -0,0 +1,36 @@
...
+    coin.transfer('javad', 'albert', { decimal: '0.1' }),
+    coin.transfer('albert', 'javad', { decimal: '0.1' }),
+  ])
+  .singer('javad', (withCapability: any) => [
+    withCapability('coin.TRANSFER', 'javad', 'albert', { decimal: '0.1' }),
+  ])
+  .singer('javad', (withCapability: any) => [
+    withCapability('coin.TRANSFER', 'albert', 'javad', { decimal: '0.1' }),
+  ])
+  .sing((signedCommand: any) => {
```
EnoF, line 28: typo

packages/libs/client/src/v2/pact.ts
```diff
@@ -0,0 +1,238 @@
...
+  ): { command: Partial<ICommand>; stringify: () => string };
+}
+
+export const commandBuilder: ICommandBuilder = (first: any, ...rest: any) => {
+  const args: Array<
+    Partial<ICommand> | ((cmd: Partial<ICommand>) => Partial<ICommand>)
+  > = [first, ...rest] as any;
+  const command = args.reduce<Partial<ICommand>>((acc, item) => {
+    const part = typeof item === 'function' ? item(acc) : item;
+    if (part.payload !== undefined) {
```
EnoF, line 75: Should this logic be here in the command builder? I'd expect this logic to be present at the reducers that would modify the `payload`. Then this function can stay dumb. (same for all the other data checks below)

packages/libs/client/src/v2/pact.ts
```diff
@@ -0,0 +1,238 @@
...
+    ...second: [
+      ...Array<
+        | Partial<ICommand>
+        | ((payload: F & Partial<ICommand>) => Partial<ICommand>)
+      >,
+    ]
+  ): { command: Partial<ICommand>; stringify: () => string };
+}
+
+export const commandBuilder: ICommandBuilder = (first: any, ...rest: any) => {
```
EnoF, line 69: I'd rather have the command builder become lazy, which would allow for further composition after initial composition.

packages/libs/client/src/v2/pact.ts
```diff
@@ -0,0 +1,238 @@
...
+  ): () => Pick<ICommand, 'signers'>;
+  <T extends any>(
+    first:
+      | string
+      | { pubKey: string; scheme?: 'ED25519' | 'ETH'; address?: string },
+    capability: (withCapability: ExtractType<T>) => ICapabilityItem[],
+  ): T;
+}
+
+export const signer: ISinger = ((
```
EnoF, line 140: Singing the wrong song :p

packages/libs/client/src/v2/pact.ts
```diff
@@ -0,0 +1,238 @@
...
+    signers: [
+      {
+        ...prop('pubKey', pubKey),
+        ...prop('scheme', scheme),
+        ...prop('address', address),
+        ...prop('clist', clist),
+      },
+    ],
+  });
+}) as any;
```
EnoF, line 173: I suppose the `any` is just a WIP?

packages/libs/client/src/v2/client/utils/request.ts
javadkh2: its like this
```TypeScript
const promise = Promise.all(results).then(mergeAll)
promise.requests = // the requests promises
```
it lets the user to await for the final result by using
```TypeScript
const finalResult = await promise;
```
or await for each individual request
```TypeScript
await promise.requests["request-key-one"]
```
We have it because poll is a batch request. This means you can await for the entire result to be ready or just await for a specific request.

packages/libs/client/src/v2/client/client.ts
javadkh2: I will rename it to `mergedPollRequestPromises` actually we just return it here and let the user to decide await for that or await for each request separately

packages/libs/client/src/v2/client/client.ts
javadkh2: I think it's a good idea to add "listen" as well. I believe they would have different use cases:"poll" for batch listening, and "listen" for a specific request key.

packages/libs/client/src/v2/pact.ts
javadkh2: hahah

packages/libs/client/src/v2/pact.ts
javadkh2: Yes its more flexible and I have tried this approach, but I couldn't find a way to write a proper typing for that
Actually the most difficult typing is capabilities in signer, that should refer types from the payload and only suggest the valid capabilities. So I had to add some limitations like forcing the first argument to be the payload. 
so basically 
```typescript
commandBuilder(
payload.exec([
  coin.transfer(from, to, { decimal: '0.01' })
]),
// withCapability comes you the proper typings 
 signer(from, (withCapability) => [
      withCapability('coin.TRANSFER', from, to, { decimal: amount }),
      withCapability('coin.TRANSFER', from, to, { decimal: '0.01' }),
    ]),
)
```

packages/libs/client/src/v2/pact.ts
javadkh2: this function infers input types from return type 🙃
I have split its interface from the implementation.  check `ISigner`
its something like this

```typeScript
const myFn = <T extends any>(cb: (arg: T) => string): T =>
  ((arg: any) => arg) as any;

const builder = (arg: { name: string; args: any[] }) => {};

// myFn argument has the same type as builder argument
builder(myFn((arg: { name: string; args: any[] }) => arg.name));
```

packages/libs/client/src/v2/pact.ts
javadkh2: yes I removed it

packages/libs/client/src/v2/pact.ts
javadkh2: I have removed the restriction, so now users can compose multiple commands easily. However, the command builder still returns the ICommand object. Although I didn't implement laziness, the input arguments can still be functions that receive the current state, such as the `signer`.

I have check the typing again, actually it would be technically straightforward to implement laziness. However, based on the current requirements, there doesn't seem to be a necessity for it since we have divided the signing and submitting steps.

Your example does not need lazyness as all of the helpers can just return the partial command and don't need to wait for any input.
```TypeScript
setMeta(devnetInfo) // probably just returns the {meta: ... }
setSigner(devnetAccount) // can return {signers:[...]}
getTransactionCreate(from, to, keysetAttributeInData, amount) // can return {payload:...}
setData(dataWithKeyset) // can set the {payload:data{...}}
```
The only concern I have about implementing laziness is that it would require all users to make two function calls in a row for typical use cases , and I'm unsure how often this is needed.
```
const command = commandBuilder(
 .....
)(/ send initial state/)
```
@alber70g any idea?

EnoF: In worse case you could introduce a `commandAndSend` that directly executes it right?
```
const commandAndSend = (
  .....
  );
  ```

alber70g (approved):

alber70g:
Some context on this PR for public viewers.

This PR has had extended discussions internally. From the API design to the 3-phase approach. Some of those are documented in various issues (in particular #139), and in the issues mentioned in [the milestone for v1.0.0](https://github.com/kadena-community/kadena.js/milestone/1).

We've chosen to do a so-called "big bang" release, as we didn't want anyone to be blocked by releasing a half-finished and non-feature-parity library.

packages/libs/client-examples/src/example-contract/safe-transfe.ts
lkononenko, line 1: typo in the file name - missing `r` at the end

PR #477, opened 2023-06-26
[@kadena/transfer] New Layout

rezaerami:


packages/apps/transfer/src/utils/styles.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (typeof prop === 'number') {
+    return {
+      width: getPropInPercentage(prop),
+    };
+  }
+
+  const { size, span, offset, pull, push, hidden } = prop;
+  const styles: IGridColMediaStyles = {};
+
+  if(size)  styles.width = getPropInPercentage(size).toString();
```
rezaerami, line 24: needs to be done like this to make sure we don't have undefined properties in the result

packages/apps/transfer/src/utils/styles.ts
```diff
@@ -0,0 +1,32 @@
...
+  if(span)  styles.flex = `0 0 ${getPropInPercentage(span)}`;
+  if(offset)  styles.marginLeft = getPropInPercentage(offset).toString();
+  if(push)  styles.marginRight = getPropInPercentage(push).toString();
+  if(pull)  styles.order = pull.toString();
+  if(hidden !== undefined)  styles.display = hidden ? 'none' : "unset";
```
rezaerami, line 29: only if hidden is passed id needs to check the value!

packages/apps/transfer/src/components/Global/Grid/GridCol.tsx
```diff
@@ -0,0 +1,34 @@
...
+  sm?: IGridColMediaType,
+  md?: IGridColMediaType,
+  lg?: IGridColMediaType,
+  xl?: IGridColMediaType,
+  xxl?: IGridColMediaType,
+  className?: string
+  children: ReactNode
+}
+
+const GridCol: FC<IGridRowProps> = ({children, className, xs = {}, sm={}, md={}, lg={}, xl={}, xxl = {}}) => {
```
rezaerami, line 19: can't make it dynamic cause it'll be a runtime type checker so the build will fail

packages/apps/transfer/package.json
```diff
@@ -21,9 +21,11 @@
     "@kadena/chainweb-node-client": "workspace:*",
     "@kadena/client": "workspace:*",
     "@kadena/cryptography-utils": "workspace:*",
+    "@kadena/kadena-docs": "workspace:*",
     "@kadena/pactjs": "workspace:*",
     "@kadena/pactjs-cli": "workspace:*",
     "@kadena/react-components": "workspace:*",
+    "@kadena/react-ui": "workspace:*",
     "@radix-ui/colors": "~0.1.8",
     "@stitches/react": "1.3.1-1",
```
MRVDH, line 30: Can't stitches be removed with the new layout?

packages/apps/transfer/src/components/Common/Layout/index.tsx
```diff
@@ -1,14 +1,52 @@
+import {Header} from "./partials";
+
 import { StyledLayout } from '@/components/Common/Layout/styles';
+import {WalletConnectButton} from "@/components/Global";
+import {Logo} from "@/resources/svg/generated";
+import useTranslation from "next-translate/useTranslation";
 import React, { type ReactNode, FC } from 'react';
+import routes from "../../../constants/routes";
```
MRVDH: absolute routes

packages/apps/transfer/src/components/Common/Layout/index.tsx
```diff
@@ -1,14 +1,52 @@
+import {Header} from "./partials";
+
 import { StyledLayout } from '@/components/Common/Layout/styles';
+import {WalletConnectButton} from "@/components/Global";
```
MRVDH: spacing should be fixed by the linter probably

packages/apps/transfer/package.json
rezaerami: nope, we use old components also, until we remove all dependencies to old react components we need to keep them

packages/apps/transfer/package.json
```diff
@@ -21,9 +21,11 @@
     "@kadena/chainweb-node-client": "workspace:*",
     "@kadena/client": "workspace:*",
     "@kadena/cryptography-utils": "workspace:*",
+    "@kadena/kadena-docs": "workspace:*",
```
webpro, line 24: Do we need/use this dependency (app in app)?

webpro (commented):
Please run `rush[x] format` on updated packages.

packages/apps/transfer/package.json
rezaerami: nope, i was trying to get docs component, thanks for reminding me, removed

alber70g (approved):

MRVDH (approved):

PR #480, opened 2023-06-26
Link component (vanilla-extract)

bearni95:
Implementation of the `<Link>` component.

It accepts the following parameters:
-   href
-  target (one of:  _blank, _self, _parent, _top')
-  children
-   iconPosition (one of: left, right)

[Asana Ticket](https://app.asana.com/0/1204808093245015/1204896526017318/f)

### DIFF
--- a/packages/libs/react-ui/src/components/Link/Link.css.ts
+++ b/packages/libs/react-ui/src/components/Link/Link.css.ts
@@ -1,0 +1,32 @@
+import { ColorType, sprinkles, vars } from '../../styles';
+
+import { createVar, style, styleVariants } from '@vanilla-extract/css';
+
+export const linkContainerClass = style([
+  sprinkles({
+    display: 'flex',
+  }),
+  {
+    selectors: {
+      '&:hover': {
+        textDecoration: 'none',
+      },
+      '&:active': {
+        color: vars.colors.$negativeContrast,
+      },
+      '&:visited': {
+        color: '#C870FF',
+      },
+    },
+  },
+]);
+
+export const iconContainerClass = style([
+  sprinkles({
+    display: 'flex',
+    marginX: '$2',
+  }),
+  {
+    textDecoration: 'none',
+  },
+]);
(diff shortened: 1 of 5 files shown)


packages/libs/react-ui/src/components/Link/Link.css.ts
```diff
@@ -0,0 +1,32 @@
...
+        color: '#C870FF',
+      },
+    },
+  },
+]);
+
+export const iconContainerClass = style([
+  sprinkles({
+    display: 'flex',
+    marginX: '$2',
```
eileenmguo, line 27: I think it would be better to apply positioning in the parent using a gap so that there isn't margin on both sides of the icon

packages/libs/react-ui/src/components/Link/Link.css.ts
```diff
@@ -0,0 +1,32 @@
...
+  {
+    selectors: {
+      '&:hover': {
+        textDecoration: 'none',
+      },
+      '&:active': {
+        color: vars.colors.$negativeContrast,
+      },
+      '&:visited': {
+        color: '#C870FF',
```
eileenmguo, line 18: We shouldn't be using strings like this for colors. If we need a new color it should be discussed with @isaozler if we can use an existing color, or to discuss how this fits into the design system

packages/libs/react-ui/src/components/Link/Link.css.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+  },
+]);
+
+export const iconContainerClass = style([
+  sprinkles({
+    display: 'flex',
+  }),
+  {
+    textDecoration: 'none',
```
eileenmguo, line 31: Can you move this to sprinkles?

packages/libs/react-ui/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { iconContainerClass, linkContainerClass } from './Link.css';
+
+import React from 'react';
+
+export interface ILinkProps {
+  href: string;
+  target?: '_blank' | '_self' | '_parent' | '_top';
+  children: React.ReactNode;
+  iconPosition?: 'left' | 'right';
+  hasIcon?: boolean;
```
eileenmguo, line 12: We should follow the same subcomponent pattern used in the button. So instead of passing hasIcon and IconPosition, expose a Link.Icon component that the user can position

PR #481, opened 2023-06-27
start react-ui tabs component

sstraatemans:


packages/libs/react-ui/src/components/Tabs/Tabs.css.ts
```diff
@@ -0,0 +1,50 @@
...
+    position: 'absolute',
+    display: 'block',
+    backgroundColor: '$primaryAccent',
+  }),
+  {
+    bottom: '-4px', // for some reason a negative cant be done with vars
+    transition: 'all .4s ease',
+    transform: `translateX(0)`,
+    width: '0',
+    height: vars.sizes.$1,
```
eileenmguo, line 48: sprinkles

packages/libs/react-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+  children?: ReactNode;
+  defaultSelected?: string;
+}
+
+interface ITabsComposition extends FC<ITabsProps> {
+  Tab: typeof Tab;
+  Panels: typeof TabsPanels;
+}
+
+// eslint-disable-next-line react/prop-types
```
eileenmguo, line 17: You should be able to remove this by assigning the type

eileenmguo: Ahh I see what is happening. Discussed this with @bearni95 that we should not be typing the "root" component as a composition because the type evaluation is not correct. The "container"/"root" component is still just a component, not a composition of components so we don't want to assign it an incorrect type and have to cast or disable errors. We've decided to go with the approach that radix takes which is naming the parent component `Tabs.Root`

packages/libs/react-ui/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const items = ItemArray.slice(0, itemsCount);
+
+    return (
+      <>
+        <Tabs defaultSelected={defaultSelected}>
+          {items.map((item) => {
+            return (
+              <Tabs.Tab key={item} value={item}>
+                {item}
+              </Tabs.Tab>
```
eileenmguo, line 48: Should these be in a list?

packages/libs/react-ui/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+            );
+          })}
+
+          <Tabs.Panels>
+            {items.map((item) => {
+              return (
+                <Tabs.Panels.Panel key={item} value={item}>
```
eileenmguo, line 55: So we also want to avoid these very nested components for readability. I would even go as far as simplying this to just Tabs.Content and only showing the content for the selected tab. Similar to this api from radix https://www.radix-ui.com/docs/primitives/components/tabs

packages/libs/react-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const Tabs: ITabsComposition = ({ children, defaultSelected = '' }) => {
+  const [selectedTab, setSelectedTab] = useState<string>(defaultSelected);
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  const lineRef = useRef<HTMLSpanElement | null>(null);
```
eileenmguo, line 21: I think a more descriptive name could improve this. Maybe selectedUnderline or something

packages/libs/react-ui/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+          if (child.type === Tab) {
+            const props = {
+              ...child.props,
+              selected: selectedTab === child.props.value,
+              handleClick,
+            };
+            return React.cloneElement(child, props);
+          }
+          return null;
+        })}
```
eileenmguo, line 58: Could this be removed by just using the React.FunctionComponentElement type?

packages/libs/react-ui/src/components/Tabs/Tabs.tsx
sstraatemans: got it

packages/libs/react-ui/src/components/Tabs/Tabs.stories.tsx
sstraatemans: that would just be an extra step.
but thought about it. and for consistancy, I removed the PanelsList.
not needed. all is taken care of in the Container component

packages/libs/react-ui/src/components/Tabs/Tabs.stories.tsx
sstraatemans: makes sense

sstraatemans: removed the nesting.

eileenmguo (approved):

eileenmguo (approved):

PR #486, opened 2023-06-27
[@kadena/transfer] Fix input defaults and env vars issue

KristinaSpasevska:


packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
```diff
@@ -27,7 +27,7 @@ export async function finishXChainTransfer(
   sender: string,
 ): Promise<ContCommand | { error: string }> {
   const host = generateApiHost(server, network, chainId);
-  const hostSPV = `${generateApiHost(server, network, '1')}/spv`;
+  const hostSPV = `${generateApiHost(server, network, '1')}`;
```
sanderlooijenga, line 30: Can now just be `const hostSPV = generateApiHost(server, network, '1')`, right?

EnoF, line 30: why is the SPV host different than the host? They should be identical

sanderlooijenga (approved):

packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
KristinaSpasevska: The SPV host is different cause it only has option '1' as chain (at least that's what we know so far, so please correct me if I'm wrong). The host has the chainId from the actual transaction and may be different than '1'.
The `getCommand` function needs the SPV host as a parameter, and for the transfer finish we need the host.

nil-amrutlal-dept (approved):

packages/apps/transfer/src/services/cross-chain-transfer-finish/finish-xchain-transfer.ts
EnoF: An SPV proof should be retrieved from the same chain the initial transaction has taken place. So if I initiated a x-chain transfer from chain 14, my SPV proof should be retrieved from chain 14. So i'm pretty sure that this will not work on a xchain transfer happening from 14 -> 15 for example.

PR #488, opened 2023-06-27
[@kadena/transfer] Rewrite constants

MRVDH:

### DIFF
--- a/packages/apps/transfer/.env.example
+++ b/packages/apps/transfer/.env.example
@@ -1,8 +1,11 @@
-# Kadena main network, omit to use default value: api.chainweb.com
+# Kadena networks, omit to use default value: api.chainweb.com
 KADENA_MAINNET_API=
-# Kadena Networks, omit to use default value
 KADENA_TESTNET_API=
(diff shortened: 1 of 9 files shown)


packages/apps/transfer/src/pages/transfer/module-explorer/index.tsx
```diff
@@ -47,8 +47,7 @@ const GetCode: FC = () => {
       const data = await describeModule(
         moduleName,
         convertIntToChainId(moduleChain),
-        chainNetwork[network].network,
-        chainNetwork[network].server,
+        network,
         'not-real',
```
nil-amrutlal-dept, line 51: Maybe this could also be a constant. For context, when we send the requests for the module explorer we set the sender as 'not-real'.

nil-amrutlal-dept (approved):

packages/apps/transfer/src/pages/transfer/module-explorer/index.tsx
MRVDH: Solved 👍

PR #494, opened 2023-06-27
Housekeeping what's being npm-published

webpro:
Saw #238 and I think that one (and our current setup included) is more complicated than necessary. Using a whitelist approach with `package.json#files` is more straightforward, imho. Also co-located to build scripts and so on in `package.json`.

I think `.npmignore` is useful for specific file patterns to ignore with directories we do want to publish.

So my suggestion would be to close #238 and keep it simple.


alber70g:
> Using a whitelist approach with package.json#files is more straightforward
@webpro For `rush publish` there's already a config that only includes `lib` and some other directories. Let's check that, and make sure we incorporate that

webpro:
@alber70g I think `publishFolder` is not the same as `outputFolderNames` (the latter is for caching purposes). I guess `temp` folders shouldn't be published to the registry (not seeing a `lib-commonjs` folder anywhere after building everything). I assume/hope `rush publish` basically does `npm publish` under the hood, or in some other way takes information from `package.json`, `.npmrc`, `.npmignore` etc. into account. Are there any specifics you're concerned about?

alber70g (approved):

PR #498, opened 2023-06-28
add eslint rules

rezaerami:
- add new rules to help with code style
- add prettier and prettier rc to sync IDE with project rules
- fix code style of files with new standard


packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
+const fs = require('fs');
+
+const prettierOptions = JSON.parse(fs.readFileSync('./.prettierrc', 'utf8'));
```
rezaerami, line 3: importing prettier this way helps to sync IDEs such as webstorm which has it's own formatted with project standards
cause now both IDE and linter will read prettier config from same config file

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
 
 module.exports = {
-  plugins: ['jest'],
+  plugins: ['prettier', 'jest'],
   extends: ['next/core-web-vitals', '@kadena-dev/eslint-config/profile/react'],
   parserOptions: { tsconfigRootDir: __dirname },
   ignorePatterns: ['**/*.spec.ts', '**/*.spec.tsx', '**/generated/**'],
+  rules: {
+    'prettier/prettier': ['error', prettierOptions],
+    '@typescript-eslint/strict-boolean-expressions': 0,
```
rezaerami, line 13: disable boolean expressions

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'arrow-body-style': [2, 'as-needed'],
```
rezaerami, line 14: simplify arrow function and use parentheses instead of return keyword when function only returns 
e.g.
```
// turns this
const foo = () => {
    return "bar";
}
// to this
const foo = () => "bar"
```

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'comma-dangle': [2, 'always-multiline'],
```
rezaerami, line 15: adds extra comma to last item of an object or array if it's multiline
```
const foo = [
    "foo",
]
```

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'prefer-template': 2,
```
rezaerami, line 16: prefers to use templates instead of string concatenates
```
// turns this
const foo = "hi" + name;
// to this
const foo = `hi ${name}`;
```

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'import/newline-after-import': 1,
+    'import/no-unresolved': 2,
```
rezaerami, line 18: this is super useful especially for windows users with a repo manager
since we have a rush, on windows we'll have root node modules, therefore if a package is not listed in the package.json of one of the packages, it'll allow you to use it since you have it in root, but then on deployment it figures it out and it.
but with this rule, if you import a package, it has to be listed in package.json

this issue happens often on windows machines

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'jsx-a11y/aria-props': 2,
```
rezaerami, line 19: JSX-ally rules are for accessibility 
forcing user to use correct roles in case they used aria-*, also they should be valid regarding the element, for example an anchor IF had aria props, then it should have role of a link

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'jsx-a11y/role-has-required-aria-props': 2,
+    'jsx-a11y/heading-has-content': 1,
+    'jsx-a11y/mouse-events-have-key-events': 2,
```
rezaerami, line 22: to improve accessibility for for users without a mouse or touchpad this is also really useful, it forces you to have the same value for onFocus and onBlur, in case you used onMouseOver and onMouseOut

packages/apps/transfer/.eslintrc.js
```diff
@@ -1,8 +1,32 @@
...
+    'jsx-a11y/role-supports-aria-props': 2,
+    'react/function-component-definition': [
```
rezaerami, line 24: prefers to have arrow functions for components

packages/apps/transfer/.eslintrc.js
sstraatemans: hm would prefer to be free in your choice with this

alber70g (changes requested):
Let's move all the rules to a next-specific config in the @kadena-dev/eslint-config project.

Also, as per conversation, we can remove some rules _only if_ it's possible to have those rules removed for `jsx` parts only. Anything not-ui related should adhere to the stricter ruleset as defined in the profiles currently

webpro:
> Let's move all the rules to a next-specific config in the @kadena-dev/eslint-config project.

Perhaps "react" or "jsx" or "ui" or "loose" is more targeted/clear. Applies to UI logic, design system and React/JSX components in general.

> Also, as per conversation, we can remove some rules _only if_ it's possible to have those rules removed for `jsx` parts only. Anything not-ui related should adhere to the stricter ruleset as defined in the profiles currently

Maybe it could be as simple as applying the new rules to all `.tsx` files?

rezaerami:
> 

yep it's totally possible tolimit certain rules to certain file formats
e.g.
```

  "overrides": [
    {
      "files": ["*.jsx"],
      "rules": {
        "@typescript-eslint/strict-boolean-expressions": "off"
      }
    }
  ],
  ```

**Comment by rezaerami** [comment]
> Maybe it could be as simple as applying the new rules to all .tsx files?
well, some of them such as arrow function prettier and import ones are not tsx-only
but indeed, we can apply some rules such as jsx ally ones only to jsx files, although when you have jsx code IDE will suggest you to convert it to .tsx | .jsx

therefore if there is no jsx code, those rules won't apply, and if there is, then the file extension will be tsx | jsx for sure

**Comment by alber70g** [comment]
We discussed this offline and @rezaerami will move most of the rules to the `@kadena-dev/eslint-config`, to either the `lib`, `react` or `next` config.

**Review by MRVDH — APPROVED** [review]

File: packages/apps/transfer/.eslintrc.js [hunk 10]
```diff
@@ -2,7 +2,7 @@ require('@rushstack/eslint-config/patch/modern-module-resolution');
 
 module.exports = {
   plugins: ['jest'],
-  extends: ['next/core-web-vitals', '@kadena-dev/eslint-config/profile/react'],
+  extends: ['@kadena-dev/eslint-config/profile/next'],
   parserOptions: { tsconfigRootDir: __dirname },
-  ignorePatterns: ['**/*.spec.ts', '**/*.spec.tsx', '**/generated/**'],
+  ignorePatterns: ['**/generated/**'],
```
alber70g, line 7: Does removing this still work with VSCode ESLint Plugin? I had issues with the base config not being able to resolve the plugins from next

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-tracker/index.tsx
```diff
@@ -1,12 +1,10 @@
...
-    <MainLayout title="Kadena Coin Transfer">
-      <h3>TO DO: Cross Chain Transfer Tracker</h3>
-    </MainLayout>
-  );
-};
+const CrossChainTransferTracker: FC = () => (
+  <MainLayout title="Kadena Coin Transfer">
+    <h3>TO DO: Cross Chain Transfer Tracker</h3>
+  </MainLayout>
+);
```
alber70g, line 8: It seems like we have decided to keep the code as it is after our discussion.

I.e.: when its written like `const x = () => { return "" }` it would **not** be rewritten to `const x = () => ""`

packages/apps/transfer/src/utils/env.ts
```diff
@@ -11,6 +11,8 @@ export interface IEnvInterface {
   DEFAULT_SENDER?: string;
 }
 
+type IRequiredEnv = Required<IEnvInterface>;
```
alber70g, line 14: types shouldn't be prefixed with `I`

packages/libs/react-components/package.json
```diff
@@ -57,6 +57,7 @@
     "@types/node": "^16.0.0",
     "@types/react": "^18.0.30",
     "eslint": "^8.15.0",
+    "eslint-import-resolver-alias": "~1.1.2",
```
alber70g, line 60: Why is this added? Can't we add it to the root config? Or will that not work with IDEs?

packages/tools/eslint-config/package.json
```diff
@@ -27,6 +27,10 @@
   "devDependencies": {
     "@rushstack/eslint-config": "~3.3.0",
     "eslint": "^8.15.0",
+    "eslint-config-prettier": "~8.8.0",
+    "eslint-plugin-import": "^2.26.0",
+    "eslint-plugin-jsx-a11y": "~6.7.1",
+    "eslint-plugin-prettier": "~4.2.1",
```
alber70g, line 33: Will adding them to `devDependencies` only work when this config is used in other packages? Do those packages need to add those dependencies themselves, or can we move this to `dependencies` and will it work as well? Or will it _not_ anyway, whether we put them as regular- _or_ devdependencies?

packages/tools/eslint-config/profile/lib.js
```diff
@@ -1,3 +1,10 @@
+const fs = require('fs');
+const path = require('path');
+
+const prettierOptions = JSON.parse(
+  fs.readFileSync(path.resolve(__dirname, '../../../../.prettierrc'), 'utf8'),
+);
+
```
alber70g, line 7: Can we add a comment that this is specific to our monorepo where, according to `rush.json#projectFolderMin/MaxDepth`, a package will **always** be in the 3rd nested directory (`packages/{apps,libs,tools}/package-name/*`). 

Should we also add to the comment that a more generic alternative could be to use:

```js
const prettierrcPath = `${require('child_process')
  .execSync('git rev-parse --show-toplevel')
  .toString()
  .replace(/\n/, '')}/.prettierrc`;
```

packages/tools/eslint-config/profile/lib.js
alber70g, line 1: Just noticed there's a ESLint rule for `sort-package-json`. Maybe we want to add it as an eslint rule instead and remove the package?

https://github.com/kellyselden/eslint-plugin-json-files#supported-rules:~:text=json%2Dfiles/sort%2Dpackage%2Djson

packages/tools/eslint-config/profile/react.js
```diff
@@ -1,19 +1,41 @@
...
+      },
+    ],
   },
   settings: {
     react: {
       version: 'detect',
     },
   },
+  // turned off boolean expressions for jsx and tsx files, so we can have non-boolean statements in react components
+  // this rule should be enabled for js and ts files even in a ReactJS or NextJS setup
```
alber70g, line 30:
```suggestion
  // disable strict-boolean-expressions for jsx and tsx files
  // this allows for truthy/falsy statements in react components
  // this rule must be enabled for js and ts files even in a ReactJS or NextJS setup
```

packages/tools/kda-cli/.eslintrc.cjs
```diff
@@ -4,4 +4,7 @@ require('@rushstack/eslint-config/patch/modern-module-resolution');
 module.exports = {
   extends: ['@kadena-dev/eslint-config/profile/lib'],
   parserOptions: { tsconfigRootDir: __dirname },
+  rules: {
+    'import/no-unresolved': 'off',
+  },
```
alber70g, line 9: Why do we have this? If there's a legitimate reason, let's document that in a comment

alber70g (changes requested):
Some questions and additions

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-tracker/index.tsx
rezaerami: ah no, this is coming from previous linter when I enabled that rule, now it's not enabled so it won't effect anything, this is just an old change

packages/apps/transfer/src/utils/env.ts
rezaerami: fixed

packages/tools/eslint-config/package.json
rezaerami: no, it will work both in dependencies and devDependencies
the reason I have put them in devDependencies is just that they are dev packages

packages/tools/eslint-config/profile/lib.js
rezaerami: done

packages/tools/eslint-config/profile/react.js
rezaerami: done

packages/tools/kda-cli/.eslintrc.cjs
rezaerami: yes, the reason is in kda-cli we are addressing ts files with .js format which I don't know the reason but when I try to remove the format it doesn't build it anymore!
for example this is an import within kda-cli
`import { IAnswers, IQuestion } from '../questions/questions.js';`

even though the extension for questions file is `.ts`

because of that all the imports will have unresolved warning, so I turned it off for this module

packages/libs/react-components/package.json
rezaerami: it's unused, removed

packages/tools/eslint-config/profile/lib.js
rezaerami: i will do that in a different PR since it'll effect all package jsons

packages/apps/transfer/.eslintrc.js
rezaerami: yes this is totally fine since we added tests on purpose to eslint ignore just for transfer since we had some difficulties, but now they are resolved and there's no need to ignore tests anymore, just like other next setups in monorepo

packages/apps/transfer/src/hooks/use-did-update-effect.ts
```diff
@@ -1,6 +1,7 @@
 import { useEffect, useRef } from 'react';
 
-export default (fn: Function, inputs: any[]) => {
+// eslint-disable-next-line @typescript-eslint/ban-types
+export default (fn: Function, inputs: unknown[]): boolean => {
```
alber70g, line 4: We can type this easily as `() => void`.

```suggestion
export default (fn: () => void, inputs: unknown[]): boolean => {
```

alber70g (approved):

packages/apps/transfer/src/hooks/use-did-update-effect.ts
rezaerami: done

packages/tools/eslint-config/package.json
alber70g: Yes but for the eslint-config they're actual dependencies, as you install `@kadena/eslint-config` for the configuration specifically

PR #499, opened 2023-06-28
Remove obsolete files

webpro:
Title says it all!

- We no longer use the all contributors thing that way (just an image in root README)
- "RandyModule" and "MyPactModule" doesn't sound like it concerns everyone in a public repo :)


MRVDH (approved):

Randynamic (approved):

Randynamic:
Are we still using the plugin? or should that be removed as well?

PR #504, opened 2023-06-29
[@kadena/transfer] Fix(transfer): empty number env vars

KristinaSpasevska:


packages/apps/transfer/src/utils/env.ts
```diff
@@ -1,3 +1,5 @@
+import { findDOMNode } from 'react-dom';
```
sanderlooijenga, line 1: import needed?

KristinaSpasevska: removed 👍

sanderlooijenga (approved):

nil-amrutlal-dept (approved):

PR #505, opened 2023-06-29
feat(docs): style the searchmodal and page

sstraatemans:


isaozler (approved):

packages/libs/react-ui/src/components/Modal/StoryComponents.tsx
```diff
@@ -10,102 +8,125 @@ import React, { FC } from 'react';
...
+
+export const Content: FC<IProps> = ({ title }) => {
   const { renderModal } = useModal();
 
   return (
     <>
-      <Button onClick={() => renderModal(<ModalContent />)} title="openModal">
+      <Button
+        onClick={() => renderModal(<ModalContent />, title)}
+        title="openModal"
```
isaozler, line 55: later we could use a prop for this as well

PR #511, opened 2023-06-29
fix(react-ui): fix the button component

sstraatemans:


eileenmguo (approved):

rezaerami (approved):

bearni95 (commented):
We can probably get rid of this commented line now:

https://github.com/kadena-community/kadena.js/blob/ea94e9a99d5daaa370b774be0d13e824409600e5/packages/libs/react-ui/src/components/Button/Button.stories.tsx#L60

PR #514, opened 2023-06-29
[WIP] Marmalade website

lkononenko:
Marmalade has completely different brand and styling than Kadena, because of that Kadena component library is not used (neither react-ui or react-components).
I use CSS modules for scoping css and Sass. Vanilla-extract is not used as based on next docs https://nextjs.org/docs/app/building-your-application/styling/css-in-js CSS-in-JS is not supported yet for server components but in meantime Eileen tested vanilla-extract with server components and that works well so can be that will move in the future if will be needed.

Design for reference https://www.figma.com/proto/UHZJ7z8Kefokmnjru2cENd/marmalade-v2?page-id=2%3A82&type=design&node-id=2-30661&viewport=612%2C334%2C0.31&scaling=scale-down-width


lkononenko:
Closing this PR as Marmalade website has moved out of the monorepo to https://github.com/kadena-io/marmalade-website

PR #521, opened 2023-07-03
Grid component improvement

bearni95:
- [Closes the related Asana ticket](https://app.asana.com/0/1204808093245015/1204873953410537/f)
- Simplified implementation parameters for Grid and its item
- Added a responsive setter for the `column` property on `Grid.Root` and `columnSpan` for `Grid.Item`:
```typescript
<Grid.Root columns={{
      sm: 6,
      md: 12,
      lg: 12,
      xl: 12,
      xxl: 12,
    }}
>
[...]

<Grid.Item columnSpan={{
      sm: 6,
      md: 4,
      lg: 3,
    }}
>
```

### DIFF
--- a/packages/libs/react-ui/src/components/Grid/Grid.css.ts
+++ b/packages/libs/react-ui/src/components/Grid/Grid.css.ts
@@ -1,11 +1,11 @@
 import { sprinkles } from '../../styles';
+import { breakpoints } from '../../styles/sprinkles.css';
 
 import { style, styleVariants } from '@vanilla-extract/css';
 
 export const gridContainerClass = style([
   {
     display: 'grid',
-    gridTemplateColumns: 'repeat(12, 1fr)',
   },
 ]);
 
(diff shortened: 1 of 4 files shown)


packages/libs/react-ui/src/components/Grid/GridItem.tsx
```diff
@@ -1,40 +1,54 @@
...
+      itemColumnVariants.lg[columnSpan.lg],
+    columnSpan &&
+      typeof columnSpan !== 'number' &&
+      columnSpan.xl &&
+      itemColumnVariants.xl[columnSpan.xl],
+    columnSpan &&
+      typeof columnSpan !== 'number' &&
+      columnSpan.xxl &&
+      itemColumnVariants.xxl[columnSpan.xxl],
+  );
```
timoheddes, line 49: Is there a way to make this a bit more readable and easy to maintain?

packages/libs/react-ui/src/components/Grid/GridRoot.tsx
```diff
@@ -0,0 +1,56 @@
...
+      columns.lg &&
+      containerColumnVariants.lg[columns.lg],
+    columns &&
+      typeof columns !== 'number' &&
+      columns.xl &&
+      containerColumnVariants.xl[columns.xl],
+    columns &&
+      typeof columns !== 'number' &&
+      columns.xxl &&
+      containerColumnVariants.xxl[columns.xxl],
```
timoheddes, line 47: And this one too. Low-effort comment as I don't have a better solution for you right now but just checking :)

packages/libs/react-ui/src/components/Grid/Grid.css.ts
```diff
@@ -28,3 +28,123 @@ export const gapVariants = styleVariants({
...
+  7: 7,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  8: 8,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  9: 9,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  10: 10,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  11: 11,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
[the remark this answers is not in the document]
sstraatemans: in the library we use `$1` as a convention we use in the lib.
Can we use that here?

packages/libs/react-ui/src/components/Table/Table.css.ts
```diff
@@ -1,6 +1,7 @@
 import { sprinkles, vars } from '../../styles';
 
 import { style } from '@vanilla-extract/css';
+
```
sstraatemans, line 4: 🔥 :)

packages/libs/react-ui/src/components/Grid/GridRoot.tsx
sstraatemans: make a small function?
`const isVariant = (columns, size) => columns &&
      typeof columns !== 'number' &&
      columns[size] &&
      containerColumnVariants[size][columns[size]],`

packages/libs/react-ui/src/components/Grid/Grid.css.ts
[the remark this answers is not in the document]
bearni95: My reasoning for using plain numbers instead of $-prefixed ones is that, in VE variables, the numbers represent a count for the measuring ($3 = 0.75rem = 12px) whereas in here the numbers represent the weight, out of 12, that each column should take.

packages/libs/react-ui/src/components/Grid/GridItem.tsx
eileenmguo: Yeah, I think there should be some way to make this easier to read and maintain since it is quite repetitive. Maybe pull it into a function?

eileenmguo (commented):
Is this the intended behavior? 
image

I think the storybook implementation is a bit confusing. Could we have a separate story for the grid.root and grid.item that showcases those specific props in a simplified setting in addition to the larger/more complex configuration?

packages/libs/react-ui/src/components/Grid/GridItem.tsx
bearni95: Moved both for `Grid.Root` and `Grid.Item` into a function:
https://github.com/kadena-community/kadena.js/blob/ceab74037d6140646af0f052e8519f0a12aec671/packages/libs/react-ui/src/components/Grid/GridItem.tsx#L18

eileenmguo (approved):

PR #527, opened 2023-07-04
Remove unused dependencies + fix some lint config (2)

webpro:
Basically a duplicate of #507. After `git fetch` this gives no diff:

```
git diff chore/remove-unused-dependencies..chore/remove-unused-dependencies-2 -- . ':(exclude)common/config/rush/pnpm-lock.yaml'
```

So there are only changes in the lockfile (not even in the repo state file).

The error in the other branch that does not reveal in this one: https://kadena-io.slack.com/archives/C04AL4NRWGP/p1688465334175419

I find this very odd :)

Will _greenify CI_ by updating the changelog.


Randynamic (approved):

packages/libs/bootstrap-lib/package.json
alber70g, line 1: Can we change the `.eslintrc` to use `@kadena-dev/eslint-config....lib` as `extends` option?

packages/apps/graph/package.json
```diff
@@ -44,7 +43,6 @@
     "@types/node": "^16.0.0",
     "embedded-postgres": "~15.1.0-beta.6",
     "prettier": "~2.8.8",
-    "sort-package-json": "~2.4.1",
-    "ts-node": "~10.8.2"
+    "sort-package-json": "~2.4.1"
```
alber70g, line 46: We can replace `sort-package-json` with https://github.com/kellyselden/eslint-plugin-json-files/blob/master/docs/rules/sort-package-json.md

packages/libs/kadena.js/.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 // This is a workaround for https://github.com/eslint/eslint/issues/3458
-require('@kadena-dev/eslint-config/profile/lib');
+require('@rushstack/eslint-config/patch/modern-module-resolution');
```
alber70g, line 2: Why was this changed?

alber70g (approved):
Question: can we also introduce linting for unused dependencies in this PR?

webpro:
> Question: can we also introduce linting for unused dependencies in this PR?

That's quite a thing. This monorepo has a lot of difficult to find unused dependencies, so even the mighty Knip lib has troubles and reports false positives. And it also requires a `package.json` in the root of the repo which we don't have. So I'm working on this, but the story isn't trivial.

jermaine150 (approved):

alber70g:
> > Question: can we also introduce linting for unused dependencies in this PR?
> 
> That's quite a thing. This monorepo has a lot of difficult to find unused dependencies, so even the mighty Knip lib has troubles and reports false positives. And it also requires a `package.json` in the root of the repo which we don't have. So I'm working on this, but the story isn't trivial.

Fair enough. What about having a whitelist of dependencies that shouldn't be touched, and a `--verify`? Something for the future I reckon

packages/libs/bootstrap-lib/package.json
webpro: Turns out this opens up a can of worms, so I'll pick this up after merging this one.

PR #532, opened 2023-07-05
Fix/poll until reject

sanderlooijenga:
Instead of `reject`ing the `pollUntil` with `this` (which is an instance of the `PactCommand` class), it would be beneficial to `reject` with the `result` of the failed request, so that one could e.g. also give more information in the UI regarding why it's failed.

### DIFF
--- a/packages/libs/client/src/pact.ts
+++ b/packages/libs/client/src/pact.ts
@@ -1,6 +1,7 @@
 import {
   ChainwebNetworkId,
   createSendRequest,
+  ICommandResult,
   IPollResponse,
   local,
   poll,
@@ -60,7 +61,7 @@ export interface ICommandBuilder<
         pollRequest: Promise<IPollResponse>,
       ) => void;
     },
-  ): Promise<this>;
+  ): Promise<ICommandResult>;
   poll(apiHost: string): Promise<IPollResponse>;
   addSignatures(
     ...sig: {
@@ -365,7 +366,7 @@ export class PactCommand
         pollRequest: Promise<IPollResponse>,
       ) => void;
     },
-  ): Promise<this> {
+  ): Promise<ICommandResult> {
     if (this.requestKey === undefined) {
       throw new Error('`requestKey` not found');
     }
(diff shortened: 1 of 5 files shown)


packages/libs/client/src/pact.ts
sanderlooijenga, line 1: @ash-vd / @alber70g do you guys think we should also `resolve` with the `result`, or keep this as is? See [here](https://github.com/kadena-community/kadena.js/pull/532/files#diff-2796cf23c9298c278f1f0c75572f1361abc1896b5a528f3462912e2571c8d023R399)

alber70g: Yes seems good. We need to verify where we're using this code and fix that too. Other than that, nice!

ash-vd: I think we can resolve with `result` as well. Polling a transaction is almost definitely the last thing you're doing with a transaction, especially when it succeeds.

packages/libs/client/src/tests/pact.test.ts
```diff
@@ -354,9 +355,40 @@ describe('Pact proxy', () => {
     // expect the first argument to be a transaction with property cmd
     expect(onPoll.mock.lastCall[0].cmd).toBeDefined();
   });
-});
 
-//TODO: Add timeout test case
+  it('should reject when timeout of polling has been exceeded', async () => {
```
alber70g, line 359:
```suggestion
  it('rejects when timeout of polling has been exceeded', async () => {
```

ash-vd, line 359: Nit:
```suggestion
  it('rejects when the timeout of polling has been exceeded', async () => {
```

sanderlooijenga: Done

sanderlooijenga: Done

ash-vd (approved):

alber70g (approved):

PR #534, opened 2023-07-05
[@kadena/transfer] Implementation Debug Library

nil-amrutlal-dept:


packages/apps/transfer/src/pages/transfer/cross-chain-transfer-finisher/index.tsx
```diff
@@ -49,7 +49,7 @@ interface IPactResultError {
 }
 
 const CrossChainTransferFinisher: FC = () => {
-  const debug = Debug('XChain-Transfer-Finisher');
+  const debug = Debug('kadena-transfer:pages:cross-chain-transfer-finisher');
```
MRVDH, line 52: `kadena-transfer:pages:transfer:cross-chain-transfer-finisher`

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-tracker/index.tsx
```diff
@@ -41,6 +42,7 @@ const CrossChainTransferTracker: FC = () => {
   const { network } = useAppContext();
   const router = useRouter();
 
+  const debug = Debug('kadena-transfer:pages:cross-chain-transfer-tracker');
```
MRVDH, line 45: `kadena-transfer:pages:transfer:cross-chain-transfer-tracker`

packages/apps/transfer/src/services/faucet/index.ts
```diff
@@ -22,6 +23,8 @@ const FAUCET_PRIVATE_KEY = env(
   '<PROVIDE_FAUCET_PRIVATEKEY_HERE>',
 );
 
+const debug = Debug('kadena-transfer:services:faucet-index');
```
MRVDH, line 26: We can leave out the index part imo `kadena-transfer:services:faucet`

MRVDH (approved):
minor naming changes

sanderlooijenga (approved):

webpro:
The `debug` lib is kinda frozen. This is not a bad thing at all, it's mature and battle-tested. Still, I think it's good to stay critical and look at alternatives (e.g. to keep using the latest best practices and new Node.js features and performance improvements etc). Would be good to know if you considered anything else?

rezaerami (approved):

nil-amrutlal-dept:
> 

In this implementation, I did not really consider other alternatives since this library is already being used by some of the applications within the mono repository. I agree that options should be explored and will do so, but for now, I will merge this PR and if a better solution is found, I think its good to change the whole library at once

nil-amrutlal-dept:
> The `debug` lib is kinda frozen. This is not a bad thing at all, it's mature and battle-tested. Still, I think it's good to stay critical and look at alternatives (e.g. to keep using the latest best practices and new Node.js features and performance improvements etc). Would be good to know if you considered anything else?

In this implementation, I did not really consider other alternatives since this library is already being used by some of the applications within the mono repository. I agree that options should be explored and will do so, but for now, I will merge this PR and if a better solution is found, I think its good to change the whole library at once

PR #536, opened 2023-07-05
Feature/chainid type

sanderlooijenga:
This PR changes the `ChainwebChainId` type from the `@kadena/chainweb-node-client`. It's now also usable during runtime. To render a list of chain IDs before people would do something like;

`Array.from(Array(20)).map...`

Now you can loop through the `CHAINS` const whilst getting the correct type.

### DIFF
--- a/packages/libs/chainweb-node-client/src/interfaces/PactAPI.ts
+++ b/packages/libs/chainweb-node-client/src/interfaces/PactAPI.ts
@@ -1,5 +1,4 @@
 import type {
-  ChainId,
   IBase64Url,
   ICommand,
   IMetaData,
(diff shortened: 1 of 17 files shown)


packages/libs/types/src/PactCommand.ts
```diff
@@ -141,28 +141,35 @@ export type NetworkId = string | undefined;
 /**
  * @alpha
  */
-export type ChainId =
-  | ''
```
sanderlooijenga, line 145: @alber70g do you think we should still accept an empty string as a value for ChainId?

packages/libs/chainweb-node-client/src/interfaces/PactAPI.ts
alber70g, line 1: We actually want to get rid of the @kadena/types package in favour of packages like chainweb-node-client.

alber70g (approved):

alber70g (approved):

PR #538, opened 2023-07-05
[@kadena/transfer] Using new React UI components

MRVDH:
- Changed components to new react ui components.
- Fixed console warning of useLayoutEffect.
- Fixed new component dark theming.
- First implementation of stitches.


eileenmguo (approved):

nil-amrutlal-dept (approved):

rezaerami (approved):

KristinaSpasevska (approved):

packages/apps/transfer/src/pages/_document.tsx
```diff
@@ -3,10 +3,13 @@ import {
   getCssText,
   globalCss,
 } from '@kadena/react-components';
+import { darkThemeClass } from '@kadena/react-ui/theme';
 
 import { Head, Html, Main, NextScript } from 'next/document';
 import React, { FC, ReactNode } from 'react';
 
+React.useLayoutEffect = React.useEffect;
```
sanderlooijenga, line 11: Could you add a comment in code about why this is done?

packages/apps/transfer/src/components/Global/AccountText/styles.ts
```diff
@@ -29,12 +30,12 @@ export const StyledAccountContainer = styled('div', {
...
+export const StyledEyeOffIcon = styled(SystemIcon.EyeOffOutline, {
   width: '24px',
   height: '24px',
 });
 
-export const StyledEyeIcon = styled(SystemIcons.EyeOutline, {
+export const StyledEyeIcon = styled(SystemIcon.EyeOutline, {
   width: '24px',
   height: '24px',
 });
```
eileenmguo, line 41: I have a feeling the error might be coming from this. Could you try to use one of the predefined size props on the icon or add more sizes if necessary?

MRVDH: I tried things like $16 and $sm but both result in the same error.

eileenmguo: I meant removing these blocks completely and passing the size as a prop to the icon wherever it's used. The icons already have predefined sizes variants: sm, md, lg, xl, and heroHeader

So wherever it's being used you should be able to apply the 24px size using the `md` prop like this
```
<SystemIcon.EyeOutline size="sm" />
```

alber70g (approved):

packages/apps/transfer/src/pages/_document.tsx
MRVDH: Decided to remove this, it resolves warnings but not very positive about the side effects

PR #540, opened 2023-07-05
implement select component

rezaerami:
- add select component
- add story book
- add unit test

### DIFF
--- a/packages/libs/react-ui/src/components/Select/Option.tsx
+++ b/packages/libs/react-ui/src/components/Select/Option.tsx
@@ -4,10 +4,7 @@ import classNames from 'classnames';
 import React, { FC } from 'react';
 
 export interface IOptionProps
-  extends Omit<
-    React.HTMLAttributes<HTMLOptionElement>,
-    'as' | 'children' | 'className' | 'value'
-  > {
+  extends Omit<React.HTMLAttributes<HTMLOptionElement>, 'as'> {
   children: React.ReactNode;
   value: string[] | string | number;
 }
--- a/packages/libs/react-ui/src/components/Select/Option.tsx
+++ b/packages/libs/react-ui/src/components/Select/Option.tsx
@@ -1,16 +1,15 @@
 import { optionClass } from './Select.css';
 
-import classNames from 'classnames';
 import React, { FC } from 'react';
 
 export interface IOptionProps
   extends Omit<React.HTMLAttributes<HTMLOptionElement>, 'as'> {
   children: React.ReactNode;
-  value: string[] | string | number;
+  value?: string[] | string | number;
 }
 
 export const Option: FC<IOptionProps> = ({ children, ...rest }) => (
-  <option className={classNames(optionClass)} {...rest}>
+  <option className={optionClass} {...rest}>
     {children}
   </option>
 );
--- a/packages/libs/react-ui/src/components/Select/Option.tsx
+++ b/packages/libs/react-ui/src/components/Select/Option.tsx
@@ -1,0 +1,19 @@
+import { optionClass } from './Select.css';
+
+import classNames from 'classnames';
+import React, { FC } from 'react';
+
+export interface IOptionProps
+  extends Omit<
+    React.HTMLAttributes<HTMLOptionElement>,
+    'as' | 'children' | 'className' | 'value'
+  > {
+  children: React.ReactNode;
+  value: string[] | string | number;
+}
+
+export const Option: FC<IOptionProps> = ({ children, ...rest }) => (
+  <option className={classNames(optionClass)} {...rest}>
+    {children}
+  </option>
+);
--- a/packages/libs/react-ui/src/components/Select/Select.css.ts
+++ b/packages/libs/react-ui/src/components/Select/Select.css.ts
@@ -20,20 +20,17 @@ export const containerClassDisabled = style([
 ]);
 
 export const selectContainerClass = style([
-  sprinkles({}),
-  {
+  sprinkles({
     display: 'flex',
-  },
+  }),
 ]);
 
 export const iconClass = style([
   sprinkles({
     marginRight: '$2',
     marginLeft: '$2',
-  }),
-  {
     display: 'block',
-  },
+  }),
 ]);
 
 export const selectClass = style([
@@ -54,5 +51,4 @@ export const optionClass = style([
     backgroundColor: '$neutral1',
     color: '$neutral5',
   }),
-  {},
 ]);
--- a/packages/libs/react-ui/src/components/Select/Select.css.ts
+++ b/packages/libs/react-ui/src/components/Select/Select.css.ts
@@ -1,0 +1,58 @@
+import { sprinkles } from '../../styles';
+
+import { style } from '@vanilla-extract/css';
+
+export const containerClass = style([
+  sprinkles({
+    backgroundColor: '$neutral1',
+    color: '$neutral5',
+    padding: '$2',
+    borderRadius: '$sm',
+  }),
+  {
+    position: 'relative',
+    border: 'solid 1px',
+  },
+]);
+export const containerClassDisabled = style([
+  sprinkles({ pointerEvents: 'none' }),
+  { opacity: 0.5 },
+]);
+
+export const selectContainerClass = style([
+  sprinkles({}),
+  {
+    display: 'flex',
+  },
+]);
+
+export const iconClass = style([
+  sprinkles({
+    marginRight: '$2',
+    marginLeft: '$2',
+  }),
+  {
+    display: 'block',
+  },
+]);
+
+export const selectClass = style([
+  sprinkles({
+    border: 'none',
+    fontSize: '$base',
+  }),
+  {
+    padding: '0',
+    backgroundColor: 'inherit',
+    color: 'inherit',
+    flex: '1',
+  },
+]);
+
+export const optionClass = style([
+  sprinkles({
+    backgroundColor: '$neutral1',
+    color: '$neutral5',
+  }),
+  {},
+]);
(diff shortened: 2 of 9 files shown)


packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div className={selectContainerClass}>
+        {Icon && (
+          <span className={iconClass}>
+            <Icon size="md" />
+          </span>
+        )}
+        <select
+          ref={ref}
+          className={selectClass}
+          disabled={Boolean(disabled)}
```
sanderlooijenga, line 57: Does this really need to be cast to a boolean? The prop type should already enforce this, right?

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -0,0 +1,58 @@
...
+export const containerClassDisabled = style([
+  sprinkles({ pointerEvents: 'none' }),
+  { opacity: 0.5 },
+]);
+
+export const selectContainerClass = style([
+  sprinkles({}),
+  {
+    display: 'flex',
+  },
```
realdreamer, line 26:
```suggestion
  sprinkles({
    display: 'flex',
  }),
```

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -0,0 +1,58 @@
...
+]);
+
+export const iconClass = style([
+  sprinkles({
+    marginRight: '$2',
+    marginLeft: '$2',
+  }),
+  {
+    display: 'block',
+  },
```
realdreamer, line 36:
```suggestion
  sprinkles({
    marginRight: '$2',
    marginLeft: '$2',
    display: 'block',
  }),
```

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -0,0 +1,58 @@
...
+    flex: '1',
+  },
+]);
+
+export const optionClass = style([
+  sprinkles({
+    backgroundColor: '$neutral1',
+    color: '$neutral5',
+  }),
+  {},
```
realdreamer, line 57:
```suggestion
  sprinkles({
    backgroundColor: '$neutral1',
    color: '$neutral5',
  }),
```

packages/libs/react-ui/src/components/Select/Option.tsx
```diff
@@ -0,0 +1,19 @@
+import { optionClass } from './Select.css';
+
+import classNames from 'classnames';
+import React, { FC } from 'react';
+
+export interface IOptionProps
+  extends Omit<
+    React.HTMLAttributes<HTMLOptionElement>,
+    'as' | 'children' | 'className' | 'value'
```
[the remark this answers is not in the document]
rezaerami: technically only `as`  is needed since we are overwriting the rest of them, but removed the rest anyway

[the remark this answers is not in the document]
rezaerami: technically only `as`  is needed since we are overwriting the rest of them, but removed the rest anyway

packages/libs/react-ui/src/components/Select/Select.tsx
rezaerami: disabled is string in HTML so when we extend from HTML select element it expects to pass string

packages/libs/react-ui/src/components/Select/Option.tsx
```diff
@@ -0,0 +1,19 @@
...
+  > {
+  children: React.ReactNode;
+  value: string[] | string | number;
+}
+
+export const Option: FC<IOptionProps> = ({ children, ...rest }) => (
+  <option className={classNames(optionClass)} {...rest}>
```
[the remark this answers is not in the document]
rezaerami: well, following discussion we had before, we should not allow it right?

packages/libs/react-ui/src/components/Select/Select.css.ts
rezaerami: image

packages/libs/react-ui/src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,17 @@
+import { Select } from '@components/Select';
+import { Option } from '@components/Select/Option';
+import { render } from '@testing-library/react';
+import React from 'react';
+
+describe('Select', () => {
+  test('renders correctly', () => {
```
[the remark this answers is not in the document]
rezaerami: added more tests

packages/libs/react-ui/src/components/Select/Option.tsx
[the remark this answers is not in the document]
sanderlooijenga: I guess you're right. You could also still provide your own component as children to the Select component to circumvent this "issue".

sanderlooijenga (approved):

packages/libs/react-ui/src/components/Select/Option.tsx
```diff
@@ -0,0 +1,16 @@
...
+import React, { FC } from 'react';
+
+export interface IOptionProps
+  extends Omit<React.HTMLAttributes<HTMLOptionElement>, 'as'> {
+  children: React.ReactNode;
+  value: string[] | string | number;
+}
+
+export const Option: FC<IOptionProps> = ({ children, ...rest }) => (
+  <option className={classNames(optionClass)} {...rest}>
```
sstraatemans, line 13: classnames not needed

packages/libs/react-ui/src/components/Select/Option.tsx
```diff
@@ -0,0 +1,16 @@
+import { optionClass } from './Select.css';
+
+import classNames from 'classnames';
+import React, { FC } from 'react';
+
+export interface IOptionProps
+  extends Omit<React.HTMLAttributes<HTMLOptionElement>, 'as'> {
+  children: React.ReactNode;
+  value: string[] | string | number;
```
sstraatemans, line 9: value should not be required.
you want to have a defaultValue

packages/libs/react-ui/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+type Story = StoryObj<
+  {
+    icon: keyof typeof SystemIcon;
+  } & Omit<ISelectProps, 'icon'>
+>;
+
+export const Dynamic: Story = {
+  name: 'Select',
+  args: {
+    icon: undefined,
```
sstraatemans, line 43: can you add a value option.
So that people can see that you can set the default value

packages/libs/react-ui/src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+  it('renders without errors', () => {
+    const { getByTestId } = render(
+      <Select value="1" onChange={() => {}}>
+        <Option value="1">Option 1</Option>
+        <Option value="2">Option 2</Option>
+      </Select>,
+    );
+
+    const selectContainer = getByTestId('kda-select');
+    expect(selectContainer).toBeInTheDocument();
```
webpro, line 16: When doing the exact same thing in the next test, we don't need this one? We probably do this more often and I don't see any benefit tbh. One test doesn't matter much but it adds up.

rezaerami: this one just tests container and rest mostly focusing on select, options and functionality, but yes indeed this one doesn't add much

packages/libs/react-ui/src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    const selectContainer = getByTestId('kda-select');
+    const selectElement = selectContainer.querySelector('select');
+    const option1 = selectContainer.querySelector('option[value="1"]');
+    const option2 = selectContainer.querySelector('option[value="2"]');
+
+    expect(selectElement).toBeInTheDocument();
+    expect(option1).toBeInTheDocument();
+    expect(option2).toBeInTheDocument();
```
realdreamer, line 34: it's better to avoid normal DOM query selectors in test as much as possible. this could be written like below
```
expect(screen.getByRole('combobox')).toBeInTheDocument();
    expect(
      screen.getByRole('option', { name: 'Option 1' }),
    ).toBeInTheDocument();
    expect(
      screen.getByRole('option', { name: 'Option 2' }),
    ).toBeInTheDocument();
    expect(
      (screen.getByRole('option', { name: 'Option 1' }) as HTMLOptionElement)
        .selected,
    ).toBe(true);
```

the tests below as well could be re-written in same format.

rezaerami: I have tried but it doesn't work
image

also, it's a common practice to use query selector in jest, + this won't improve coverage or benefit in any way

packages/libs/react-ui/src/components/Select/Select.test.tsx
webpro: Everything that happens in this test happens in the others. So why keep it?

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,62 @@
+import { SystemIcon } from '../Icons';
```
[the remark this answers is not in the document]
sstraatemans: @rezaerami  for a11y you would at least need a aria-label.
screenreaders have no idea at the moment.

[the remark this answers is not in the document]
sanderlooijenga: Don't we have an eslint rule for the a11y stuff?

PR #543, opened 2023-07-05
[@kadena/transfer] Module explorer

jessevanmuijden:


nil-amrutlal-dept:
Is this PR completed? I noticed the code viewer is displayed after the list (which can be really big sometimes)

jessevanmuijden:
@nil-amrutlal-dept not quite. I just learned that the code editor should render the contract in JavaScript, which should be updatable and executable. The UX is indeed not optimal with the editor below the list. I would prefer a detail view on a page that is separate from the list page.

jessevanmuijden:
@nil-amrutlal-dept can you please help me get this PR merged after all?

nil-amrutlal-dept:
Note: when we have a design for this page, the affected pages should be changed accordingly. Appart from that, everything looks good

nil-amrutlal-dept (approved):

nil-amrutlal-dept (approved):
Note: when we have a design for this page, the affected pages should be changed accordingly. Appart from that, everything looks good

PR #544, opened 2023-07-05
Housekeeping lint config

webpro:
Sorry, one thing led to another so this PR is larger than intended... Bottom line is that we missed quite a few issues (warnings/errors), caused by the way things are configured. Let's try to align things a bit more moving forward.

- Fix `import/no-unresolved` (both from Rush scripts and in VS Code) and increase severity from `warn` to `error`, and fix revealed issues
- Replace `@rushstack/eslint-config/profile/node` with `@kadena-dev/eslint-config/profile/lib`, this revealed lint issues, also fixed in this PR
- Remove `prettier` plugin from ESLint (we already have that in `format` commands/scripts)
- Remove `next lint` from Next.js packages and migrate to central lint config
- Consistent `lint` script in each `package.json`
- Remove last traces of `lint-staged`
- Cleanups & dedupes, etc.

From a review perspective, please make sure things aren't off in your packages and workflows.

### DIFF
--- a/packages/apps/graph-client/src/components/chainweb/chain-block/chain-block.tsx
+++ b/packages/apps/graph-client/src/components/chainweb/chain-block/chain-block.tsx
@@ -14,7 +14,7 @@ interface IChainBlockProps {
   textColor: string;
 }
 
-export function ChainBlock(props: IChainBlockProps): JSX.Element {
+export const ChainBlock = (props: IChainBlockProps): JSX.Element => {
   const { color, textColor, block } = props;
 
   return (
@@ -49,4 +49,4 @@ export function ChainBlock(props: IChainBlockProps): JSX.Element {
       ) : null}
     </Container>
   );
-}
+};
--- a/packages/tools/eslint-config/profile/lib.js
+++ b/packages/tools/eslint-config/profile/lib.js
@@ -1,22 +1,3 @@
-const fs = require('fs');
-const path = require('path');
-
-/**
- * according to rush.json#projectFolderMin/MaxDepth packages specific to this
- * monorepo are always placed in 3rd nested directory
- * e.g. packages/{apps,libs,tools}/package-name/*
- */
-const prettierOptions = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, '../../../../.prettierrc'), 'utf8'),
-);
-/**
- * an alternative to above code is
- */
-// const prettierrcPath = `${require('child_process')
-//   .execSync('git rev-parse --show-toplevel')
-//   .toString()
-//   .replace(/\n/, '')}/.prettierrc`;
-
 module.exports = {
   root: true,
   extends: [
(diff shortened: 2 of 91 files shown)


sstraatemans (approved):

alber70g:
I remember @rezaerami adding the prettier config because not all IDEs pick up the prettier config from the root. This way you get autoformat through eslint plugins.

packages/apps/graph-client/src/components/chainweb/chain-block/chain-block.tsx
```diff
@@ -14,7 +14,7 @@ interface IChainBlockProps {
   textColor: string;
 }
 
-export function ChainBlock(props: IChainBlockProps): JSX.Element {
+export const ChainBlock = (props: IChainBlockProps): JSX.Element => {
```
alber70g, line 17: Do we maybe prefer this `const Component: React.FC<IChainBlockProps> = (props) => {}` over components having types separately for props and return value?

If so, it'd be nice if there's a way to auto-fix this construct.

packages/tools/eslint-config/profile/lib.js
alber70g, line 1: Do we want to include [sort-package-json](https://github.com/kellyselden/eslint-plugin-json-files/blob/master/docs/rules/sort-package-json.md) in the future?

packages/apps/graph-client/src/components/chainweb/chain-block/chain-block.tsx
webpro: Nothing in this PR is necessarily a preference, things like this are all auto-fixed based on existing configuration.

packages/tools/eslint-config/profile/lib.js
```diff
@@ -29,29 +10,22 @@ module.exports = {
...
-  plugins: [
-    '@kadena-dev/eslint-plugin',
-    'import',
-    'simple-import-sort',
-    'prettier',
-  ],
+  plugins: ['@kadena-dev/eslint-plugin', 'import', 'simple-import-sort'],
   rules: {
     '@kadena-dev/no-eslint-disable': 'error',
-    'prettier/prettier': ['warn', prettierOptions],
```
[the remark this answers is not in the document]
webpro: I've restored prettier plugin/config

packages/tools/kda-cli/package.json
```diff
@@ -27,6 +27,7 @@
     "format:pkg": "sort-package-json",
     "format:src": "prettier config src --write",
     "kda": "node lib/index.js",
+    "lint": "eslint ./src --ext .js,.t,.tsx --fix",
```
sanderlooijenga, line 30: `.t` should be `.ts`?

webpro: > `.t` should be `.ts`?

Oh that's some serious 🦅 👀 here! Thanks

sanderlooijenga: Haha yes, that's what I'm known for 😎

sanderlooijenga (approved):

alber70g (approved):

rezaerami (approved):

PR #547, opened 2023-07-06
Fenced code blocks always `sh`

webpro:
This one is a bit "over the top" and totally over-engineered, but it's to show how we can leverage the Markdown pipeline we have. We can add more remark plugins to manipulate the Markdown AST (mdast) and improve and streamline it in many ways. See the `@kadena-dev/markdown` package in our monorepo for more details. Or AMA:)

### DIFF
--- a/common/changes/@kadena/client/chore-fenced-code-blocks-sh_2023-07-06-08-29.json
+++ b/common/changes/@kadena/client/chore-fenced-code-blocks-sh_2023-07-06-08-29.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/client",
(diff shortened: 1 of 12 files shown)


EnoF (approved):
Consistency aside, does using `sh` over `bash` have any significant impact on how code blocks are displayed?

webpro:
> Consistency aside, does using `sh` over `bash` have any significant impact on how code blocks are displayed?

[No](https://github.com/github-linguist/linguist/blob/master/lib/linguist/languages.yml#L6393-L6400), I just happened to see @sstraatemans fix them somewhere yesterday, and in my previous PR I noticed I'm doing it different so here I am opening this PR.

alber70g (approved):

PR #553, opened 2023-07-06
chore(client): use @kadena/client 0.5.0 until the new version is released

alber70g:
I have added the version to `common-versions.json` to allow discrepancies between versions of @kadena/client, specifically `0.5.0` instead of `workspace:*`


alber70g:
Closed as we can do this in the branch

PR #554, opened 2023-07-06
Tooltip component

bearni95:
- Closes ["Add tooltip component"](https://app.asana.com/0/1204808093245015/1204934728239264/f) Asana ticket
- Using an IconButton as the hovereable component in stories
- No animations
- Just intended to be right-aligned, for now
- Exports `Tooltip.Root`, which is the component, and `Tooltip.handler`, the method that toggles the tooltip's visibility
- It uses react's refs, as well as some js logic to handle the positioning and visibility; the stories should be pretty illustrative on how to work with it


packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -0,0 +1,46 @@
...
+    paddingX: '$4',
+    paddingY: '$3',
+    border: 'none',
+    fontSize: '$base',
+    backgroundColor: '$neutral1',
+    color: '$neutral6',
+    width: 'max-content',
+  }),
+  {
+    position: 'absolute',
```
sstraatemans, line 20: sprinkle

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -0,0 +1,46 @@
...
+    top: '50%',
+    marginRight: '-50%',
+    border: `${vars.borderWidths.$md} solid ${vars.colors.$neutral2}`,
+    display: 'none',
```
sstraatemans, line 24: sprinkle?

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -0,0 +1,46 @@
...
+    top: `calc(50% - ${vars.sizes.$4} / 2)`,
+    left: `calc((-1 * ${vars.sizes.$4} / 2) - 1px)`,
+    rotate: '45deg',
+    borderLeft: `${vars.borderWidths.$md} solid ${vars.colors.$neutral2}`,
+    borderBottom: `${vars.borderWidths.$md} solid ${vars.colors.$neutral2}`,
+  },
+]);
+
+export const visibleClass = style({
+  display: 'block',
```
sstraatemans, line 45: sprinkle

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
bearni95: For `position` and `display` I agree that they should be turned into sprinkles, not so sure that it's worth adding the 50% and -50% rules to `top` and `marginRight`, as they seem like less common values to be used in other designs

sstraatemans (approved):

PR #555, opened 2023-07-07
ESLint config external

webpro:
* Fix reusability of .prettierrc file finder in eslint config (3862e375)
* Fix dependencies for eslint packages (e8e3a5d5)

### DIFF
--- a/packages/tools/eslint-config/profile/lib.js
+++ b/packages/tools/eslint-config/profile/lib.js
@@ -2,7 +2,6 @@ const fs = require('fs');
 const path = require('path');
 
 const findUp = (filename, dir = process.cwd()) => {
-  console.log({ filename, dir });
   const filePath = path.join(dir, filename);
   if (fs.existsSync(filePath)) return filePath;
   const parentDir = path.dirname(dir);
--- a/packages/tools/eslint-config/profile/lib.js
+++ b/packages/tools/eslint-config/profile/lib.js
@@ -1,8 +1,17 @@
 const fs = require('fs');
 const path = require('path');
 
+const findUp = (filename, dir = process.cwd()) => {
+  console.log({ filename, dir });
+  const filePath = path.join(dir, filename);
+  if (fs.existsSync(filePath)) return filePath;
+  const parentDir = path.dirname(dir);
+  if (parentDir === dir) return;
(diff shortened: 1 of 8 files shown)


packages/tools/eslint-config/profile/lib.js
```diff
@@ -1,8 +1,17 @@
...
+  const filePath = path.join(dir, filename);
+  if (fs.existsSync(filePath)) return filePath;
+  const parentDir = path.dirname(dir);
+  if (parentDir === dir) return;
+  return findUp(filename, parentDir);
+};
+
 const prettierOptions = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, '../../../../.prettierrc'), 'utf8'),
+  fs.readFileSync(findUp('.prettierrc'), 'utf8'),
```
alber70g, line 13: How does this work when this package is a dependency of a package that has a `.prettierrc` and resides in the `node_modules`?
Maybe we should consider going up to the higest `node_modules` dir first.

Something like (pseudocode)
```
const nodeModulesPath = findUpHighest(`node_modules`)
const firstNonNodeModulesPrettierrcPath = findUp(from: nodeModulesPath, `.prettierrc`)
```

alber70g: This should cover it

`findUp('.prettierrc', process.cwd().split('node_modules')[0])`

webpro: This surely isn't flawless, but I simply tested the following scenarios:

- Run within this monorepo from a package dir, e.g. `rushx lint` or `rush lint --only ...`
- Run within this monorepo from root with `rush lint`
- Run from root in another repo

Guess that's alright for now?

alber70g: I'd say so. But it's a small change to make it work everywhere, no? 

```js
findUp('.prettierrc', __dirname.split('node_modules')[0])
```

alber70g (approved):

packages/tools/eslint-config/profile/lib.js
webpro: Discussed offline that the current solution is good/intentional.

PR #558, opened 2023-07-10
Input component leading text: option to manually provide the width

bearni95:
- Closes ["Bug: Input component leading text doesn't line up when grouped together"](https://app.asana.com/0/1204808093245015/1204956494923588/f)
- Adds the `leadingTextWidth` to the `Input` component
- Adds the `labelWidth` to the `InputWrapper` component, which passes down the value given to the `leadingTextWidth` prop of all `Input` children inside it
- In both cases, the value given can be `undefined` for the current behavior, or a value from `vars.sizes` to set a fixed width


timoheddes (approved):

sstraatemans (changes requested):
have placed a comment

as an overall remark.
I think I like the size thing, but do we want to have this responsive?
maybe with vw and clamp in css? so it gets a min and max?

bearni95:
> have placed a comment
> 
> as an overall remark. I think I like the size thing, but do we want to have this responsive? maybe with vw and clamp in css? so it gets a min and max?

am having a hard time figuring out the API to provide those values when calling the `<Input>` and `<InputWrapper>` components, as well as a way to provide them outside of their style props. I agree that the whole list of responsive sizes is quite too much for this feature, probably worth discussing on a call with @eileenmguo's input on it

packages/libs/react-ui/src/components/Input/Input.css.ts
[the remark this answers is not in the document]
eileenmguo: if both inputs use clamp will they change in size (within the range) together or is it possible that they would adjust their size based on the contents?

packages/libs/react-ui/src/components/InputWrapper/InputWrapper.tsx
```diff
@@ -13,13 +14,15 @@ export interface IInputWrapperProps extends Omit<IInputHeaderProps, 'label'> {
   disabled?: boolean;
   helperText?: string;
   label?: string;
+  labelWidth?: keyof typeof vars.sizes;
```
eileenmguo, line 17: the name is a bit misleading. I think this should be "leadingTextWidth" or something since the label for this component is referring to something else

eileenmguo (commented):
I think the size prop is ok for now. It would be nice if the size could be determined by the text. So like setting the width to `${leadingText.length + 2}ch`, but not sure how well that would work in practice

packages/libs/react-ui/src/components/Input/Input.css.ts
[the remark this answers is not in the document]
bearni95: Been testing with clamp, as long as the wrapper sends down the same values the inputs behave together!

packages/libs/react-ui/src/components/Input/Input.css.ts
```diff
@@ -68,8 +73,17 @@ export const leadingTextClass = style([
...
     paddingX: '$4',
   }),
+  {
+    minWidth: 'max-content',
+  },
 ]);
 
+export const leadingTextWidthVariant = styleVariants(vars.sizes, (size) => {
+  return {
+    width: size,
```
sstraatemans, line 83: what happens when the text is to big for the width?
should it break the word? use ellipses?

packages/libs/react-ui/src/components/Input/Input.stories.tsx
```diff
@@ -40,6 +41,15 @@ const meta: Meta<
         type: 'text',
       },
     },
+    leadingTextWidth: {
```
sstraatemans, line 44: In the story of Inputgroup I also see leadingTextwidth as a prop. But does not do anything.
image

packages/libs/react-ui/src/components/Input/Input.css.ts
sstraatemans: example of a large label on smallscreen
image


example of 1 large label with multpile fields
image

sstraatemans (changes requested):
especially when there are large labels, things seem to break

packages/libs/react-ui/src/components/Input/Input.css.ts
eileenmguo: i think it should use ellipses

packages/libs/react-ui/src/components/Input/Input.stories.tsx
bearni95: Should be fixed with the new ellipsis strategy

sstraatemans (approved):

PR #566, opened 2023-07-11
Create @kadena-dev/lint-package

webpro:
Adds a custom packager linter (`@kadena-dev/lint-package`). Runs with `rush lint` after `eslint`.

Current rules:

- Check `extends` is correct in `.eslintrc.c?js`
- Check `extends` is correct in `tsconfig.json`
- Check `heft` is used in `"build"` and `"test"` scripts in `package.json` (except for apps and react packages)
- Check `"lint"` or `"lint:src"` script includes the `--fix` argument
- If `config/api-extractor.json` exists, the `"types"` should point to `dist/pkg-name.d.ts`

Using [`@kadena-dev/create-package` to create new packages](https://github.com/kadena-community/kadena.js/pull/556) will also help, but that's only a start. Later on we can add more rules and extend it to auto-fix things.

**EDIT** This PR will be unblocked by https://github.com/kadena-community/kadena.js/pull/612

Made it look a bit like the ESLint output. Example:

Screenshot 2023-07-13 at 7 25 10 AM

By the way, the ESLint stuff is what takes up >90% time, time added in this example by this package is 0.3s per project.

---
  - https://app.asana.com/0/0/1204936094493567


packages/tools/lint-package/.eslintrc.cjs
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  extends: ['@kadena-dev/eslint-config/profile/lib'],
+  parserOptions: { tsconfigRootDir: __dirname },
+  rules: {
+    '@rushstack/typedef-var': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/naming-convention': 'off',
+    '@typescript-eslint/strict-boolean-expressions': 'off',
+  },
```
alber70g, line 12: Why are they disabled?

packages/tools/lint-package/lib/index.js
alber70g, line 1: Is this one still needed when using an autoinstaller? 
This one's now added to git, and because of that the build-cache cannot be used. The build-cache uses the tracked files to determine the hash for the cached files (the files named in `package.json#files`

packages/tools/lint-package/src/helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const loadJSON = async <T>(filePath: string): Promise<T> =>
+  JSON.parse(String(await fs.readFile(filePath)));
+
+export const loadJSONC = async <T>(filePath: string): Promise<T> =>
+  JSONC.parse(String(await fs.readFile(filePath)));
+
+export const loadContents = async (filePath: string) =>
+  String(await fs.readFile(filePath));
```
alber70g, line 29: We can remove the need for `String()` by adding the encoding

```suggestion
export const loadJSON = async <T>(filePath: string): Promise<T> =>
  JSON.parse(await fs.readFile(filePath, 'utf-8'));

export const loadJSONC = async <T>(filePath: string): Promise<T> =>
  JSONC.parse(await fs.readFile(filePath, 'utf-8'));

export const loadContents = async (filePath: string): Promise<string> =>
 await fs.readFile(filePath, 'utf-8');
```

packages/tools/lint-package/src/index.ts
alber70g, line 1: Warning: nitpick

We could wrap this in a function? I'm not a fan of implicitly executing a bunch of code.

Like: 

```ts
async function main() {...}
main().catch(console.error)
```

Also, we could move all promises to a `Promise.all([])` that way they'll be done in parallel

packages/tools/lint-package/src/rules/api-extractor/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    const name = pkg.name?.replace(/.+\//, '');
+    const types = `dist/${name}.d.ts`;
+    if (!pkg.types || pkg.types !== types) {
+      issues.push(
+        `API Extractor is enabled, "types" in package.json should be "${types}"`,
+      );
+    }
+  } catch (error) {
+    // console.warn('Missing config/api-extractor.json');
```
alber70g, line 21: Why do we swallow all errors here?

rush.json
```diff
@@ -606,6 +606,14 @@
       "shouldPublish": false
     },
 
+    {
+      "packageName": "@kadena-dev/lint-package",
+      "tags": ["tools"],
+      "projectFolder": "packages/tools/lint-package",
+      "shouldPublish": false,
+      "versionPolicyName": "alpha"
```
alber70g, line 614: We must not have a `versionPolicyName` when `shouldPublish=false`.

(This should be part of the linter maybe?)

packages/tools/lint-package/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+const pkgRules = await loadRules('package-json');
+const tsRules = await loadRules('typescript');
+const eslintRules = await loadRules('eslint');
+const apiExtractorRules = await loadRules('api-extractor');
+
+const pkg = await loadJSON<PackageJson>(PACKAGE_JSON);
+const tsConfig = await loadJSONC<TSConfig>(TS_CONFIG);
+const [eslintConfigPath] = await fg('.eslintrc*', { cwd: PACKAGE_DIR });
+const eslintConfig = await loadContents(eslintConfigPath);
+
```
alber70g, line 22: Should we add the required files to the loader of each linter?

So each lint rule exports a function/object with the files it needs? That way this file doesn't need to know about the files the rules need, but it can still remove duplicates.

```ts

type RequiredFiles = Array<{ path: string }>;

type LintRule = {
  files: RequiredFiles;
  rules: Array<{
    ruleName: string;
    rule: (files: RequiredFiles[number] & { getContents: () => string }) => Issue[];
  }>;
};
```

alber70g (commented):
I tried to run it in the repo with `rush lint-package` but I got an error:

```log
node:internal/process/esm_loader:97
    internalBinding('errors').triggerUncaughtException(
                              ^

[Error: ENOENT: no such file or directory, open '/Users/albert/projects/kadena-community/kadena.js-pull-requests/package.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/albert/projects/kadena-community/kadena.js-pull-requests/package.json'
}
```

packages/tools/lint-package/lib/index.js
webpro: Unfortunately there was no other way I got all commands to work (and running `rush lint-package` directly is not even working yet). Somehow the `rush install` sequence expects this file here so it can link it. Other packages do basically the same, yet they don't have this issue. Will try some more.

packages/tools/lint-package/src/helpers.ts
webpro: Ah of course, thanks

packages/tools/lint-package/src/index.ts
webpro: I think we should just exit with a non-zero code for exceptions. Otherwise they might get lost (local or CI). But I understand where it comes from: Rush's magic, which we shouldn't rely on.

rush.json
webpro: Alright, thanks

packages/tools/lint-package/src/index.ts
webpro: In my implementation, simply all file contents are sent to each rule (in `options`). This is a bit coarse, but sometimes we need more than just the relevant file (e.g. when having `config/api-extractor.json` we need `package.json` too). The advantage is that (future) rules can read (and fix) all of them to do their thing.

Having that said, this code is not very pretty. Improving the rules loading mechanism could be a nice follow-up.

packages/tools/lint-package/src/rules/api-extractor/types.ts
alber70g: Maybe add a comment indicating that

packages/tools/lint-package/src/index.ts
alber70g: Not sure why this is about Rush 😕
It's a best practice to have no top-level await, and have at least a `try-catch` block around the promises.
Also `console.error` is what you might referring to when talking about "Rush's magic", but the literal `console.error` was merely an example. This could be easily implemented by passing an error handling function that ends with `process.exit(1)`

packages/tools/lint-package/src/rules/api-extractor/types.ts
webpro: Alright, added a comment

packages/tools/lint-package/src/index.ts
webpro: I mentioned Rush, because I saw only `console.error` in the example and no exit code, and [Rush acts on `stderr` and not just exit codes](https://github.com/kadena-community/kadena.js/pull/442#issuecomment-1604214895).

Totally agree with the sentiment of wrapping the main sequence, so I did. Although currently in its initial phase it basically still does the same thing. We can still add friendlier error messages later on.

packages/tools/lint-package/.eslintrc.cjs
webpro: They didn't spark joy.

Yet, thanks to your comment and after careful consideration we are able to remove two of them:

- Still disabled. See the screenshot below. I just doesn't make sense to type simple things that TS infers, especially in this case. Sometimes it has value for sure, such as with library code and/or when things need to be narrowed down, etc. Not here, though.
- Still disabled. Basically similar to the previous point.
- Removed. Not sure what happened, but when I removed this ESLint is complaining.
- Removed. I don't like to write `pkg.types === undefined || pkg.types === ''` when I can do `!pkg.types`. It's a good part of JavaScript, actually. But it's one occasion so I might be able to get over it.


Screenshot 2023-07-14 at 2 55 28 PM

packages/tools/lint-package/lib/index.js
alber70g: Yes `rush install` links the binaries to where they're used. However the `lib/index.js` isn't there yet. What I did for the `@kadena/pactjs-cli` is add a script to initialize with an empty `index.js`. https://github.com/kadena-community/kadena.js/blob/main/rush.json#L315

webpro: Just yesterday I had the same issue in another branch, fixed it in similar fashion (`touch file` from `preinstall` script), but closer to the source of the issue: https://github.com/kadena-community/kadena.js/pull/612/commits/8d7ff0927afc371ccbd801e1370978067e206d4a

If necessary, later on I can apply that solution here too.

PR #572, opened 2023-07-12
fix(kadena-docs): improve slug method for special characters + strings ends with special characters

realdreamer:
- Updated a createSlug method to pass `index`
- Append `index` values only if string has any special characters at the end. To avoid a use case like `not` and `not?`  to have same link reference
- if string contains only special characters then it replaced with `operator-${index}`
- Update components to pass `index` props

### DIFF
--- a/packages/apps/docs/src/components/Layout/components/Aside/ListItem.tsx
+++ b/packages/apps/docs/src/components/Layout/components/Aside/ListItem.tsx
@@ -12,6 +12,7 @@ interface IProps {
   item: ISubHeaderElement;
   setActiveItem: React.Dispatch<React.SetStateAction<string>>;
   activeItem?: string;
+  index?: number;
 }
 
 export const ListItem: FC<IProps> = ({
@@ -19,11 +20,12 @@ export const ListItem: FC<IProps> = ({
   setActiveItem,
   activeItem,
   scrollArea,
+  index,
 }) => {
   const router = useRouter();
 
   if (item.title === undefined || item.title === '') return null;
-  const slug = `#${createSlug(item.title)}`;
+  const slug = `#${createSlug(item.title, index)}`;
 
   const handleItemClick = (ev: MouseEvent<HTMLAnchorElement>): void => {
     ev.preventDefault();
@@ -53,7 +55,7 @@ export const ListItem: FC<IProps> = ({
     >
       {item.children.length > 0 && (
         <AsideList inner={true}>
-          {item.children.map((innerItem) => {
+          {item.children.map((innerItem, index) => {
             return (
               <ListItem
                 key={innerItem.title}
@@ -61,6 +63,7 @@ export const ListItem: FC<IProps> = ({
                 item={innerItem}
                 activeItem={activeItem}
                 setActiveItem={setActiveItem}
+                index={index + 1}
               />
             );
           })}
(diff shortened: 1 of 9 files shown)


packages/apps/docs/src/utils/createSlug.ts
```diff
@@ -1,10 +1,21 @@
...
-  return str
+  const normalizedSlug = str
     .normalize('NFD')
     .replace(/[\u0300-\u036f]/g, '')
     .replace(/[^\w\s]+/g, '')
     .replace(/ /g, '-')
     .toLowerCase()
     .replace(/^-+|-+$/g, '');
+
+  if (normalizedSlug === '' && index !== undefined) return `operator-${index}`;
```
sstraatemans, line 11: could we make `operator` more generic? now its just there for operators.
maybe just `header-idx`

packages/apps/docs/src/components/Layout/components/Aside/ListItem.tsx
```diff
@@ -12,18 +12,20 @@ interface IProps {
   item: ISubHeaderElement;
   setActiveItem: React.Dispatch<React.SetStateAction<string>>;
   activeItem?: string;
+  index?: number;
```
sstraatemans, line 15: make the name a bit more descriptive.
like `headerIdx`

packages/apps/docs/src/components/Markdown/Heading/Heading3.tsx
```diff
@@ -2,6 +2,10 @@ import { IHeader, TaggedHeading } from './Heading';
 
 import React, { FC } from 'react';
 
-export const Heading3: FC<IHeader> = ({ children }) => {
-  return <TaggedHeading as="h3">{children}</TaggedHeading>;
+export const Heading3: FC<IHeader> = ({ children, ...rest }) => {
+  return (
+    <TaggedHeading as="h3" {...rest}>
```
sstraatemans, line 7: this rest is not needed. that was just for test, right?

realdreamer: no! it's required then only `index` get passed to down the component. Else I could use only `index` explicitly

packages/apps/docs/src/scripts/remarkHeadersToProps.mjs
```diff
@@ -38,9 +38,15 @@ const remarkHeadersToProps = () => {
       },
     ];
 
-    headers.forEach((item) => {
+    headers.forEach((item, index) => {
       const parent = lastHeading(startArray[0], item);
 
+      item.data = {
```
sstraatemans, line 44: can you add a comment, of what and why we are doing this?

packages/apps/docs/src/utils/createSlug.ts
realdreamer: how about `aside-menu-idx` or `menu-idx` ? because header it's more of the component where it's used but we need this value for link reference

packages/apps/docs/src/components/Layout/components/Aside/ListItem.tsx
```diff
@@ -53,14 +55,15 @@ export const ListItem: FC<IProps> = ({
     >
       {item.children.length > 0 && (
         <AsideList inner={true}>
-          {item.children.map((innerItem) => {
+          {item.children.map((innerItem, index) => {
```
sstraatemans, line 58: hm...can we maybe do this calculation in the remark script?
it makes the react code a bit harder to read.

sstraatemans (changes requested):
got some comments

sstraatemans (approved):

PR #573, opened 2023-07-12
[@kadena/react-ui] Added colors to masked value component

nil-amrutlal-dept:


realdreamer (approved):

sanderlooijenga (approved):

sanderlooijenga:
Can we see the different colors in the storybook entry?

nil-amrutlal-dept:
> different colors in the storybook entry?

If you switch between light and dark mode we can see the colors changing

PR #574, opened 2023-07-12
chore(docs): replace Typography and switch themes fix

sstraatemans:
replace the headings and text components from react-components to react-ui

i was replacing the react-components typography  to react-ui components
I found a couple of issues in Heading component. so fixed that

And I found a problem in our switch between themes, which forced me to create our own switch instead of using the package (most code is copied).
The problem was that we are using 2 component libraries. and the next-themes package can only switch themes for 1. not for 2.


packages/apps/docs/src/pages/_app.tsx
```diff
@@ -47,33 +47,25 @@ export const MyApp = ({
 }): JSX.Element => {
   const props = deserializePageProps(pageProps);
 
+  console.log(darkThemeClass, stitchesDarkTheme.className);
```
realdreamer, line 50: this could be removed?

packages/apps/docs/src/hooks/useTheme/useTheme.tsx
```diff
@@ -0,0 +1,321 @@
...
+        value ? `var x=${JSON.stringify(value)};` : ''
+      }${updateDOM(value ? `x[e]` : 'e', true)}}else{${updateDOM(
+        defaultTheme,
+        false,
+        false,
+      )};}${fallbackColorScheme}}catch(t){}}();`;
+    })();
+
+    return (
+      <script nonce={nonce} dangerouslySetInnerHTML={{ __html: scriptSrc }} />
```
realdreamer, line 149: adding comments give some idea about, why are we executing the script from `react`?

sstraatemans: about executing the script. it is just a copy paste of the usetheme.. so actually I didnt jump into that one

it looks like that that part of the script needs to be executed directly

realdreamer (approved):
🔥

PR #576, opened 2023-07-13
Feat/refactor button

timoheddes:
**Storybook changes:**
- adds `controls` addon to enable conditional enabling/disabling of controls

**Button changes**
- rendering as anchor tag wasn't working correctly, now fixed
- made icon a prop of Button so we can simply export and import `Button` (implementation to be discussed with Eileen and Steven)
- added iconAlign to also allow controlling of icon position through props
- removed the arbitrary stories that load in slightly different variants of the button which can also be achieved using the controls
- fine-tuned and expanded on the existing controls with conditionals, descriptions etc.


packages/libs/react-ui/src/components/Button/Button.stories.tsx
```diff
@@ -13,14 +14,32 @@ const meta: Meta<
...
+    onClick: { action: 'clicked', if: { arg: 'as', eq: 'button' } },
+    icon: {
+      options: [
+        ...['-'],
+        ...Object.keys(SystemIcon),
+      ] as (keyof typeof SystemIcon)[],
       control: {
         type: 'select',
       },
+      if: { arg: 'loading', eq: false },
```
timoheddes, line 34: https://storybook.js.org/docs/react/essentials/controls/#conditional-controls

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,43 +1,72 @@
...
-  href?: string;
+  color?: keyof typeof colorVariants;
   children: React.ReactNode;
-  title?: string;
   disabled?: boolean;
-  color?: keyof typeof colorVariants;
+  href?: string;
+  icon?: keyof typeof SystemIcon;
+  iconAlign?: 'left' | 'right';
+  loading?: boolean;
```
timoheddes, line 17: todo: animate/spin loading icon with some css :)

sanderlooijenga: I've implemented it using this example; https://codepen.io/eveness/pen/BjLaoa

timoheddes: Ohh I thought of just doing a continuous rotate transform of 0 to 360deg would do it as the image we're using is just a quarter slice of a circle outline ... if that makes sense haha

timoheddes:
Closed this in favour of https://github.com/kadena-community/kadena.js/pull/655

Seemed easier to just create a new PR rather than clean up this one.

PR #582, opened 2023-07-13
Add tooltip option to display on all four sides of the element

bearni95:
- Closes ["Add tooltip option to display on all four sides of the element"](https://app.asana.com/0/1204958723576385/1205043261263000/f)
- Added support for indicating top/bottom/left/right placement for the tooltip, as well as the relevant styles for each


sstraatemans (approved):

realdreamer (commented):
changes LGTM 👍🏽  but when there is a scrollbar in the content area if you scroll the page hover the icon then tooltip is not positioned properly.  

image

bearni95:
> changes LGTM 👍🏽 but when there is a scrollbar in the content area if you scroll the page hover the icon then tooltip is not positioned properly.
> image

Managed to fix it adding window.scrollY and window.scrollX on the calculation for the tooltip's positioning

PR #586, opened 2023-07-14
fix(client): updated examples rename createPactCommand to composePact…

javadkh2:
Renamed createPactCommand to composePactCommand
changed ISignFunction add overloads to  accept only one command and returns a single signed command instead 
updated the client-examples


packages/libs/client-examples/src/example-contract/crosschian-transfer.ts
webpro, line 1: Typo in filename: `chian`

packages/libs/client-examples/src/example-contract/describe-account.ts
```diff
@@ -42,9 +25,10 @@ async function getBalanceMain() {
     .setNetworkId('mainnet04')
     .createTransaction();
 
-  const res = await local(tr);
+  const res = await dirtyReady(tr);
 
   console.log(res);
 }
```
alber70g, line 31: Shouldn't this be `dirtyRead`? Similar to https://api.chainweb.com/openapi/pact.html#:~:text=In%20a%20blockchain%20environment%20this%20would%20be%20a%20node%2Dlocal%20%E2%80%9Cdirty%20read%E2%80%9D

javadkh2: its a typo

alber70g (approved):

PR #589, opened 2023-07-15
Update README.md

trendzetter:
fix the link to the functional programming style example


webpro:
Thanks @trendzetter! To fix the links, please (also) fix them in the definitions (the url's) at the bottom. Any chance you could do that? Otherwise I'm happy to pick this up myself.

trendzetter:
@webpro Should be better now. I didn't know about that part of the markdown syntax

webpro:
We also need to update the changelog, but I opened #653 to get done with it, hope you don't mind :) Thanks again, @trendzetter!

PR #594, opened 2023-07-17
fix(pactjs-generator): parse symbols as string

javadkh2:
fixed str parser to parse pact symbols (`'symbol_name`) as string. 
this will fix the issue with finding module namespace


jermaine150 (approved):

webpro (approved):
Seems good enough for now, but I can imagine we'll want more test scenarios with various types of (nested) quoting?

PR #595, opened 2023-07-17
Feat/accordion

rezaerami:
- add tree component to render a list of menu and sub menus
- add docs for tree component
- add test for tree component
- add accordion component
- add docs rfor accordion component
- add test for accordion component

image

image


sstraatemans (changes requested):
some overall comments:
- can you add to the tree functionality that only 1 can be opened
- when clicking on an option (both tree and accordion) can you add a call back function for that. docs needs to calculate the height for the side menu so the menu can hover.

I still need to see how I can use the accordion and the tree at the same time.
Because I the docs it is the same component.
1 accordion is the first branch of the tree.

isaozler (approved):

MRVDH:
> * can you add to the tree functionality that only 1 can be opened

@sstraatemans Do we really want this? I'm not a fan of websites that obfuscate navigation and having to click every time to open/close something. You lose overview and it requires clicks to get the overview back, which is not something you want for navigation.

Some examples of successful apps that allow multiple open items in the side nav bar:

- Figma
- VS Code
- Slack
- Github
- Asana

And those are just the tabs/apps I have open 😅

rezaerami:
@sstraatemans I will implement the linked feature for the tree, but it will add performance complexity, since now each tree is only responsible for itself, but adding a linked feature, it'll be responsible for its children instead
regarding open and close callbacks it's fine, will add them

sstraatemans:
@MRVDH well according to design we want this. we can discuss with @isaozler 
Same goes for the remark from @rezaerami 
If this is not wanted by @isaozler I'm fine with removing this.
Maybe we should have a small call later today to discuss the issues, before building more

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  return (
+    <div className={accordionWrapperClass} data-testid="kda-accordion-wrapper">
+      {sections.map((section, index) => (
+        <AccordionSection
+          isOpen={isOpen(index)}
+          onToggle={() => handleToggle(index)}
+          title={section.title}
+          key={index}
```
eileenmguo, line 36: we should avoid using indices as keys. Is there a value in section that can be used for this?

packages/libs/react-ui/src/components/TreeList/TreeList.css.ts
```diff
@@ -0,0 +1,63 @@
...
+export const treeBranchWrapperClass = style([
+  {
+    transition: 'all 0.2s ease',
+    overflow: 'hidden',
+  },
+]);
+export const treeBranchWrapperVariant = styleVariants({
+  isParent: [
+    {
+      marginLeft: '0',
```
eileenmguo, line 55: can use a sprinkle here

eileenmguo (commented):
Regarding the conversation about having one section open at a time: For the sake of reducing complexity and also improving UX (in my opinion), I don't think we should only allow one section to be open. From a UX perspective, the restriction of one open section would require one of the following if a section is already open:
- The user has to close the already open section + open the section they want to view resulting in 2 clicks rather than 1
- Expanding a section would automatically close a different section that is like a side effect of the intended action which I usually try to avoid if possible

If there are differing opinions/lack of clear consensus, I would say taking the less complex approach is preferable at this point in time

packages/libs/react-ui/src/components/TreeList/TreeList.css.ts
rezaerami: nope, it will throw error saying "0" is not recognized as a value

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
rezaerami: well, each element only has a title which is a react node, and nodes can't be used for keys, I can assign id, but then developer will be forced to add id to items of array which has no usage but the key!

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,43 @@
+import { accordionWrapperClass } from './Accordion.css';
+import { AccordionSection, IAccordionSection } from './AccordionSection';
+
+import React, { FC, useState } from 'react';
+
+export interface IAccordionProps {
+  className?: string;
+  sections: Omit<IAccordionSection, 'isOpen' | 'onToggle'>[];
+  linked?: boolean;
```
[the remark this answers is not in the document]
rezaerami: it'll link siblings together, so on open of one, the rest will get closed

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,43 @@
+import { accordionWrapperClass } from './Accordion.css';
+import { AccordionSection, IAccordionSection } from './AccordionSection';
+
+import React, { FC, useState } from 'react';
+
+export interface IAccordionProps {
+  className?: string;
```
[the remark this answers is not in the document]
eileenmguo: Right now I don't think it is getting applied to any of the JSX though?

In general we are trying to keep it more strict. This will result in more difficulty when trying to do something specific with components, but the hope is that it will result in more conformity. In what instances have you needed to change button styles or text fields?

packages/libs/react-ui/src/components/TreeList/TreeList.css.ts
eileenmguo: I just tried it out and it is a bit strange - it works but only when you pass the 0 as a number instead of a string
```suggestion
      marginLeft: 0,
```

isaozler:
We have to distinguish two different appliances on the left sidebar.

The first one; is the sidebar as a toolbar, and the second is navigation.

The use case @MRVDH is implementing is mostly (the tools dApp) going to be used as a toolbar. So the user has to have multiple branches open, to be able to use the application efficiently.

The use case of @sstraatemans (docs website) is different because we use the left sidebar as a navigation element where we have a hierarchal flow in most cases. We want to guide the user through the sections; in this case, therefore, we don't want the user to get distracted by other content (noise).

So eventually, both appliances are valid UX-wise. The documentation sidebar should be able to show only one branch at a time; as for the k:dApps, we should avoid this complexity.

eileenmguo (approved):

PR #605, opened 2023-07-18
[@kadena/transfer] Fix typing errors from UI package update

MRVDH:


packages/apps/transfer/src/pages/transfer/account-transactions/index.tsx
```diff
@@ -142,7 +142,7 @@ const CheckTransactions: FC = () => {
                 inputProps={{
                   id: 'account-input',
                   placeholder: t('Account'),
-                  onChange: (e) =>
+                  onChange: (e: any) =>
```
webpro, line 145: Is there really not some `...Event` type from React?

timoheddes: Something like `React.ChangeEvent<HTMLInputElement>` right?

MRVDH:
Fixed in another PR

PR #611, opened 2023-07-18
Making aria-label required for select component

bearni95:
- Closes ["Making aria-label required for select component"](https://app.asana.com/0/1204958723576385/1205070479996794/f)

### DIFF
--- a/packages/libs/react-ui/src/components/Select/Select.tsx
+++ b/packages/libs/react-ui/src/components/Select/Select.tsx
@@ -19,6 +19,7 @@ export interface ISelectProps
   value: string[] | string | number;
   onChange: React.ChangeEventHandler<HTMLSelectElement>;
   ref?: React.ForwardedRef<HTMLSelectElement>;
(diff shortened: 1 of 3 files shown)


eileenmguo (dismissed):
Once merged can you communicate this to the docs and tools teams that this is a breaking change that will require updates on their side?

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -19,6 +19,7 @@ export interface ISelectProps
   value: string[] | string | number;
   onChange: React.ChangeEventHandler<HTMLSelectElement>;
   ref?: React.ForwardedRef<HTMLSelectElement>;
+  ariaLabel: string;
```
eileenmguo, line 22: can you omit "aria-label" from the HTMLSelectElement props since we are passing it in via this prop?

eileenmguo (approved):

PR #612, opened 2023-07-18
Migrate to pnpm + turborepo

webpro:
This monorepo is managed using Rush. The tool is opinionated and may work well
for certain projects and teams, yet it's not performing to our standards and
needs (which is also opinionated, lol). The main issue is that it's basically a
framework in front of everything. A framework works well, until it doesn't.

Let's follow widely used standards and the Unix philosophy a bit more: pick the
right tool for each job separately. So we can update and swap out tools
separately as the ~~trend~~ need arises.

Rush is also an extra unnecessary blocker/hurdle for using the latest versions of Node.js and pnpm.

## What jobs?

- Build and test packages
- Format and lint source code
- Cache results of builds and tests to save time and resources
- Build and test packages, and only those affected in a pull request
- Publish packages
- Dependency management, e.g. to fixate versions of dependencies across packages

## But how?

- `pnpm` does a lot out of the box already (with this PR we go from v6 to v8 btw!)
- [turborepo][1] seems great for caching and "affected packages" logic
- [changesets][2] seems great for changelogs and publishing

## Guide

- Use more standards where possible, e.g. leverage `"private": true` for packages that should not be published.
- Use exit code to define results in CI (not `stderr` + warnings allowed). Also see #442.
- Commands are run through the package manager (npm, pnpm, ...) based on the standardized `package.json#scripts`. When we replace a tool the commands don't change. You _can_ run `turbo build` instead of `pnpm build`, but that's on you.
- The `rush build` job had linting included, such commands/actions should be separated (unless significant performance gain or something).

## TODO

- [x] Fix `pactjs` bin workarounds/issues e.g. in `@kadena/client-examples`
- [x] Enforce fixed version across monorepo for certain/all dependencies
- [x] Fix build of `@kadena/client-examples`
- [x] Implement alternative to [packageTable.ts][3]
- [x] Update docs
- [x] Revisit ESLint rules (warning vs error) (#442)

## TODO in follow-up PRs

- Find out wat heft/rigs actually do for us
- Look into version policies (e.g. to "group" packages)
- Look into publishing with changesets
- Propose flat `packages/*` structure

## Misc.

```sh
pnpm run test --filter @kadena/react-ui
pnpm run build --filter @kadena/client-examples
pnpm -r run "/^format:(md|pkg|src)/"
```

```sh
pnpm recursive list --json | jq --arg cwd "$PWD" '[ .[] | { name, version, private, path: (.path | sub("^" + $cwd + "/"; "")) } ]' > workspace.json
```

[1]: https://turbo.build
[2]: https://github.com/changesets/changesets
[3]: packages/tools/remark-plugins/src/commentMarkers/packageTable.ts

---
  - https://app.asana.com/0/0/1205075729086347


.syncpackrc.js
```diff
@@ -0,0 +1,38 @@
...
+      dependencyTypes: ['prod'],
+      pinVersion: 'workspace:*',
+    },
+    {
+      label:
+        'Types and internal dev packages are banned from dependencies (only allowed in devDependencies)',
+      packages: ['**'],
+      dependencies: ['@types/*', '*/types', '@kadena-dev/*'],
+      dependencyTypes: ['prod'],
+      isBanned: true,
```
webpro, line 35: [syncpack](https://github.com/JamieMason/syncpack) requires a bit of configuration, but is actually pretty powerful and has some advantages over the Rush solution:

- It also considers internal packages
- We can properly separate production and dev packages
- We can ban all sorts of packages containing types from `dependencies`

This PR contains quite some fixes detected by syncpack.

CONTRIBUTING.md
```diff
@@ -97,14 +79,6 @@ You are expected to install your own workflow the way you like it. For example,
...
 push their changes. That's why this repository does not auto-install Git hooks.
 
-If you want to make sure you don't forget to update the changelog before you
-push code, here's an example to install a Git hook for that:
-
-```bash
-echo "rush change --verify" > .git/hooks/pre-push
-chmod +x .git/hooks/pre-push
-```
-
```
webpro, line 107: changesets will take care of this, it's awesome. If you forget to run `pnpm changesets` you will be notified through a comment in the PR and an email.

CONTRIBUTING.md
```diff
@@ -113,18 +87,7 @@ To publish a new version of updated packages, please make sure you:
...
-- Bump the version
-- Publish updated packages
-
-```bash
-rush build
-rush test
-rush version --bump -b main
-rush publish --apply --publish --include-all --add-commit-details --set-access-level public --target-branch main
-```
+TODO
```
webpro, line 90: Let's pick this up later on (w/ changesets). Noticed much of publishing we do is pretty manual so not much harm done by removing this process atm I guess.

README.md
```diff
@@ -26,77 +26,122 @@ Overview of the main packages maintained in this repository:
...
+| [@kadena/docs][38]                   | [![version][40]][39] |
+| [@kadena/graph][41]                  | [![version][43]][42] |
+| [@kadena/kda-cli][44]                | [![version][46]][45] |
+| [@kadena/pactjs][47]                 | [![version][49]][48] |
+| [@kadena/pactjs-cli][50]             | [![version][52]][51] |
+| [@kadena/pactjs-generator][53]       | [![version][55]][54] |
+| [@kadena/react-components][56]       | [![version][58]][57] |
+| [@kadena/react-ui][59]               | [![version][61]][60] |
+| [@kadena/types][62]                  | [![version][64]][63] |
+| [kadena.js][65]                      | [![version][67]][66] |
```
webpro, line 49: This list now contains packages that do not have `private: true`

packages/libs/bootstrap-lib/package.json
```diff
@@ -42,18 +40,18 @@
...
   },
   "devDependencies": {
     "@kadena-dev/eslint-config": "workspace:*",
     "@kadena-dev/heft-rig": "workspace:*",
     "@kadena-dev/markdown": "workspace:*",
+    "@kadena/types": "workspace:*",
     "@rushstack/eslint-config": "~3.3.0",
     "@rushstack/heft": "~0.50.6",
     "@types/heft-jest": "~1.0.3",
+    "@types/isomorphic-fetch": "~0.0.36",
```
webpro, line 54: Changes in dependencies of this `packages.json` are examples of what synpack detected.

turbo.json
```diff
@@ -0,0 +1,24 @@
...
+      "outputMode": "new-only"
+    },
+    "format": {
+      "outputMode": "new-only"
+    },
+    "format:ci": {
+      "outputMode": "new-only"
+    }
+  }
+}
```
webpro, line 24: Yes, this is all it takes for us to turborepo 🚀

EnoF (approved):

javadkh2 (approved):
Awesome!

Takadenoshi (approved):
Happy to try out this approach

ggobugi27 (approved):

jermaine150 (approved):

.syncpackrc.js
```diff
@@ -0,0 +1,38 @@
+module.exports = {
```
glennreyes, line 1: Let's include types here like in this snippet:

https://jamiemason.github.io/syncpack/config-file#typescript-intellisense

(Applies to all config files that have types provided)

workspace.json
```diff
@@ -0,0 +1,170 @@
...
+    "private": true,
+    "path": "packages/tools/remark-plugins"
+  },
+  {
+    "name": "@kadena-dev/rush-fix-versions",
+    "version": "0.0.6",
+    "private": false,
+    "path": "packages/tools/rush-fix-versions"
+  }
+]
```
glennreyes, line 170: What is this file for?

.syncpackrc.js
webpro: Ah yes, thanks, done.

workspace.json
webpro: It's a generated replacement for `rush.json#projects`, can be used to generate other things. Currently it's only used to generate the tabel in the root README.md.

Generated using some crazy script (chatgpt ftw!) in `pnpm _list`. We should make things more clear here, wasn't sure yet how.

glennreyes (approved):

sstraatemans (approved):

eileenmguo (approved):

workspace.json
```diff
@@ -0,0 +1,176 @@
...
+    "version": "0.0.1",
+    "private": false,
+    "path": "packages/libs/react-components"
+  },
+  {
+    "name": "@kadena/react-ui",
+    "version": "0.0.1",
+    "private": false,
+    "path": "packages/libs/react-ui"
+  },
```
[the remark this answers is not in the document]
webpro: This is copied over from `package.json`, so it should be fixed there then :)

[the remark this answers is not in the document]
eileenmguo: I've made a PR for this: https://github.com/kadena-community/kadena.js/pull/691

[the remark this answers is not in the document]
webpro: Thanks! Updated this PR with the changes

alber70g:
> ## What jobs?
> 
> - Build and test packages
> - Format and lint source code
> - Cache results of builds and tests to save time and resources
> - Build and test packages, and only those affected in a pull request
> - Publish packages
> - Dependency management, e.g. to fixate versions of dependencies across packages
 
I'd like to extend the features of what we expect from the monorepo tool(s) in a little more detail:

- manage publishing packages in tandem (release multiple packages with the same version, together)
- the ability to release pre-release packages
- verify to-be-released packages on correct Semantic Versioning usage (currently with api-extractor)
  - automatic Semantic Versioning based on this mechanism
- automatic changelog generation (combined with Semantic Versioning?)
- automatic git-tagging of releases for easy traceability
- have a consistent build/test/deploy/release mechanism across (types of) packages
  - distinguish between types of packages and have adaptable build options that are consistent in each type (e.g. libs are built/linted/tested in their specific way, apps are built/linted/tested in their own way)
- no phantom dependencies


What features do we get from Rush even though it's not a specific requirement as such:

- pnpm workspace:* version references in package.json are resolved when releasing packages
- specify node supported versions
- ability to distinguish on project types (currently we have `apps`, `libs`, `tools`)
- ability to choose between cache mechanisms 
  - tar via `build-cache.json` or 
  - via git repo state, when a project was previously built and the repo state didn't change)
- seamless integration of repo wide commands (currently `rush fix-versions`, `format`, `lint` 
  - wide range of customization when integrating with the other rush commands (e.g. `enableParrallelism`, `ignoreDependencyOrder`, etc)
- phased builds to improve CI times
- make warnings non-blocking in local development, and blocking in build
- API documentation generation
-

webpro:
Here are comments that should address most of the concerns expressed by @alber70g.

## Changesets

Looks like Changesets does what we want for changelog generation and publishing. Some relevant docs:

- [Using Changesets](https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md)
- [Prereleases](https://github.com/changesets/changesets/blob/main/docs/prereleases.md)
- [Fixed packages](https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md)

## pnpm

- The `workspace:` protocol is a feature (invention, even?) of pnpm.
- Minimum Node.js version can be set with `package.json#engineStrict`.
- Distinguish project types, why exactly? Why is a simple flat `packages/*` structure not sufficient?
- pnpm (and Turborepo) has [filtering](https://pnpm.io/filtering) to target specific packages. Complex filters could be stored in `package.json#scripts`.
- Caching is handled by Turborepo and `actions/setup-node@v3`.
- Repo commands are done using the package manager, pnpm (docs updated).
- What does "seamless" and "wide range" with the other Rush commands mean? If anything, it will be much easier to install new tooling without us [having to jump through hoops and Rush blocking pull requests](https://github.com/kadena-community/kadena.js/pull/566).
- We didn't use phased builds in Rush; what would we miss?
- About warnings...
  - [`allowWarningsInSuccessfulBuild: false` is a mistake](https://github.com/kadena-community/kadena.js/pull/442#issuecomment-1604214895)
  - [Here's another recent example why](https://github.com/kadena-community/kadena.js/pull/660/files#diff-44bdaab7f1499cc925897e808bb19f57354f4ee163e8015c70208b4d56d2fddd)
  - [It's just confusing](https://github.com/kadena-community/kadena.js/pull/643#issuecomment-1647720409)
- I'm sure we can use API Extractor without Rush.

alber70g:
FYI, Rush now also supports pnpm 8 https://github.com/microsoft/rushstack/blob/main/apps/rush/CHANGELOG.md#:~:text=Add%20support%20for%20PNPM%208.

webpro:
Rush has the principle of not allowing a `package.json` and `node_modules` ("hoisted modules"). This issue is mainly relevant when using npm, which defaults to **hoisted** `node_modules`. However, pnpm defaults to **isolated** `node_modules`.

This still does not address the issue that dependencies in the root of the project can be used from packages. Options to address this include:

- [Plug'n'Play. The strictest configuration](https://pnpm.io/blog/2020/10/17/node-modules-configuration-options-with-pnpm#plugnplay-the-strictest-configuration) - this leads to other issues and is too strict for us.
- Lint: use [the upcoming project linter](https://github.com/kadena-community/kadena.js/pull/566) to also lint the root "package", and/or a tool like [Knip](https://github.com/webpro/knip).
- Use [hooks](https://pnpm.io/pnpmfile#hooks) like in 40869d4445aef6901a4487bf3390abef8d2343ca so `pnpm install` will throw when dependencies are found in the root package.

alber70g (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,6 +5,6 @@ Thanks for contributing to this project!
 - Link to the related issue (if any)
 - Add use cases, scenarios, images and screenshots
 - Add documentation and tutorials
-- Run `rush test` and `rush change`
+- Run `pnpm install` and `pnpm test`
```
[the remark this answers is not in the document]
webpro: This is handled in another PR: #712

packages/tools/heft-rig/profiles/default/tsconfig-base.json
```diff
@@ -21,8 +21,9 @@
     "types": [],
 
     "module": "commonjs",
-    "target": "es2017",
-    "lib": ["es2017", "DOM"]
+    "target": "es2019",
+    "lib": ["es2019", "DOM"],
+    "skipLibCheck": true
```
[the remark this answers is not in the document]
webpro: 16 is EOL in a few weeks so I think we're good

workspace.json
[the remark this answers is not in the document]
webpro: I renamed it to `packages.json` and added a script to update it

PR #616, opened 2023-07-18
JSON parse try/catch

Takadenoshi:
JSON.parse was not try/caught before

Introduce a lenient json parsing for heights/data (not fatal) but make it fatal for initial event (need this to validate the config)


webpro (approved):
LGTM

PR #617, opened 2023-07-18
Feat/chromatic

Ghislain89:
This PR:

adds basic visual regression testing with Chromatic. If a change is done to the react-ui library that results in a visual change, this will have to be approved (or denied). Workflow only runs if a commit is being done in the react-ui packages.
Enables a11y for Storybook allowing developers to directly check storybook for any accessibility issues with components.

---
  - https://app.asana.com/0/0/1204934727485712


eileenmguo (approved):

timoheddes (approved):

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,38 @@
+name: Chromatic
+# All pull requests, and
+# Workflow dispatch allows you to run this workflow manually from the Actions tab
+on:
+  pull_request:
+    paths:
+      - packages/libs/react-ui/**
+  push:
+    branches:
+      - main
```
webpro, line 10: This means that the workflow also runs after the merge, is that intentional?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,38 @@
...
+  workflow_dispatch:
+
+jobs:
+  chromatic:
+    name: Visual Regression Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history, needed to determine diffs.
```
webpro, line 22: I think it's better to only fetch the `main` branch right? You can do something like this:

```
      - name: Fetch base branch
        if: github.ref_name != 'main'
        run: git fetch origin ${{ github.base_ref }}
```

Then the `fetch-depth` can be set to `1` or `2`.

.github/workflows/chromatic.yml
Ghislain89: As far as I can see, all tests run again after a merge to main. I intended to follow this convention by doing the same for these. Practically, we don't **have** to run them again.

.github/workflows/chromatic.yml
Ghislain89: Chromatic requires git history for all branches to be able to determine diffs ([according to the docs](https://www.chromatic.com/docs/github-actions#support-for-codeactionscheckoutv2code-and-above)) I'd be willing to try your suggestion but I'm not entirely sure it would work.

Ghislain89: Based on a different fetch depth, I had odd results with when it was able to determine diffs or not.

Randynamic (approved):

PR #618, opened 2023-07-18
Bug: Tooltip Storybook story is erroring

bearni95:
- Closes ["Bug: Tooltip Storybook story is erroring"]()
- Seems that the error is related to using `forwardRef` and the imports into storybook
- It works alright when importing the `Tooltip` component and `tooltipHandler` method from their source files, rather than the index

### DIFF
--- a/packages/libs/react-ui/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/libs/react-ui/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,7 +1,8 @@
 import { IconButton, SystemIcon } from '..';
 
 import { container } from './stories.css';
-import { ITooltipProps, Tooltip } from '.';
+import { ITooltipProps, Tooltip } from './Tooltip';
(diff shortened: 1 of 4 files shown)


eileenmguo (commented):
Did you check if the tooltip works in consuming apps as well since they will be exported from the root directory barrel file?

bearni95:
> Did you check if the tooltip works in consuming apps as well since they will be exported from the root directory barrel file?

Now that actually unlocked it! Turns out the first index for components was exporting `Tooltip/Tooltip`, rather than `Tooltip` (the component's index). Storybook works fine with it now, as well as the test on the docs app (not included in the commits)

eileenmguo (approved):
I just realized something else about the tooltip. It currently has a root and handler, but only the Tooltip.Root is a component. I would expect this component to actually just export `Tooltip` as a component and `tooltipHandler` separately. This would be outside of this PR, but what do you think about this?

PR #626, opened 2023-07-19
@kadena/chainweb-stream-client - Changed default export at index to named export

Takadenoshi:
Default export didn't work with out CJS-targeted output

Also broke the example in the README

### DIFF
--- a/common/changes/@kadena/chainweb-stream-client/chore-chainweb-stream-client-change-default-export_2023-07-19-12-19.json
+++ b/common/changes/@kadena/chainweb-stream-client/chore-chainweb-stream-client-change-default-export_2023-07-19-12-19.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/chainweb-stream-client",
+      "comment": "Changed default export at index to named export",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@kadena/chainweb-stream-client"
+}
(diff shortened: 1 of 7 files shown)


alber70g (approved):

common/changes/@kadena/chainweb-stream-client/chore-chainweb-stream-client-change-default-export_2023-07-19-12-19.json
alber70g, line 1: This one needs to be created with changesets now

alber70g (approved):

common/changes/@kadena/chainweb-stream-client/chore-chainweb-stream-client-change-default-export_2023-07-19-12-19.json
Takadenoshi: done

Takadenoshi:
This needs publishing, the 0.1.0 version that was published to npm from main overwrote the 0.0.10 version that had this fix

alber70g (approved):

webpro:
> This needs publishing, the 0.1.0 version that was published to npm from main overwrote the 0.0.10 version that had this fix

- We should not bump `version` ourselves manually (Changesets does this for us)
- On the same note, we do need to make sure the changelog reflects the bump

I see `minor` in the changelog, but a `patch` bump in `package.json#version`

Takadenoshi:
- I reverted the manual bump in package.json
- I changed the changeset bump to a patch

Ideally I wanted to stay on 0.0.x for a while until it is a bit more stable but now we are on 0.1.x

PR #628, opened 2023-07-19
feat(docs-most-popular): implement most popular pages component + api

realdreamer:
[Closes most popular pages](https://app.asana.com/0/1204958723576385/1205069074849202)


packages/apps/docs/src/components/MostPopular/MostPopular.tsx
```diff
@@ -0,0 +1,26 @@
+import { BrowseSection } from '../BrowseSection';
+
+import { useFetchMostPopularPages } from './useFetchMostPopularPages';
+
+import Link from 'next/link';
+import React from 'react';
+
+const MostPopular = (): JSX.Element => {
+  const { data } = useFetchMostPopularPages();
```
sstraatemans, line 9: we could do this call in getStaticProps
that way it is done in build time instead of seperate api call.

packages/apps/docs/src/components/MostPopular/MostPopular.tsx
```diff
@@ -0,0 +1,26 @@
+import { BrowseSection } from '../BrowseSection';
+
+import { useFetchMostPopularPages } from './useFetchMostPopularPages';
+
+import Link from 'next/link';
+import React from 'react';
+
+const MostPopular = (): JSX.Element => {
```
sstraatemans, line 8: can you use FC (more inline with the other components (also with the react-ui lib components))

packages/apps/docs/src/pages/api/mostPopular.ts
```diff
@@ -0,0 +1,171 @@
+import {
```
sstraatemans, line 1: can we build in to ignore the `__tests` results

sstraatemans (changes requested):
got some comments

sstraatemans (approved):

PR #631, opened 2023-07-20
fix(client): addKeyset

javadkh2:
fix addKeyset

### DIFF
--- a/packages/libs/client/src/client/client.ts
+++ b/packages/libs/client/src/client/client.ts
@@ -32,7 +32,10 @@ import {
 type IOptions = IPollOptions &
   (INetworkOptions | { networkId?: undefined; chainId?: undefined });
 
-interface ISubmit {
+/**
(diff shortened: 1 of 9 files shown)


packages/libs/client/src/client/client.ts
```diff
@@ -294,6 +297,7 @@ export const getClient: IGetClient = (host = kadenaHostGenerator): IClient => {
         chainId: cmd.meta.chainId,
         networkId: cmd.networkId,
       });
+      console.log('send', hostUrl);
```
ash-vd, line 300: Forgotten console.log?

ash-vd (approved):

webpro (approved):

packages/libs/client/src/client/client.ts
javadkh2: removed

PR #632, opened 2023-07-20
FormStatusNotification component

sanderlooijenga:
It uses the Notification component of `@kadena/react-ui` and made it more generic and reusable on other pages as well. Implementations on (other) pages will be done in separate PRs.

---
  - https://app.asana.com/0/0/1205042506053898


packages/apps/transfer/locales/en/common.json
```diff
@@ -99,5 +99,13 @@
   "Transfer pending - wating for proof": "Transfer pending - wating for proof",
   "Transfer pending - waiting for continuation command": "Transfer pending - waiting for continuation command",
   "Could not obtain proof information": "Could not obtain proof information",
-  "Transfer completed successfully": "Transfer completed successfully"
+  "Transfer completed successfully": "Transfer completed successfully",
+  "form-status-title-erroneous": "Something went wrong",
```
MRVDH, line 103: I think we need to decide what kind of keys we should use here. Regular text, or domain-specific, or something else

packages/apps/transfer/src/pages/faucet/existing/index.tsx
```diff
@@ -101,6 +85,9 @@ const ExistingAccountFaucetPage: FC = () => {
       <StyledForm onSubmit={onFormSubmit}>
         <FormStatusNotification
           status={requestStatus.status}
+          statusBodies={{
+            successful: 'The coins have been funded to the given account.',
```
MRVDH, line 89: translation

packages/apps/transfer/locales/en/common.json
jessevanmuijden, line 103: why did you deviate from the original convention of using the human readable string as key? I prefer the kebab-case by the way.

packages/apps/transfer/src/services/faucet/index.ts
```diff
@@ -28,11 +28,11 @@ const debug = Debug('kadena-transfer:services:faucet');
 export const fundExistingAccount = async (
   account: string,
   chainId: ChainwebChainId,
+  amount = 100,
   onPoll?: (
     transaction: IPactCommand & ICommandBuilder<Record<string, unknown>>,
     pollRequest: Promise<IPollResponse>,
   ) => void,
-  amount = 100,
```
jessevanmuijden, line 35: don't we need this?

packages/apps/transfer/src/pages/faucet/existing/index.tsx
```diff
@@ -25,7 +26,10 @@ import React, {
 // TODO: This needs to be changed to 100, when the contract is redeployed
 const AMOUNT_OF_COINS_FUNDED: number = 20;
 
-export type RequestStatus = 'not started' | 'pending' | 'succeeded' | 'failed';
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isCustomError = (error: any): error is ICommandResult => {
```
jessevanmuijden, line 30: could we use `unknown` here?

packages/apps/transfer/src/pages/faucet/existing/index.tsx
jessevanmuijden, line 89: should the string not be translated here?

packages/apps/transfer/src/pages/faucet/existing/index.tsx
```diff
@@ -34,56 +38,36 @@ const ExistingAccountFaucetPage: FC = () => {
...
-        await fundExistingAccount(
-          accountName,
-          chainID,
-          onPoll,
-          AMOUNT_OF_COINS_FUNDED,
-        );
+        await fundExistingAccount(accountName, chainID, AMOUNT_OF_COINS_FUNDED);
 
-        setRequestStatus({ status: 'succeeded' });
+        setRequestStatus({ status: 'successful' });
```
jessevanmuijden, line 54: Perhaps we could use a string enum to hold the possible statuses.

packages/apps/transfer/src/components/Global/FormStatusNotification/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const { t } = useTranslation('common');
+  const [show, setShow] = useState(status !== 'idle');
+
+  useEffect(() => {
+    setShow(status !== 'idle');
+  }, [status]);
+
+  const titles: Titles = {
+    ...{
+      idle: t('form-status-title-idle'),
```
jessevanmuijden, line 50: we could consider using nested translation to slightly reduce code duplication by being able to use `t('form-status-title', {status: 'successful'})` here

packages/apps/transfer/src/pages/faucet/existing/index.tsx
sanderlooijenga: Sure, will update

packages/apps/transfer/src/pages/faucet/existing/index.tsx
sanderlooijenga: We're already achieving that, `status` is of `type FormStatus = 'idle' | 'successful' | 'erroneous' | 'processing';`

packages/apps/transfer/locales/en/common.json
sanderlooijenga: True, I don't really mind which one, this approach seemed more logical to me. I guess I will transform it to be in line with the approach you took

packages/apps/transfer/src/services/faucet/index.ts
sanderlooijenga: It's still in there, just move one position up of parameters, see change on the right

packages/apps/transfer/src/pages/faucet/existing/index.tsx
sanderlooijenga: Done ✅

packages/apps/transfer/src/components/Global/FormStatusNotification/index.tsx
sanderlooijenga: Ah yes, good suggestion. But depends on which approach we're taking with the translation keys. If we're sticking to what we're doing now, I guess it doesn't make much sense perhaps..

packages/apps/transfer/locales/en/common.json
sanderlooijenga: Will be refactored later, in a separate PR. Created a ticket for this; https://app.asana.com/0/1204958723576385/1205129264294989/f

sanderlooijenga: Will be refactored later, in a separate PR. Created a ticket for this; https://app.asana.com/0/1204958723576385/1205129264294989/f

packages/apps/transfer/src/components/Global/FormStatusNotification/index.tsx
sanderlooijenga: Will be refactored later, in a separate PR. Created a ticket for this; https://app.asana.com/0/1204958723576385/1205129264294989/f

MRVDH (approved):

PR #633, opened 2023-07-20
feat(docs): poc of new search

sstraatemans:
The current search loads an index file. 
And that is already 1MB.
So in search of something else.
This solution uses the same OpenAI and PineCone as the AI search we are using


packages/apps/docs/package.json
```diff
@@ -56,7 +54,8 @@
     "remark-frontmatter": "~4.0.1",
     "remark-gfm": "~3.0.1",
     "shiki": "~0.14.2",
-    "styled-components": "~5.3.10"
+    "styled-components": "~5.3.10",
```
eileenmguo, line 57: what are we using styled components for?

eileenmguo (approved):

packages/apps/docs/package.json
sstraatemans: it is used for the component that creates the API Swagger page.

packages/apps/docs/src/utils/embedder.ts
webpro, line 1: Re. filename: I think this is not an "embedder" in that it does the embedding, but rather it creates one.

packages/apps/docs/src/utils/embedder.ts
```diff
@@ -0,0 +1,34 @@
...
+    method: 'POST',
+    body: JSON.stringify({ input: text, model: 'text-embedding-ada-002' }),
+  });
+
+  const { error, data } = await response.json();
+
+  if (error !== undefined) throw new Error(error.message);
+
+  if (data[0].embedding === undefined)
+    throw new Error('No embedding returned from the completions endpoint');
```
webpro, line 27:
```suggestion
    throw new Error('No vector returned from the embeddings endpoint');
```

packages/apps/docs/src/pages/api/semanticsearch.ts
```diff
@@ -0,0 +1,129 @@
...
+    });
+    res.end();
+  }
+
+  const queryEmbedding = await embed(query as string);
+
+  const result = await index.query({
+    queryRequest: {
+      vector: queryEmbedding.values,
+      topK: 50,
```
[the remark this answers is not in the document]
sstraatemans: that is a good point. it was not. it is different now

packages/apps/docs/package.json
webpro: It's not referenced anywhere, so looks like it shouldn't be in this list? Not part of this PR, though.

packages/apps/docs/package.json
```diff
@@ -56,7 +54,8 @@
...
+    "@pinecone-database/pinecone": "~0.1.6"
```
webpro, line 58: Please run `rush[x] format`

packages/apps/docs/src/components/Search/Search.tsx
```diff
@@ -0,0 +1,62 @@
...
+  } = useSearch();
+  const {
+    results: semanticResults,
+    error: semanticError,
+    isLoading: semanticIsLoading,
+    handleSubmit: handleSemanticSubmit,
+  } = useSemanticSearch(limitResults);
+
+  useEffect(() => {
+    if (query?.trim() !== undefined && query?.trim() !== '') {
```
webpro, line 30:
```suggestion
    if (query !== undefined && query.trim() !== '') {
```

packages/apps/docs/src/components/Search/components/SearchResults.tsx
```diff
@@ -63,30 +75,52 @@ export const SearchResults: FC<IProps> = ({
...
+                {semanticError}
+              </Notification.Root>
+            ) : (
+              <>
+                <ResultCount count={semanticResults.length} />
+                <StaticResults
+                  limitResults={limitResults}
+                  results={semanticResults}
+                />
+                {limitResults !== undefined && query !== undefined ? (
```
webpro, line 99: Didn't we have less crazy eslint rules for JSX now? So we can just use `!query`, also rendering `null` when `query` is any falsey value (such as the empty string we are handling in the effect above).

webpro (approved):

trendzetter:
I tried to review this using the test deployment but search didn't produce any results

PR #643, opened 2023-07-24
Feat/package upgrade reactui

timoheddes:
Updated outdated packages for react-ui.
PR touches on other packages and libs as well so please review and check if everything still works as it should.

We (react-ui) plan to do this every sprint so we can stay up to date and let it get out of hand. This PR doesn't update everything. Major versions with breaking changes were avoided for now.


packages/apps/graph-client/package.json
```diff
@@ -47,10 +46,10 @@
     "@rushstack/heft": "~0.50.6",
     "@types/heft-jest": "~1.0.3",
     "@types/lodash.isequal": "~4.5.6",
-    "@types/node": "^16.0.0",
+    "@types/node": "^20.4.2",
```
webpro, line 49: What are the implications of this? We still support Node.js v16, so we might get in trouble with this?

packages/apps/transfer/src/pages/faucet/existing/index.tsx
```diff
@@ -123,7 +123,11 @@ const ExistingAccountFaucetPage: FC = () => {
               leftIcon: SystemIcon.KIcon,
             }}
           />
-          <ChainSelect onChange={onChainSelectChange} value={chainID} />
+          <ChainSelect
+            onChange={onChainSelectChange}
+            value={chainID}
+            ariaLabel="Account name"
+          />
```
webpro, line 130: Nice, I saw this issue in `main` too. A bit odd to put these changes in a single PR, but fine by me.

packages/libs/client-examples/src/example-contract/crosschain-transfer.ts
```diff
@@ -89,7 +89,7 @@ export async function doCrossChainTransfer(
           pactId: status.continuation?.pactId,
           proof,
           rollback: false,
-          step: '1',
+          step: 1,
```
webpro, line 92: Nice, I saw this issue in `main` too

packages/apps/transfer/src/pages/faucet/existing/index.tsx
timoheddes: Ah yeah, happened to spot it and it seems minor so I figured I'll fix it right away.

packages/apps/graph-client/package.json
timoheddes: Yeah you're right, doesn't make sense to update this when we still support 16. I'll revert this.

timoheddes:
Not really sure why it's failing. Locally it builds fine and the logs seem to only give warnings.

webpro:
> Not really sure why it's failing. Locally it builds fine and the logs seem to only give warnings.

It's odd that linting is part of `rush build`, but formatting issues should be fixed:

Screenshot 2023-07-24 at 1 08 43 PM

timoheddes:
Yeah but they are warnings, shouldn't fail the build? These formatting issues weren't introduced by this PR. I'll fix them and see if it makes a difference though.

webpro:
> Yeah but they are warnings, shouldn't fail the build?

[Agreed](https://github.com/kadena-community/kadena.js/pull/442), but negative.

timoheddes:
I see 🥲

webpro:
FYI, perhaps good to know about some upcoming tools that can assist with staying on top of things:

- [syncpack](https://github.com/JamieMason/syncpack) will come with with #612, this helps to keep dependency versions in sync across packages
- [knip](https://github.com/webpro/knip) will come after to help with detecting unused dependencies

eileenmguo (approved):

webpro (approved):

ggobugi27 (approved):

jermaine150 (approved):

PR #646, opened 2023-07-24
[@tools-app] Add bottom navigation to tools app

KristinaSpasevska:


packages/apps/transfer/src/components/Common/Layout/partials/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const openModal = (): void => renderModal(<OptionsModal />);
+
+  const toggleTheme = (): void => {
+    const newTheme = theme === 'dark' ? 'light' : 'dark';
+    setTheme(newTheme);
+  };
+
+  const links = [
+    {
+      title: 'Tutorial',
```
MRVDH, line 25: We can use the `t(` translation module here

packages/apps/transfer/src/components/Common/Layout/partials/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { linkClass } from './styles.css';
+
+import { OptionsModal } from '@/components/Global/OptionsModal';
+import Link from 'next/link';
+import { useTheme } from 'next-themes';
+import useTranslation from 'next-translate/useTranslation';
+import React, { FC } from 'react';
+
+const FooterWrapper: FC = () => {
+  const { theme, setTheme } = useTheme();
```
MRVDH, line 12: Is the theme preference saved in the session? For example if it is dark and you set it to light and reload the page, will it still be light?

packages/apps/transfer/src/components/Common/Layout/partials/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const { renderModal } = useModal();
+
+  const { t } = useTranslation('common');
+  const openModal = (): void => renderModal(<OptionsModal />);
+
+  const toggleTheme = (): void => {
+    const newTheme = theme === 'dark' ? 'light' : 'dark';
```
MRVDH, line 19: Can you base the default theme on the user's system settings? https://stackoverflow.com/a/57795518/5816268

packages/apps/transfer/src/components/Global/OptionsModal/OptionsModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  return (
+    <Modal title="Settings">
+      <div className={modalOptionsContent}>
+        {renderOptions()}
+
+        <div className={modalButton}>
+          <Button.Root
+            title="Save"
```
MRVDH, line 98: Translations

packages/apps/transfer/src/components/Global/OptionsModal/styles.css.ts
```diff
@@ -0,0 +1,40 @@
+import { sprinkles } from '@kadena/react-ui/theme';
+
+import { style } from '@vanilla-extract/css';
+
+export const modalOptionsContent = style([
```
MRVDH, line 5: I think we have to postfix all these exported consts with `Style`, so `modalOptionsContentStyle`

packages/apps/transfer/src/constants/dev-options.ts
```diff
@@ -0,0 +1,34 @@
...
+  text: string;
+  icon: React.FC;
+  tag?: string;
+}
+
+export const devOptions: {
+  [Key in DevOption]: IDevOption;
+} = {
+  BASIC: {
+    title: 'Basic User Interface',
```
MRVDH, line 18: Translations

packages/apps/transfer/src/pages/_app.tsx
```diff
@@ -1,17 +1,33 @@
...
 import React, { FC } from 'react';
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
 const App: FC<AppProps> = ({ Component, pageProps }: AppProps) => (
-  <AppContextProvider>
-    <Layout>
-      <Component {...pageProps} />
-    </Layout>
-  </AppContextProvider>
+  <ThemeProvider
```
MRVDH, line 14: nice, glad we are finally using this 😄

packages/libs/pactjs-test-project/src/example-templates/generated-tx-templates/index.ts
MRVDH, line 1: This file needs to be excluded

MRVDH (changes requested):

packages/apps/transfer/package.json
```diff
@@ -72,6 +72,7 @@
     "ignore-loader": "~0.1.2",
     "jest": "~29.5.0",
     "jest-environment-jsdom": "~29.5.0",
+    "next-themes": "~0.2.1",
```
sanderlooijenga, line 75: Should this be a devDep?

nil-amrutlal-dept (approved):

packages/apps/transfer/src/components/Common/Layout/partials/Footer/Footer.tsx
KristinaSpasevska: Yes, it's saved in storage, so on reload you'll get the last set theme

packages/apps/transfer/src/components/Common/Layout/partials/Footer/Footer.tsx
nil-amrutlal-dept: I believe the `next-themes` library already does that by default according to [this](https://www.npmjs.com/package/next-themes). I also checked the app local storage and when not defined, it is set to system
image

packages/apps/transfer/package.json
MRVDH: This needs to be a regular dep indeed

packages/apps/transfer/src/components/Global/OptionsModal/OptionsModal.tsx
nil-amrutlal-dept: Should the title of the component also use translations? Or just the visible text in the UI?

MRVDH: I think this one is also visible, if you hover over the button and keep the mouse still, right? (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)

If not then ignore what I said here.

nil-amrutlal-dept: Yep, that is correct. It is visible. Will change it

packages/apps/transfer/src/constants/dev-options.ts
nil-amrutlal-dept: Since I can't use `useTranslations` hook outside functional components, maybe I will remove this constants file and integrate this code in the `packages/apps/transfer/src/components/Global/OptionsModal/OptionsModal.tsx`

MRVDH (approved):

PR #647, opened 2023-07-24
chore(docs): add medium2019

sstraatemans:


webpro:
Is there a way to repeat this? Would be good to have scripts or instructions or something to be able to repeat it when db/namespace is removed, cleared, etc

PR #649, opened 2023-07-24
implement sidebar

rezaerami:
- add layout context to control sidebar from pages
- add icons
- add button list
- add accordion and tree to sidebar
- add hide sidebar button

image
image
image
image


MRVDH (changes requested):
1. The colours of the states don't match the design:

https://www.figma.com/file/A11XXPwJSHNSqrRBEFfRc8/K%3AApps?node-id=861%3A3182&mode=dev
image

2. There is a border around the tooltip which is not there in the design

3. In the design there is a small border between the top item and the header, I don't see that here.

3. I am unable to close or open the sidebar, or select anything, clicking doesn't do anything so unable to review further :/

4. When I switch to light mode (removing the dark theme classes) the sidebar becomes light as well. This should remain dark though.

rezaerami:
colors are coming from the ui library, and light vs dark mode apparently are not matching, as you see in react components was the same color but it is changed in react ui
react components color: 
image

the border has been added after these screenshots

open or close only works when there is a sub-menu for each toolbar item so it is not meant to be opened in this PR since @isaozler mentioned there is no need to have the third level menu in transfer tools, but if one page is needed it's better to have it already in place so we can easily show it without breaking change on layout

MRVDH (approved):

eileenmguo (approved):

PR #652, opened 2023-07-25
fix: switch to type only imports/exports

timoheddes:
Be explicit when importing/exporting types.

This will remove all these warnings currently in the console:

Screenshot 2023-07-25 at 11 47 52

https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html


eileenmguo (approved):
Nice! Thanks for cleaning up warnings!

PR #655, opened 2023-07-25
Refactor Button Component

timoheddes:
**Storybook changes:**

- adds controls addon to enable conditional enabling/disabling of controls

**Button changes**

- rendering as anchor tag wasn't working correctly, now fixed
- made icon a prop of Button so we can simply export and import Button
- added iconAlign to also allow controlling of icon position through props
- removed the arbitrary stories that load in slightly different variants of the button which can also be achieved using the controls
- fine-tuned and expanded on the existing controls with conditionals, descriptions etc.


packages/libs/react-ui/.storybook/main.ts
```diff
@@ -18,7 +19,7 @@ const config: StorybookConfig = {
     autodocs: true,
   },
   // Add this for Vanilla Extract
-  webpackFinal(config, options) {
```
timoheddes, line 21: unused

packages/libs/react-ui/src/components/Button/Button.css.ts
```diff
@@ -51,18 +84,18 @@ const colors: Record<ColorType, ColorType> = {
   negative: 'negative',
 };
 
-export const colorVariants = styleVariants(colors, (color) => {
+export const colorVariants = styleVariants(variants, (variant) => {
   return [
     container,
     sprinkles({
-      color: `$${color}Surface`,
```
timoheddes, line 58: Changed this because of: https://github.com/storybookjs/storybook/issues/14704

packages/libs/react-ui/src/components/Button/Button.css.ts
```diff
@@ -37,11 +53,28 @@ export const container = style([
...
 
-const colors: Record<ColorType, ColorType> = {
+const rotate = keyframes({
+  '0%': { transform: 'rotate(0deg)' },
+  '100%': { transform: 'rotate(360deg)' },
+});
+
+export const buttonLoadingClass = style({
+  pointerEvents: 'none',
+});
```
eileenmguo, line 68: Is this necessary if it is being applied via the selector on the `container` class?

packages/libs/react-ui/src/components/Button/Button.test.tsx
```diff
@@ -3,12 +3,50 @@ import { render } from '@testing-library/react';
...
+  it('renders as an anchor element when `as` prop = "a"', () => {
+    const { getByTestId } = render(
+      <Button title="Button" as="a" href="https://kadena.io/">
+        Hello, Button!
+      </Button>,
+    );
+
+    const buttonContainer = getByTestId('kda-button') as HTMLAnchorElement;
+
+    console.log(buttonContainer.nodeName);
```
eileenmguo, line 36: cleanup?

packages/libs/react-ui/src/components/Button/ButtonIcon.tsx
```diff
@@ -1,11 +1,12 @@
...
+type IconProps = JSX.IntrinsicElements['i'];
+export interface IButtonIconProps extends IconProps {
   icon: (typeof SystemIcon)[keyof typeof SystemIcon];
 }
 
-export const ButtonIcon: FC<IButtonIconProps> = ({ icon }) => {
+export const ButtonIcon: FC<IButtonIconProps> = ({ icon, className }) => {
   const Icon = icon;
-  return <Icon size="md" />;
+  return <Icon size="md" className={className} />;
```
eileenmguo, line 11: What is the reason for passing className to the Icon?

packages/libs/react-ui/src/components/Button/index.tsx
```diff
@@ -1,15 +1,4 @@
...
-interface IButton {
-  Root: FC<IButtonProps>;
-  Icon: FC<IButtonIconProps>;
-}
-
-export const Button: IButton = {
-  Root: ButtonContainer,
-  Icon: ButtonIcon,
-};
+export { Button };
```
eileenmguo, line 4: you can write this in one line 
```
export {Button, IButtonProps} from "./Button";
```

eileenmguo (commented):
Looking good! Can you also fix this in consuming apps as part of this PR? The plan is to have all projects that are affected by a react-ui change rebuild so that they have to be fixed in the same PR. We might want to change that in the future and will have a discussion about this on Monday, but for now could you apply those changes here?

packages/libs/react-ui/src/components/Button/index.tsx
timoheddes: You can... but: https://github.com/kadena-community/kadena.js/pull/652

timoheddes:
> Can you also fix this in consuming apps as part of this PR?

Will do!

packages/libs/react-ui/src/components/Button/index.tsx
timoheddes: Ah I was mixing up things. Changed it now! And I changed it for all components in the PR I mentioned :D

packages/libs/react-ui/src/components/Button/Button.css.ts
timoheddes: I didn't realise I left the selector in. Will remove that as it doesn't work in VE.

packages/libs/react-ui/src/components/Button/ButtonIcon.tsx
timoheddes: Because I pass the `iconClassname` when the button is in a loading state, which animates the loading icon.

packages/apps/transfer/src/components/Global/WalletConnectButton/index.tsx
```diff
@@ -6,9 +6,9 @@ import React, { FC } from 'react';
 export const WalletConnectButton: FC = () => {
   const { t } = useTranslation();
   return (
-    <Button.Root title={t('Connect your wallet')} color="positive">
+    <Button title={t('Connect your wallet')} color="positive">
       {t('Connect your wallet')}
       <SystemIcon.Link />
```
eileenmguo, line 11: should this be updated to use the prop?

packages/apps/transfer/src/pages/faucet/existing/index.tsx
```diff
@@ -143,7 +143,7 @@ const ExistingAccountFaucetPage: FC = () => {
             ) : (
               <SystemIcon.TrailingIcon />
             )}
-          </Button.Root>
```
eileenmguo, line 146: Should this also be updated to use the icon and loading prop?

packages/apps/transfer/src/pages/transfer/cross-chain-transfer-tracker/index.tsx
```diff
@@ -182,10 +182,10 @@ const CrossChainTransferTracker: FC = () => {
             />
           </StyledAccountForm>
           <StyledFormButton>
-            <Button.Root title={t('Search')}>
+            <Button title={t('Search')}>
               {t('Search')}
               <SystemIcon.Magnify />
-            </Button.Root>
```
eileenmguo, line 188: icon prop

packages/libs/react-ui/src/components/Button/ButtonIcon.tsx
eileenmguo: Would it be possible to apply the animation via the style attribute? I saw it being done in the transfer app like this 
```
<SystemIcon.Loading
                style={{ animation: '2000ms infinite linear spin' }}`
```

This way we don't make the icon component more flexible. 

If not, I think this is fine as well. More of a nitpick

packages/apps/transfer/src/components/Global/WalletConnectButton/index.tsx
timoheddes: Yes, to all of those haha I was too focused on just fixing the errors. Will update them.

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,43 +1,92 @@
...
 export const Button: FC<IButtonProps> = ({
   as = 'button',
-  color = 'primary',
-  onClick,
-  href,
   children,
+  variant = 'primary',
+  href,
+  icon,
+  iconAlign,
```
eileenmguo, line 33: generally I think the icon will be aligned to the right. I think it would be beneficial to give this prop a default value so that most often this prop won't need to be set. What do you think?

eileenmguo (approved):
Looks good! The only thing is that I think making that giving iconAlign prop a default value of `right` would make usability a bit better

packages/libs/react-ui/src/components/Button/Button.tsx
timoheddes: Yeah I noticed the same, good idea! Will change it.

PR #659, opened 2023-07-26
Using single endpoint for search

webpro:
The dual search functionality is working fine, but we're doing the same request twice to obtain an embedding vector in both the regular/semantic search and the chat completion.

This PR has the dual search submission merged into a single search request/response. The `metadata` now includes the original `content`, and it comes as the first event/chunk, so that part should still be fast.

The required changes to include the `content` in the `metadata` have been made in 7d.


packages/apps/docs/src/components/Search/Search.tsx
```diff
@@ -12,43 +12,28 @@ interface IProps {
...
 
-      analyticsEvent(EVENT_NAMES['click:search'], {
-        query,
-        tabName,
-      });
+  useEffect(() => {
+    if (query !== undefined && query.trim() !== '') {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      handleSubmit(query);
+      analyticsEvent(EVENT_NAMES['click:search'], { query });
```
webpro, line 33: Note that the `tabName` is gone here. We could restore it, but they now both contain results for the same query so might be confusing? I think it makes sense to have a generic event for the tab clicks (which we likely already have?)

sstraatemans (approved):

PR #660, opened 2023-07-26
feat: immutable records app boilerplate

barthuijgen:
Immutable-records will be a new dApp meant for auctioning NFTs, as well as being an open source example of a dApp.

This MR just provides the boilerplate, meaning basically next-create-app setup. This is to get started with it and being able to already set up deployment and testing environments for further development and eventually golive.


.github/CODEOWNERS
```diff
@@ -24,3 +24,4 @@
 /packages/tools/integration-tests               @ggobugi27
 /packages/tools/remark-plugins                  @webpro
 /packages/tools/rush-fix-versions               @alber70g @webpro
+/packages/apps/immutable-records                @barthuijgen
```
webpro, line 27: Please remove this or add someone (like @webpro), since you can't approve your own PR.

packages/apps/immutable-records/README.md
```diff
@@ -0,0 +1,34 @@
...
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js/) - your feedback and contributions are welcome!
+
+## Deploy on Vercel
+
+The easiest way to deploy your Next.js app is to use the [Vercel Platform](https://vercel.com/new?utm_medium=default-template&filter=next.js&utm_source=create-next-app&utm_campaign=create-next-app-readme) from the creators of Next.js.
+
+Check out our [Next.js deployment documentation](https://nextjs.org/docs/deployment) for more details.
```
webpro, line 34: Please add this at the top like the other READMEs:

```
<!-- genericHeader start -->

<!-- genericHeader end -->

```

...and run `rush format`.

packages/apps/immutable-records/README.md
```diff
@@ -0,0 +1,34 @@
...
+This project uses [`next/font`](https://nextjs.org/docs/basic-features/font-optimization) to automatically optimize and load Inter, a custom Google Font.
+
+## Learn More
+
+To learn more about Next.js, take a look at the following resources:
+
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js/) - your feedback and contributions are welcome!
```
webpro, line 28: Not sure if we need this promotion here?

packages/apps/immutable-records/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "format:pkg": "sort-package-json",
+    "format:src": "prettier src --write",
+    "lint": "next lint",
+    "start": "next start",
+    "test": "echo 'boilerplate'"
+  },
+  "dependencies": {
+    "@types/node": "^16.0.0",
+    "@types/react": "^18.0.30",
+    "@types/react-dom": "18.0.11",
```
webpro, line 19: Types shouldn't be in `dependencies` (but `devDependencies` instead)

packages/apps/immutable-records/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@vanilla-extract/css": "1.11.1",
+    "eslint": "^8.15.0",
+    "eslint-config-next": "13.4.5",
```
webpro, line 22: More devDeps here

packages/apps/immutable-records/public/next.svg
webpro, line 1: Do we need these images?

packages/apps/immutable-records/src/app/page.module.css
```diff
@@ -0,0 +1,229 @@
...
+}
+
+@keyframes rotate {
+  from {
+    transform: rotate(360deg);
+  }
+  to {
+    transform: rotate(0deg);
+  }
+}
```
webpro, line 229: Is this the _vanilla extract way?_ Or why is that dependency installed? :)

packages/apps/immutable-records/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "@kadena/immutable-records",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "build": "next build",
+    "dev": "next dev",
+    "format": "npm run format:pkg && npm run format:src",
+    "format:ci": "prettier src --check",
+    "format:pkg": "sort-package-json",
+    "format:src": "prettier src --write",
```
webpro, line 11: Please `format:md` like the others.

webpro:
Please add a description, no idea what this is about?

.github/CODEOWNERS
barthuijgen: Approving your own PR is not the purpose of CODEOWNERS, it's to approve when someone else makes changes to that code right? But I'll remove it.

packages/apps/immutable-records/package.json
barthuijgen: This gives me the following, I'll leave it out for now bug can investigate later.
image

webpro: You can add `"@kadena-dev/markdown": "workspace:*",` to devDeps. It's more correct to also add `remark-cli`, but you can omit it for now.

packages/apps/immutable-records/build.mjs
```diff
@@ -0,0 +1,36 @@
...
+
+const build = spawn('node_modules/.bin/next', ['build']);
+
+build.stdout.on('data', (data) => {
+  process.stdout.write(data);
+});
+
+build.stderr.on('data', (data) => {
+  process.stderr.write(muteStderrWarning(data));
+});
```
webpro, line 36: Warnings are annoying, but not worth a workaround like this. Do we need this to let the build succeed in CI? I have a hunch this is caused by https://github.com/kadena-community/kadena.js/blob/main/common/config/rush/command-line.json#L165-L185

webpro (approved):

packages/apps/immutable-records/README.md
webpro: This is not resolved? In general I think comments/issues in PRs should be marked as resolved by the reviewer, not the pull-requester.

barthuijgen: Right, sorry about that. I assumed it would be pretty straightforward since I added the genericHeader as other apps have it. How is it not resolved yet?

packages/apps/immutable-records/build.mjs
barthuijgen: Ah, that does look like exactly the properly we would need to solve this! indeed much better then a custom startup script.

I am wondering if there is a way to enable this for only the "@kadena/immutable-records" app? From what I see this is configured once for the whole repo. That would not be my intention.

Would you rather have this setting changed for the `build` command in the entire repo, or keep the single-app script?

barthuijgen: Looking at it more thoroughly, there doesn't seem to be a build command in `commnad-line.json`, so I'm not sure how `allowWarningsInSuccessfulBuild` would be added.

packages/apps/immutable-records/README.md
barthuijgen: Thanks, I did notice format had some additional changes. It actually deletes the description under the `# @kadena/immutable-records` header, not sure if that is intentional. I moved it to below the generic header now.

webpro: It uses basically this template:

```
# {pkg.name}

{pkg.description}

<img>
```

Definitely a part that's lacking in our docs. I'll see what I can do about that. Thanks for bearing with me!

PR #663, opened 2023-07-26
[@react-ui] Feature: Progress Bar

nil-amrutlal-dept:
image


packages/libs/react-ui/src/components/ProgressBar/ProgressBar.css.ts
```diff
@@ -0,0 +1,109 @@
...
+    }),
+  ],
+  pending: [
+    sprinkles({
+      backgroundColor: '$negativeAccent',
+    }),
+  ],
+  incomplete: [
+    {
+      border: '2px solid #9EA1A6',
```
rezaerami, line 17: don't we have this color in color system?

packages/libs/react-ui/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  title: string;
+  status: 'complete' | 'pending' | 'incomplete';
+}
+
+export const ProgressBar: FC<IProgressBarProps> = ({ checkpoints }) => {
+  return (
+    <div
+      className={progressBarStyle}
+      data-testid="kda-progress-bar"
+      key={'progress-bar'}
```
rezaerami, line 29: since it's not in a loop doesn't need a key

rezaerami (approved):

packages/libs/react-ui/src/components/ProgressBar/ProgressBar.css.ts
```diff
@@ -0,0 +1,109 @@
...
+    width: '$sm',
+    height: '$sm',
+  }),
+]);
+
+export const lineStyle = style([
+  sprinkles({
+    borderRadius: '$sm',
+    height: '$2',
+    width: '$1',
```
eileenmguo, line 82: I see in storybook the line looks a bit thick. We don't have a smaller size, but this is a case where I think it would be fine to use a custom width

packages/libs/react-ui/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+              key={index}
+              data-testid={`kda-checkpoint-container-${index}`}
+            >
+              <div
+                className={classNames(
+                  circleStyle,
+                  circleColorVariant[checkpoint.status],
+                )}
+              />
+              <div className={textContainerStyle}>{checkpoint.title}</div>
```
eileenmguo, line 67: I see that the text has different styles depending on the status, but I don't think they are being applied to this component
image

eileenmguo (commented):
looking good so far!

packages/libs/react-ui/src/components/ProgressBar/ProgressBar.tsx
sanderlooijenga: Key's can be used outside of loops as well btw, see; https://www.youtube.com/watch?v=vXJkeZf-4-4 or read; https://kentcdodds.com/blog/understanding-reacts-key-prop 😁

packages/libs/react-ui/src/styles/sprinkles.css.ts
nil-amrutlal-dept, line 1: Found a typo in the `SystemProperties` which was not allowing me to use the `borderStyle` within sprinkles method

eileenmguo (approved):

PR #667, opened 2023-07-27
Footer fixes in transfer and react-ui

MRVDH:
- Applied the new colour scheme and fixed minor styling issues
- Removed legacy react-components


eileenmguo (approved):
Should this also remove `@kadena/react-components` from the package.json?

MRVDH:
> Should this also remove `@kadena/react-components` from the package.json?

@eileenmguo not yet, since stitches is still used in some places (will be phased out in other PRs)

PR #672, opened 2023-07-27
fix tooltip

rezaerami:
- use position fixed for tooltip since its position is calculated based on screen position of related element

### DIFF
--- a/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
+++ b/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
@@ -14,7 +14,7 @@ export const container = style([
     backgroundColor: '$neutral1',
     color: '$neutral6',
     width: 'max-content',
-    position: 'absolute',
+    position: 'fixed',
     display: 'none',
   }),
   {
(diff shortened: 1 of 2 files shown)


eileenmguo (commented):
What happens if the content is scrollable?

rezaerami:
@eileenmguo added new fix, removed scroll height, so it will always re calculate to current position of the element

eileenmguo (commented):
Could we do something about the tooltip positioning on scroll? Right now it stays in the same spot on the screen when the page moves and looks a bit broken.  Maybe just making it disappear right away?

rezaerami:
@eileenmguo 
i have pushed new changes for scroll, it also supports the scrollable elements
image
as you see this box is scrollable but the tooltip is shown in the correct position

the only issue is with IFrames which miscalculates the position since getBoundingClientRect will give you the position relative to the top of the screen which will not be correct for an iframe that doesn't fit the entire screen

let me know your thoughts about it, but anyways, the positioning issue that you saw is now fixed

eileenmguo (approved):

PR #674, opened 2023-07-27
[@kadena/react-ui] Fix Card Component Styling

MRVDH:


packages/libs/react-ui/src/components/Card/Card.css.ts
```diff
@@ -1,26 +1,33 @@
...
     width: 'max-content',
     position: 'relative',
   }),
   {
-    border: `${vars.borderWidths.$md} solid ${vars.colors.$neutral2}`,
-    color: fallbackVar(textColor, vars.colors.$neutral6),
+    border: `1px solid ${hexToRgba(colorPalette.$gray40, 0.4)}`,
+    selectors: {
+      [`${darkThemeClass} &`]: {
+        backgroundColor: hexToRgba(colorPalette.$gray60, 0.4),
```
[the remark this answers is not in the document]
MRVDH: Correct, this was a mistake in Figma, I have picked the closest colour from the palette now. 👍

eileenmguo (dismissed):

sstraatemans (approved):

packages/libs/react-ui/src/styles/vars.css.ts
```diff
@@ -150,23 +145,29 @@ export const vars = createGlobalTheme(':root', {
...
+
+    $background: colorPalette.$gray10,
+    $foreground: colorPalette.$gray100,
+
+    $neutral1: colorPalette.$gray10,
+    $neutral2: colorPalette.$gray20,
+    $neutral3: colorPalette.$gray40,
+    $neutral4: colorPalette.$gray60,
+    $neutral5: colorPalette.$gray90,
+    $neutral6: colorPalette.$gray100,
```
eileenmguo, line 162: Do we want to deprecate these? These are still dynamic depending on the theme, but I'm not sure how well they have been working in context

eileenmguo: What are your thoughts @sstraatemans and @isaozler

sstraatemans: At the moment I still use these.
If we have them in another place / name and they are still dynamic then let me know. I can change my components.
But sometimes they are needed.

isaozler: I suggest moving to the new tokens in the ideal world, but I don't know the best development-wise. Because otherwise, it could also create a lot of confusion/chaos...

eileenmguo: The value in those neutral color tokens is that they change colors depending on the applied theme. If we use `gray10` we will have to manually add a dark theme version. My opinion is that we still want to keep those tokens and that functionality and instead of deprecating them, we just make sure they are using the correct colors

MRVDH: I haven't really been able to apply the neutral colours yet with the new colour system though. Because even though almost all of the colours from neutral are in the new colour palette, the light and dark pairs don't match with most of the designs (gray10 light + gray 90 dark instead of gray20 light + gray 90 dark for example)

eileenmguo: yeah, since @isaozler is out right now, we will have to wait to refine them. I think this is fine and we can revisit this when he gets back

eileenmguo (approved):

packages/libs/react-ui/src/styles/vars.css.ts
MRVDH: Created a task for it on the board: https://app.asana.com/0/1204958723576385/1205169835342679/f

PR #690, opened 2023-07-31
implement drawer

rezaerami:
- add drawer component
- add resource link component
- use drawer and resource link component in home

image
image

### DIFF
--- a/packages/apps/tools/src/components/Common/DrawerToolbar/index.tsx
+++ b/packages/apps/tools/src/components/Common/DrawerToolbar/index.tsx
@@ -1,0 +1,63 @@
+import { IconButton, SystemIcon, Text } from '@kadena/react-ui';
+
+import {
+  buttonWrapperClass,
+  expandedDrawerContentClass,
+  expandedDrawerTitleClass,
+  gridItemCollapsedSidebarStyle,
+} from './styles.css';
+
+import classNames from 'classnames';
+import React, { type ReactNode, FC, useState } from 'react';
+
+export interface IDrawerToolbarSection {
+  icon: (typeof SystemIcon)[keyof typeof SystemIcon];
+  title: string;
+  children: ReactNode;
+}
+interface IProps {
+  sections: IDrawerToolbarSection[];
+}
+
+export const DrawerToolbar: FC<IProps> = ({ sections }: IProps) => {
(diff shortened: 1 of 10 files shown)


packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import { style } from '@vanilla-extract/css';
+
+export const gridItemCollapsedSidebarStyle = style([
+  sprinkles({
+    position: 'fixed',
+    right: 0,
+    top: '$16',
+    bottom: '$10',
+    backgroundColor: '$neutral1',
```
MRVDH, line 11: use of neutral colour, good idea?

packages/apps/tools/src/components/Common/DrawerToolbar/index.tsx
MRVDH, line 1: Could you post some screenshots on how it would look open and closed? Also in terms of content overlap etc

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
rezaerami: on light mode it is set to #fafafa which is neutral1, and on dark mode closest color to #111111 is also neutral1

packages/apps/tools/src/components/Common/DrawerToolbar/index.tsx
rezaerami: added

MRVDH (approved):

packages/apps/tools/src/components/Global/ResourceLinks/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  titleWrapperClass,
+} from './styles.css';
+
+import Link from 'next/link';
+import useTranslation from 'next-translate/useTranslation';
+import React, { FC } from 'react';
+
+interface ILinkProps {
+  title: string;
+  href: string;
```
sanderlooijenga, line 17: Could perhaps also reuse type from `LinkProps`;

```ts
interface ILinkProps extends Pick<LinkProps, 'href'> {
    ...
}
```

or

```ts
href: LinkProps['href']
```

sanderlooijenga (approved):

PR #693, opened 2023-07-31
chore(@kadena/react-ui): Refactor NavFooter component naming

eileenmguo:
Updated some of the footer component names:
- Updated the component from `Footer` to `NavFooter` to make it more descriptive (There will be a NavHeader component as well)
- Updated the subcomponents to match the naming convention of existing components
  - FooterIconItem -> FooterIconButton
  - FooterLinkItem -> FooterLink
- Updated the variant names
  - web -> dynamic (since it changes depending on the theme)
  - application -> dark (since it stays the dark color)

Also updated the tools app's usage to match the new naming convention


sstraatemans (approved):

MRVDH (commented):
I was thinking of removing the 'web' and 'application' toggle. It does not really add anything, since how we fixed this component to always stay dark in the tools app is by simply wrapping it in an element with the darkThemeClass on it.

Let me know what you think, because if we do want to keep the web/app distinction then we have to apply this to a lot of other components as well, which I don't think is a good idea...

PR #694, opened 2023-07-31
Refactor/getClient input

javadkh2:


javadkh2:
We decided to not go with this approach

PR #702, opened 2023-08-01
Set up Chromatic

Ghislain89:
This PR:

adds basic visual regression testing with Chromatic. If a change is done to the react-ui library that results in a visual change, this will have to be approved (or denied). Workflow only runs if a commit is being done in the react-ui packages.
Enables a11y for Storybook allowing developers to directly check storybook for any accessibility issues with components.


eileenmguo (approved):

webpro (commented):
Where can I see something? Got examples of results?

webpro (approved):

PR #706, opened 2023-08-01
Navigation Header component

timoheddes:
### Implements Navigation Header component.
[Figma Link](https://www.figma.com/file/A11XXPwJSHNSqrRBEFfRc8/K%3AApps?type=design&node-id=685-2777&mode=design&t=Es6oj3FDIpnwEEro-0)

Also created new Logo component, see its readme for some more info.

_Note that there is no mobile design available. In the Docs website this was solved by creating a hamburger menu but Steven indicated that this doesn't work ideal and Isa is still going to look into it._

image

image


MRVDH (approved):

packages/libs/react-ui/src/components/NavHeader/NavHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+        style={{ transform: `translateX(${glowX}px)` }}
+        ref={glowRef}
+      >
+        <NavGlow />
+      </div>
+      <div className={logoClass} ref={logoRef}>
+        {logoVariants.includes(brand) && (
+          <Link.Root href="/" target="_self">
+            <Logo variant={brand} />
+          </Link.Root>
```
eileenmguo, line 77: Just a note that this should change in the future: Made a ticket to update the Link to match the Button API so it wouldn't have any subcomponents: https://app.asana.com/0/1205163122576860/1205171349208298/f

packages/libs/react-ui/src/components/NavHeader/NavHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+                  onClick={() => setActiveNav(index + 1)}
+                >
+                  <a
+                    className={classNames(linkClass, {
+                      [activeLinkClass]: activeNav === index + 1,
+                    })}
+                    href={item.href}
+                    target={item.target}
+                  >
+                    {item.title}
```
eileenmguo, line 96: I don't believe this accounts for when Next.js apps want to use the `Link` component

packages/libs/react-ui/src/components/Logo/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    case 'Docs':
+      return KadenaDocsLogo;
+    case 'DevTools':
+      return KadenaDevToolsLogo;
+    default:
+      return KadenaLogo;
+  }
+};
+
+const Logo: FC<{ variant?: LogoVariant }> = ({ variant }) => {
```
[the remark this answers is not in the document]
timoheddes: I was thinking we could extend it later though to include Marmalade logo and so on. What do you think? Otherwise yeah, makes sense to change it to be more explicit.

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -0,0 +1,133 @@
...
+      },
+    },
+  },
+]);
+
+export const activeLinkClass = style([
+  {
+    backgroundColor: vars.colors.$white,
+    color: vars.colors.$gray90,
+  },
```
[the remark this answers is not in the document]
timoheddes: How do I use sprinkles with selectors?

packages/libs/react-ui/src/components/NavFooter/FooterLink.tsx
```diff
@@ -14,11 +14,16 @@ export const FooterLink: FC<IFooterLinkProps> = ({ children }) => {
   };
 
   const clones = React.Children.map(children, (child) => {
-    // @ts-ignore
```
timoheddes, line 17: Unrelated to PR. Just something I noticed and seemed small enough to push through.

packages/libs/react-ui/src/components/NavHeader/NavHeader.tsx
timoheddes: Ah right. I guess a similar approach to the footer is necessary for that.. I'll see if I can think of another solution or update it to that.

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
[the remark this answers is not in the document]
MRVDH: The selectors not, but this section can be sprinkled:

```
sprinkles({
    backgroundColor: '$white',
    color: '$gray90',
  }),
```

packages/libs/react-ui/src/components/Logo/index.tsx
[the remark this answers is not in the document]
eileenmguo: Ahh ok. Maybe `BrandLogo`?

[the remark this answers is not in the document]
timoheddes: Sounds good! Done :)

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -0,0 +1,132 @@
...
+]);
+
+export const navListClass = style([
+  sprinkles({
+    alignItems: 'stretch',
+    display: 'flex',
+    justifyContent: 'center',
+  }),
+  {
+    alignSelf: 'center',
```
eileenmguo, line 56: can you apply this to via the parent to avoid adding these extra styles?

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.children.tsx
```diff
@@ -0,0 +1,17 @@
+import { Button } from '@components/Button';
+import React, { FC } from 'react';
+
+export const NavHeaderChildren: FC = () => {
+  return (
+    <Button
+      as="button"
+      icon="Link"
+      onClick={() => {}}
```
eileenmguo, line 9: I feel like this is unfinished?

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.children.tsx
```diff
@@ -0,0 +1,17 @@
...
+      style={{ marginLeft: '1rem' }}
```
eileenmguo, line 10: Can you move this to styles?

packages/libs/react-ui/src/components/NavHeader/index.ts
```diff
@@ -0,0 +1,7 @@
+export type { INavHeaderProps } from './NavHeader';
+export type { INavHeaderNavigationProps } from './NavHeaderNavigation';
+export type { INavHeaderContentProps } from './NavHeaderContent';
+
+export { NavHeader } from './NavHeader';
+export { NavHeaderNavigation } from './NavHeaderNavigation';
+export { NavHeaderContent } from './NavHeaderContent';
```
eileenmguo, line 7: This needs to be updated to use our subcomponent naming convention
`NavHeader.Root`, `NavHeader.NavLinks`, `NavHeader.Content`

eileenmguo (changes requested):

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+  }) => {
+    const navItems = useCustomNavigation ? customNavigation : sampleNavItems;
+    return (
+      <NavHeader brand={brand}>
+        <NavHeaderNavigation>
+          {navItems.slice(0, linksCount).map((item, index) => {
+            return (
+              <a href={item.href} target={item.target} key={index}>
+                {item.title}
+              </a>
```
eileenmguo, line 112: Why did you decide not to expose a `NavLink.Link` component?

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.children.tsx
timoheddes: It's not. It's just sample content.

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.children.tsx
timoheddes: This is just sample content. It's really irrelevant. I only added it so you can have some example content in the right hand side of the header. I could remove it all together perhaps?

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
timoheddes: To the container? That wouldn't work.

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.children.tsx
eileenmguo: ahh yeah, I missed that this was in a story 👍

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
eileenmguo: The styles would be applied directly to the link, but also it matches the pattern we use for our component structure. So you are able to create the whole navHeader with our components. Currently I don't think we have any similar components that nest native html elements, they all expose buttons or links

eileenmguo (commented):
Did you test this with Next.js Link component?

eileenmguo (commented):
You're missing the following styles: 
Brand focus state: 
image
Link active state - I think @Isa meant "selected" by "active" So it should have the white background when the glow is there 
image

image

timoheddes:
Can't reply to your comment directly @eileenmguo but the "Link active state" you mention is referred to as a "focus" state in the design, which I did implement - though not yet the focused state of an active item, which I will add now.

Design:
image

Implementation, focused nav item:
image

packages/libs/react-ui/src/components/NavHeader/NavHeaderNavigation.tsx
```diff
@@ -0,0 +1,56 @@
...
+export type INavItemTarget = '_self' | '_blank';
+export interface INavItem {
+  href: string;
+  target?: INavItemTarget;
+  title: string;
+}
+export type INavItems = INavItem[];
+
+export interface INavHeaderNavigationProps {
+  children: React.ReactNode;
```
eileenmguo, line 17: Small improvement: You can type the children to be an array of a specific component 
```
export interface IBreadcrumbsProps {
  children?: FunctionComponentElement<IBreadcrumbItemProps>[];
  icon?: (typeof ProductIcon)[keyof typeof ProductIcon];
}
```

eileenmguo (approved):
Minor improvement on the typing, but looks good overall 👍

packages/libs/react-ui/src/components/NavHeader/NavHeader.tsx
```diff
@@ -15,7 +16,7 @@ export type INavItems = INavItem[];
 
 export interface INavHeaderContainerProps {
   brand?: LogoVariant;
-  children?: React.ReactNode;
+  children?: FunctionComponentElement<INavHeaderProps>[];
```
eileenmguo, line 19: hmm is this correct? I believe these props equate to the whole NavHeader object that is getting exported, not each of the individual subcomponents of the NavHeader

PR #708, opened 2023-08-01
fix: client addData and pactjs-cli namespace warnings

javadkh2:
- The client throws an exception if a key is duplicated in the addData function, 
- The pactjs-cli logs a warning message if the namespace is not available.


alber70g (approved):

packages/libs/client/src/composePactCommand/utils/addData.ts
```diff
@@ -9,6 +9,26 @@ export type ValidDataTypes =
...
+  ) {
+    return cmd.payload.cont.data[key];
+  }
+
+  return undefined;
+};
+
 /**
  * Reducer to add `data` to the {@link IPactCommand.payload}
  *
```
alber70g, line 34: We can add `@throws` here
https://tsdoc.org/pages/tags/throws/

packages/tools/pactjs-cli/src/contract-generate/generate.ts
```diff
@@ -104,6 +104,11 @@ async function generator(
   const moduleDtss = new Map();
 
   Object.keys(modules).map((name) => {
+    if (['', undefined, null].includes(modules[name].namespace)) {
+      console.log(`
+      No namespace found for module "${name}". You can pass --namespace as a fallback.
+      `);
```
alber70g, line 110: Maybe we want to prepend the warning with "WARNING:"?

```suggestion
      console.log(`
      WARNING: No namespace found for module "${name}". You can pass --namespace as a fallback.
      `);
```

common/changes/@kadena/pactjs-cli/fix-client-addData-and-namespace-warnings_2023-08-16-11-34.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/pactjs-cli",
+      "comment": "show warning message if namespace is not presented",
```
alber70g, line 5:
```
show warning message if namespace is not present
```

PR #722, opened 2023-08-02
[docs] Refactor Markdown/Code and Markdown/Figure

bearni95:
- Part of [this ticket](https://app.asana.com/0/1204958723576385/1205091936807439/f)
- Code refactor from stitches into vanilla-extract for the Code and Figure markdown components

---
  - https://app.asana.com/0/0/1205194857763852


packages/apps/docs/src/components/Markdown/Code/style.css.ts
```diff
@@ -0,0 +1,154 @@
...
+  }),
+]);
+
+export const codeTitle = style([
+  {
+    display: 'flex',
+    alignItems: 'center',
+    fontFamily: `${vars.fonts.$main}`,
+    background: `${vars.colors.$neutral3}`,
+    padding: `${vars.sizes.$2}`,
```
eileenmguo, line 93: can you move this to sprinkles

packages/apps/docs/src/components/Markdown/Code/Code.tsx
```diff
@@ -1,15 +1,55 @@
...
+          (grandchild: ReactNode) => {
+            if (!React.isValidElement(grandchild) || !grandchild) {
+              return null;
+            }
+
+            return React.cloneElement(grandchild, {
+              ...grandchild.props,
+              className: codeWord,
+            });
+          },
```
eileenmguo, line 44: I'm confused by what's happening here. Can you explain the context?

bearni95: The stitches code used nested styles, which we can't really do with sprinkles:
https://github.com/kadena-community/kadena.js/blob/acfba63a66039844edf8972a235310c40029956e/packages/apps/docs/src/components/Markdown/Code/styles.ts#L9

https://github.com/kadena-community/kadena.js/blob/acfba63a66039844edf8972a235310c40029956e/packages/apps/docs/src/components/Markdown/Code/styles.ts#L106

This nested recursion over children and grandchildren is to assign the `codeLine` and `codeWord` respectively.

eileenmguo: ahh ok. If that is the only reason for this, I think this is an instance where using `globalStyle` would be worth doing considering it's much simpler and easier to maintain/understand

bearni95: Tried using `globalStyle` but for some reason the styles aren't being applied 

Implemented the `grandchildren` portion using globalStyle in the last push, but if visiting a code portion (ie http://localhost:3000/docs/pact/beginner/hello-world) doesn't show the code in the `$mono` font

packages/apps/docs/src/components/Markdown/Code/style.css.ts
```diff
@@ -78,6 +79,10 @@ export const codeLine = style([
   },
 ]);
 
+globalStyle(`${code} span`, {
+  fontFamily: vars.fonts.$mono,
+})
+
```
eileenmguo: Since this is a code block, can you just make everything inside fontFamily mono?

bearni95: This bit doesn't work! Despite what's said in the vanilla-extract docs, using `globalStyle` with custom classes doesn't apply them.

The way to get this block to work, and avoid that children and grandchildren stuff, would be:
````typescript
globalStyle(`code span`, {
  fontFamily: vars.fonts.$mono,
})
```

eileenmguo: hmm strange. As long as this works, I'm ok with it it as well 👍  i think it is still cleaner. What do you think?

bearni95: I agree that's way cleaner, happy to go with it

packages/apps/docs/src/components/Markdown/Code/Code.tsx
```diff
@@ -1,15 +1,40 @@
...
+
+  const [themedProps, setThemedProps] = React.useState(props);
+
+  useEffect(() => {
+    setThemedProps({
+      ...props,
+      'data-active-theme': theme,
+    });
+  }, [props, theme]);
+
```
eileenmguo, line 21: what is the purpose of this theme stuff?

bearni95: Carried over from the original implementation:
https://github.com/kadena-community/kadena.js/blob/277bd6904a9a229a84c42efd6cd389c56d20011b/packages/apps/docs/src/components/Markdown/Code/styles.ts#L87

For some reason both a light and a dark themed version of the code block appear, and one must be hidden.

bearni95: Managed to get it done with pure CSS

packages/apps/docs/src/components/Markdown/Code/Title.tsx
```diff
@@ -1,15 +1,35 @@
...
+
+  const [themedProps, setThemedProps] = React.useState(props);
+
+  useEffect(() => {
+    setThemedProps({
+      ...props,
+      'data-active-theme': theme,
+    });
+  }, [props, theme]);
+
```
eileenmguo, line 21: can we apply the same thing here?

packages/apps/docs/src/components/Markdown/Code/Code.tsx
```diff
@@ -8,8 +8,21 @@ interface IProp {
...
+        if (!React.isValidElement(child) || !child) {
+          return null;
+        }
+
+        return React.cloneElement(child, {
+          ...child.props,
+          className: codeLine,
+        });
+      })}
+    </code>
```
eileenmguo, line 26: Is mapping through the children still necessary?

bearni95: yes! they need to get assigned the `codeLine` class:
```typescript
export const codeLine = style([
  sprinkles({
    width: '100%',
    display: 'inline-block',
    fontFamily: '$mono',
  }),
  {
    selectors: {
      '&::before': {
        counterIncrement: 'line',
        content: 'counter(line)',
        overflowWrap: 'normal',
        display: 'inline-block',
        width: '1rem',
        marginRight: `${vars.sizes.$4}`,
        marginLeft: `${vars.sizes.$4}`,
        textAlign: 'right',
        fontSize: `${vars.sizes.$sm}`,
        color: `${vars.colors.$neutral3}`,
      },
    },
  },
]);

eileenmguo: Do you think this is an instance in which using globalStyle would be a better option?

bearni95: Even if the style applied with the parent class as a selector, we cannot use selectors inside `globalStyle`

eileenmguo: oh i see that is not possible either. Nevermind 👍

eileenmguo (approved):

PR #734, opened 2023-08-03
 [docs] Refactor Markdown/Heading and Markdown/MDNotification

bearni95:
Code refactor from stitches into vanilla-extract for the `<Heading>` and `<MDNotification>` markdown components

---
  - https://app.asana.com/0/0/1205194857763853


packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
```diff
@@ -93,12 +48,101 @@ export const TaggedHeading: FC<IProp> = ({
...
+          onMouseOver={onMouseOver}
+          onFocus={() => {}}
+          className={headerVariants[variant ?? as]}
+        >
+          {children}
+          {ArticleLink}
+        </h2>
+      );
+    case 'h3':
+      return (
```
realdreamer, line 87: why are you changing into switch case here? can't we re-use Heading component and achieve the same?

realdreamer (commented):
can you add suffix as `class` for all classNames? this makes it consistent across our code-base?

packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
bearni95: We need to apply the `headerVariants` class to them, which is not an accepted parameter by `Heading`. Should we change the component instead?

bearni95: The articles use <h1> and <h2> quite often. Without the resize (which makes them smaller than the default) the titles are less aggressive. That said, the resize is here because it was part of the stitches implementation, not sure what motivated it, originally

eileenmguo: by resize do you mean the reponsive fontSize changes on the Heading component styles?

eileenmguo: So I think it's fine to have a separate heading component for this if the styles really are meant to differ greatly, but I think this implementation can be improved. For example, the hover states and focus states can be applied via css rather than passing in props

bearni95: The hover/focus states need to alter the style of`ArticleLink` but are bound to the `Heading` atop of them. Do we have a way to do that in vanilla-extract? Even with `globalStyle` we cannot use pseudo-selectors like `:hover`:
https://vanilla-extract.style/documentation/global-api/global-style/

eileenmguo: looking at your code, it looks like the only style difference on hover is the opacity. You believe you can achieve the same effect by using variables. You can define an `articleLinkOpacity` var and set it to it's default opacity in the normal heading styles, and to the hover opacity through the hover selector. Then you can use this variable in the css definition for the articleLink. Hope this helps and let me know if you have any questions

bearni95: You're absolutely right, didn't even think of using variables for this one!

Just pushed some changes using `useVariable`, works like a charm :smiley:

packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
```diff
@@ -93,12 +44,54 @@ export const TaggedHeading: FC<IProp> = ({
...
-    </StyledHeader>
+  const ArticleLink = (
+    <a className={headerIconLinkClass} id={slug} href={`#${slug}`}>
+      <SystemIcon.Link />
+    </a>
   );
+  switch (as) {
+    case 'h2':
+      return (
+        <h2 className={headerClassVariants[variant ?? as]}>
```
eileenmguo, line 55: another improvement to make this better is move `headerClassVariants[variant ?? as]` to a variable so you can change it in one place rather than copy it to each of the classNames

packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
```diff
@@ -93,12 +44,54 @@ export const TaggedHeading: FC<IProp> = ({
...
   const slug = createSlug(slugInputStr, index, parentTitle);
 
-  return (
-    <StyledHeader as={as} variant={variant ?? as}>
-      {children}
-      <StyledLinkIcon id={slug} href={`#${slug}`}>
-        <SystemIcons.Link />
-      </StyledLinkIcon>
-    </StyledHeader>
+  const ArticleLink = (
```
eileenmguo, line 47: Could you make this `content` and also put children here? Just to reduce duplicated code

packages/apps/docs/src/components/Markdown/MDNotification/style.css.ts
```diff
@@ -0,0 +1,12 @@
...
+import { sprinkles } from '@kadena/react-ui/theme';
+
+import { style } from '@vanilla-extract/css';
+
+export const wrapperClass = style([
+  sprinkles({
+    marginY: '$5',
+    marginX: 0,
+  }),
+]);
```
eileenmguo, line 12: Can you apply these styles in the same way that you did in the other PR? Then I think they will get resolved in the merge by default (I would expect)

timoheddes (approved):

eileenmguo (approved):

PR #735, opened 2023-08-03
 [docs] Refactor Markdown/Paragraph and Markdown/Table

bearni95:
Code refactor from stitches into vanilla-extract for the `<Paragraph>` and `<Table>` markdown components

---
  - https://app.asana.com/0/0/1205194857763854


eileenmguo (commented):
Are you going to apply this same style on a lot of different markdown components? If so, I would recommend moving to the parent directory and using the same style for everything so that you only have one implementation for it

eileenmguo (approved):

PR #717, opened 2023-08-02
 [docs] refactor react-components components into react-ui

bearni95:
## Rebased version [this PR](https://github.com/kadena-community/kadena.js/pull/696) to include only relevant commits

- Closes ["refactor react-components components"](https://app.asana.com/0/1204958723576385/1205091936807435/f)
- This PR replaces the components from the `@kadena/react-components` with their equivalents from `@kadena/react-ui`
- The `Divider` component has been added to `@kadena/react-ui`
- This PR does NOT deal with the following:
  - Components implemented for the docs page specifically
  - Components that depend/extend others that use `@kadena/react-components`

## Altered components:
- [x]  hooks/useMediaQuery.ts
- [x]  components/NotFound/NotFound.tsx
- [x]  components/BottomPageSection/BottomPageSection.tsx
- [x]  components/BottomPageSection/components/Subscribe.tsx
- [x]  components/BottomPageSection/components/EditPage.tsx
- [x]  components/Layout/components/Footer/Footer.tsx
- [x]  components/Layout/components/TitleHeader/TitleHeader.tsx
- [x]  components/Layout/components/SideMenu/SideMenu.tsx
- [x]  components/Layout/components/Header/HamburgerMenuToggle.tsx
- [x]  components/Layout/components/Header/ThemeToggle.tsx
- [x]  components/Layout/components/Header/Header.tsx
- [x]  components/Layout/components/Header/SearchButton.tsx
- [x]  components/Layout/components/Header/AsideToggle.tsx
- [x]  components/Layout/Landing/components/Headers/HomeHeader.tsx
- [x]  components/Layout/Landing/components/Headers/SearchHeader.tsx
- [x]  components/Search/components/SearchResults.tsx
- [x]  pages/_error.tsx
- [x]  pages/docs/chainweb/index.tsx
- [x]  pages/docs/__tests/index.tsx
- [x]  pages/docs/blogchain/2022/index.tsx
- [x]  pages/docs/blogchain/2021/index.tsx
- [x]  pages/docs/blogchain/2018/index.tsx
- [x]  pages/docs/blogchain/2023/index.tsx
- [x]  pages/docs/blogchain/2019/index.tsx
- [x]  pages/docs/marmalade/index.tsx
- [x]  pages/docs/contribute/index.tsx
- [x]  pages/docs/build/index.tsx
- [x]  pages/docs/kadena/index.tsx
- [x]  pages/index.tsx

PR #737, opened 2023-08-04
Fix button style react-ui

timoheddes:
We noted that in Docs, after switching from [react-components to react-ui](https://github.com/kadena-community/kadena.js/pull/717), the hover style of a button that's rendered as anchor wasn't correct. This fixes that by correctly setting the button color and font weight.

Current/faulty hover:
image

New/fixed hover:
image

### DIFF
--- a/packages/libs/react-ui/src/components/Button/Button.css.ts
+++ b/packages/libs/react-ui/src/components/Button/Button.css.ts
@@ -39,30 +39,30 @@ export const container = style([
       '&[href]': {
         display: 'inline-flex',
       },
-      '&:hover': {
-        backgroundColor: bgHoverColor,
-        color: colorPalette.$white,
-      },
-      '&:active': {
-        backgroundColor: bgActiveColor,
-      },
-      '&:focus-visible': {
-        outlineOffset: '2px',
-        outlineWidth: vars.borderWidths.$md,
-        outlineStyle: 'solid',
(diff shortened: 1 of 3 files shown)


packages/libs/react-ui/src/components/Button/Button.css.ts
```diff
@@ -13,44 +13,47 @@ const bgHoverColor = createVar(),
...
+        outlineStyle: 'solid',
+        outlineColor: focusOutlineColor,
+      },
+      '&:disabled': {
+        opacity: 0.7,
+        backgroundColor: '$neutral3',
+        color: '$neutral1',
+        cursor: 'not-allowed',
+        pointerEvents: 'none',
+      },
```
eileenmguo, line 53: These don't need to be placed in selectors to work since they are basic selectors

packages/libs/react-ui/src/components/Button/Button.css.ts
```diff
@@ -13,44 +13,47 @@ const bgHoverColor = createVar(),
...
     },
+    color: 'white',
```
eileenmguo, line 55: what about dark mode?

timoheddes:
Dark/light mode support.

https://github.com/kadena-community/kadena.js/assets/79908211/c98cae8a-f0c1-4755-8931-45373a14e67f

eileenmguo (approved):

PR #738, opened 2023-08-04
fix(docs): header with inline codes

realdreamer:
This PR fixes an issue with headers with inline code
Example - " ## Header with `features` "

### DIFF
--- a/packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
+++ b/packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
@@ -2,13 +2,13 @@ import { styled, StyledComponent, SystemIcons } from '@kadena/react-components';
 import { Heading } from '@kadena/react-ui';
 
 import { createSlug } from '@/utils';
-import React, { FC, ReactNode } from 'react';
+import React, { FC, ReactElement } from 'react';
 
 type TagType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 interface IProp {
   as: TagType;
   variant?: TagType;
-  children: string | string[] | ReactNode[];
+  children: ReactElement[];
   index?: number;
   parentTitle?: string;
 }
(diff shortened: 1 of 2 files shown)


packages/apps/docs/src/scripts/importReadme.mjs
```diff
@@ -46,14 +46,24 @@ export const createSlug = (str) => {
...
-  // TODO: flatten all children recursively to prevent issue with
+  // flatten all children recursively to prevent issue with
   // E.g. ## some title with `code`
   const node = pageAST.children[0];
   if (node.type !== 'heading' || node.depth !== 2) {
     throw new Error('first node is not a Heading');
   }
 
-  return node.children[0].value;
+  return node.children.reduce((acc, val) => {
```
webpro, line 56: Same as a above, `filter` + `join` would be fine?

packages/apps/docs/src/components/Markdown/Heading/Heading.tsx
```diff
@@ -8,7 +8,7 @@ type TagType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 interface IProp {
   as: TagType;
   variant?: TagType;
-  children: string;
+  children: string | string[];
```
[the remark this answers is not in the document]
realdreamer: `unknown` causes an issue with type of `ReactNode`. I added an extra `ReactNode` type

packages/apps/docs/src/scripts/importReadme.mjs
```diff
@@ -46,14 +46,14 @@ export const createSlug = (str) => {
...
-  // TODO: flatten all children recursively to prevent issue with
+  // flatten all children recursively to prevent issue with
   // E.g. ## some title with `code`
   const node = pageAST.children[0];
   if (node.type !== 'heading' || node.depth !== 2) {
     throw new Error('first node is not a Heading');
   }
 
-  return node.children[0].value;
+  return node.children.filter((child) => child.type === 'text').join(' ');
```
webpro, line 56: Could you please test with a title like in the comment:

```
## some title with `code`
```

I think the items of `node.children` can be arrays as well, so you could try e.g. `node.children.flat()`

webpro (approved):

PR #744, opened 2023-08-04
[@kadena/tools] Remove custom generated icons

MRVDH:
- Removed icons that we didn't use. (Removed from header as well since header will be replaced)
- Improved dev startup and build speed


rezaerami (approved):

packages/apps/tools/src/components/Global/RequestKeyField/index.tsx
```diff
@@ -5,7 +5,7 @@ import React, { type FC } from 'react';
 import { FieldError } from 'react-hook-form';
 import * as z from 'zod';
 
-export const RequestLength = { MIN: 43, MAX: 44 };
+export const RequestLength: { MIN: number; MAX: number } = { MIN: 43, MAX: 44 };
```
sanderlooijenga, line 8: We need to disable this ESLint warning IMO, TS should just infer this 😛

sanderlooijenga (approved):

packages/apps/tools/src/components/Global/RequestKeyField/index.tsx
MRVDH: Agreed, but I feel like that's a discussion for another time 😅

sanderlooijenga: Yeah, approved nonetheless

PR #748, opened 2023-08-07
 [docs] Refactor Markdown/UnorderedList

bearni95:
Code refactor from stitches into vanilla-extract for the `<UnorderedList>` markdown components

---
  - https://app.asana.com/0/0/1205194857763855


packages/apps/docs/src/components/Markdown/UnorderedList/styles.css.ts
```diff
@@ -0,0 +1,27 @@
...
+        fontWeight: '$bold',
+      },
+    },
+  },
+]);
+
+globalStyle(`${ulList} ${ilItem} ul`, {
+  paddingLeft: vars.sizes.$4,
+  margin: 0,
+  backgroundColor: 'red',
```
realdreamer, line 26: is it intended?

packages/apps/docs/src/components/Markdown/UnorderedList/styles.css.ts
```diff
@@ -0,0 +1,19 @@
...
+export const ulListClass = style([
+  sprinkles({
+    marginY: '$5',
+    marginX: 0,
+    position: 'relative',
+  }),
+]);
+
+// sub-ul, won't work with classNames
+globalStyle('ul ul', {
```
realdreamer, line 14:
```suggestion
globalStyle(`${ulListClass} ul`, {
```

I think this works and it also makes styles only applied to this specific children of ul instead of affecting all `ul ul`

realdreamer: how about the `li` marker styles, is it fixed separately?

eileenmguo: maybe something like this would work? ```
app.css.ts
import { style, globalStyle } from '@vanilla-extract/css';

export const parentClass = style({});

globalStyle(`${parentClass} > a`, {
  color: 'pink'
});```

bearni95: problem is, at least on my end, `globalStyle` isn't working with the classes defined, only with the html tags instead! It makes no sense, but that's what justifies the style choice here:


> I think this works and it also makes styles only applied to this specific children of ul instead of affecting all `ul ul`

eileenmguo: could you try `globalStyle(`${ulListClass} ul`, {...}`

bearni95: already did, and doesn't work! I've found a work-around to also solve the issue @realdreamer pointed out:
```typescript
// should work, but doesn't
// globalStyle(`${ulListClass} li > ul`, {
globalStyle(`article ul > li >ul`, {
  paddingLeft: vars.sizes.$4,
  margin: 0,
  top: `calc(-1 * ${vars.sizes.$4})`,
  position: 'relative',
});

File: packages/apps/docs/src/components/Markdown/UnorderedList/styles.css.ts [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+  sprinkles({
+    marginY: '$5',
+    marginX: 0,
+    position: 'relative',
+  }),
+]);
+
+// should work, but doesn't
+// globalStyle(`${ulListClass} li > ul`, {
+globalStyle(`article ul > li >ul`, {
```
eileenmguo, line 15: this concerns me because it is only targeting native elements :/ Can you try `${ulListClass} > li > ul`?

packages/apps/docs/src/components/Markdown/UnorderedList/UnorderedList.tsx
```diff
@@ -1,11 +1,11 @@
-import { StyledList } from './styles';
+import { ulListClass } from './styles.css';
 
-import React, { FC } from 'react';
+import React, { FC, ReactNode } from 'react';
 
 interface IProp {
-  children: string;
+  children: ReactNode;
```
timoheddes, line 6: Instead of switching from string to any React Node you could make this something like:

```
FunctionComponentElement<HTMLOListElement | HTMLUListElement>[]
```

packages/apps/docs/src/components/Markdown/UnorderedList/styles.css.ts
realdreamer: @eileenmguo - updated! please review it again.

eileenmguo (approved):

PR #750, opened 2023-08-07
[client] Adds signWithEckoWallet and quicksignWithEckoWallet functions

ash-vd:
Adds two functions to sign a transaction with the [Ecko Wallet API](https://docs.ecko.finance/eckodex/getting-started/eckowallet/eckowallet-api).

Usage, where `unsignedTransaction` is a `IPactCommand`:
```
const signWithEckoWallet = createEckoWalletSign();
await signWithEckoWallet(unsignedTransaction);
```
```
const quicksignWithEckoWallet = createEckoWalletQuicksign();
await quicksignWithEckoWallet(unsignedTransaction);
```

On both `signWithEckoWallet` and `quicksignWithEckoWallet` there are several helpers available:
```
const isInstalled = signWithEckoWallet.isInstalled();
const isConnected = await signWithEckoWallet.isConnected();
const accountInfo = await getAccountInfo();
await signWithEckoWallet.connect();
```

`connect()` is also automatically called when signing for a transaction

### DIFF
--- a/common/changes/@kadena/client/feat-signWithEckoWallet_2023-08-08-11-29.json
+++ b/common/changes/@kadena/client/feat-signWithEckoWallet_2023-08-08-11-29.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/client",
+      "comment": "Adds signWithEckoWallet and quickSignWithEckoWallet",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@kadena/client"
+}
--- a/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
+++ b/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
@@ -5,7 +5,7 @@ import {
 
 export const isInstalled: ICommonEckoFunctions['isInstalled'] = () => {
   const { kadena } = window;
-  return Boolean(kadena && kadena.isKadena);
+  return Boolean(kadena && kadena.isKadena && kadena.request);
 };
 
 export const isConnected: ICommonEckoFunctions['isConnected'] = async (
@@ -21,11 +21,7 @@ export const isConnected: ICommonEckoFunctions['isConnected'] = async (
       networkId,
     });
 
-  if (checkStatusResponse?.status === 'fail') {
-    return false;
-  }
-
-  return true;
+  return checkStatusResponse?.status === 'success';
 };
 
 export const connect: ICommonEckoFunctions['connect'] = async (networkId) => {
--- a/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
+++ b/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
@@ -1,0 +1,60 @@
+import {
+  IEckoAccountsResponse,
+  IEckoConnectOrStatusResponse,
+} from './eckoTypes';
+
+export const isInstalled: ICommonEckoFunctions['isInstalled'] = () => {
+  const { kadena } = window;
+  return Boolean(kadena && kadena.isKadena);
+};
+
+export const isConnected: ICommonEckoFunctions['isConnected'] = async (
+  networkId,
+) => {
+  if (!isInstalled()) {
+    return false;
+  }
+
+  const checkStatusResponse =
+    await window.kadena?.request<IEckoConnectOrStatusResponse>({
+      method: 'kda_checkStatus',
+      networkId,
+    });
+
+  if (checkStatusResponse?.status === 'fail') {
+    return false;
+  }
+
+  return true;
+};
+
+export const connect: ICommonEckoFunctions['connect'] = async (networkId) => {
+  if (!isInstalled()) {
+    throw new Error('Ecko Wallet is not installed');
+  }
+
+  if (await isConnected(networkId)) {
+    return true;
+  }
+
+  const connectResponse =
+    await window.kadena?.request<IEckoConnectOrStatusResponse>({
+      method: 'kda_connect',
+      networkId,
+    });
+
+  if (connectResponse?.status === 'fail') {
+    throw new Error('User declined connection');
+  }
+
+  return true;
+};
+
+export const getAccountInfo = async (
+  networkId: string,
+): Promise<IEckoAccountsResponse | undefined> => {
+  return window.kadena?.request<IEckoAccountsResponse>({
+    method: 'kda_requestAccount',
+    networkId,
+  });
+};
--- a/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
+++ b/packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
@@ -1,5 +1,5 @@
 import {
-  IEckoAccountsResponse,
+  ICommonEckoFunctions,
   IEckoConnectOrStatusResponse,
 } from './eckoTypes';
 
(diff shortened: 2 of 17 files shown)


packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
```diff
@@ -0,0 +1,46 @@
...
+    await window.kadena?.request<IEckoConnectOrStatusResponse>({
+      method: 'kda_checkStatus',
+      networkId,
+    });
+
+  if (checkStatusResponse?.status === 'fail') {
+    return false;
+  }
+
+  return true;
```
Takadenoshi, line 23: Is this the best way to check for active status or should we return true on a positive signal?

packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (await isConnected(networkId)) {
+    return true;
+  }
+
+  const connectResponse =
+    await window.kadena?.request<IEckoConnectOrStatusResponse>({
+      method: 'kda_connect',
+      networkId,
+    });
```
Takadenoshi, line 39: Does this ever throw?

packages/libs/client/src/signing/eckoWallet/quicksignWithEckoWallet.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return isList ? transactions : transactions[0];
+  }) as IEckoSignFunction;
+
+  quicksignWithEckoWallet.isInstalled = isInstalled;
+  quicksignWithEckoWallet.isConnected = isConnected;
+  quicksignWithEckoWallet.connect = connect;
+
+  return quicksignWithEckoWallet;
```
EnoF, line 86: why do we set these props on the function?

Takadenoshi (approved):
LGTM aside from two small questions

packages/libs/client/src/signing/eckoWallet/quicksignWithEckoWallet.ts
ash-vd: So consumers can do something like:

```
const quicksignWithEckoWallet = createEckoWalletQuicksign();

await quicksignWithEckoWallet.connect();
```

to manually connect to Ecko. They're just exposed utility functions.

packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
ash-vd: Although Ecko returns a 'fail' status when it isn't connected, I agree that it might be better to turn things around to only return `true` when the returned status is `success`. Then it'll also return false when Ecko returns `success`.

packages/libs/client/src/signing/eckoWallet/eckoCommon.ts
ash-vd: Only when `request` doesn't exist on `window.kadena`, which is checked in `isConnected`. I've now added an extra check in that function to check if `window.kadena.request` is there.

EnoF (approved):

packages/libs/client/src/signing/eckoWallet/eckoTypes.ts
```diff
@@ -0,0 +1,73 @@
...
+  isConnected: typeof isConnected;
+  connect: typeof connect;
+}
+/**
+ * Interface to use when writing a signing function for Ecko Wallet that accepts a single transaction
+ * @public
+ */
+export interface IEckoSignSingleFunction
+  extends ISingleSignFunction,
+    ICommonEckoFunctions {}
```
javadkh2, line 25: mixing a function with properties might lead to a confusing api for users since function itself has some methods like apply, call ,... 
what if we instead add the sign as a method like this
```ts
export interface IEckoSignFunction {
   isInstalled: typeof isInstalled;
   isConnected: typeof isConnected;
   connect: typeof connect;
   sign : ISingleFunction,
   quickSign : ISingleFunction,
}

**Comment by javadkh2** [comment]
Maybe instead of exposing `createEckoWalletSign` and `createEckoWalletQuicksign` we can have one creator like `createEchoWalletConnector`
```ts

export interface IEckoWalletConnector {
   isInstalled: typeof isInstalled;
   isConnected: typeof isConnected;
   connect: typeof connect;
   sign : ISingleSignFunction,
   quickSign : ISignFunction,
}
const echoWalletConnector: IEckoWalletConnector = createEchoWalletConnector();

File: packages/libs/client/src/signing/eckoWallet/eckoTypes.ts [hunk 5]
```diff
@@ -0,0 +1,73 @@
...
+
+import { connect, isConnected, isInstalled } from './eckoCommon';
+
+export type EckoStatus = 'success' | 'fail';
+
+/**
+ * Interface that describes the common functions to be used with Ecko Wallet
+ * @public
+ */
+export interface ICommonEckoFunctions {
```
javadkh2, line 14: its better to keep the type file independent form the implementations so I would write this interface like this 
```
export interface ICommonEckoFunctions {
  isInstalled: () => boolean
  isConnected: (networkId:string) => Promise<boolean>
  connect: (networkId:string) => Promise<boolean>;
}

File: packages/libs/client/src/signing/eckoWallet/eckoTypes.ts (hunk 4 above)
Comment (alber70g) [reply packages/libs/client/src/signing/eckoWallet/eckoTypes.ts:25]: We've thought about that, but we wanted to make the usage of this function the same as the other functions to allow drop-in replacement.

**Comment by alber70g** [comment]
> Maybe instead of exposing `createEckoWalletSign` and `createEckoWalletQuicksign` we can have one creator like `createEchoWalletConnector`
> 
> ```ts
> export interface IEckoWalletConnector {
>    isInstalled: () => boolean;
>    isConnected: (networkId:string) => Promise<boolean>;
>    connect: (networkId:string) => Promise<boolean>;
>    sign : ISingleSignFunction,
>    quickSign : ISignFunction,
> }
> const echoWalletConnector: IEckoWalletConnector = createEchoWalletConnector();
> ```
> 
> What do you say @alber70g Maybe we even can have a separate package for `connectors`? like `@kadena/wallet-connectors` then we can focus of creating some connectors for common wallets there.

There are multiple things at hand here

1. as said in above comment, we wanted to make it as uniform as possible when working with the wallets and sign functions
2. we want to deprecate `sign` at some point. It's easier to deprecate that function when it has its own separate code-branch
3. A separate package would be nice. This would provide a clear entry point into all the external connections that can be made and provide an easy way to integrate new connections. Let's think about this, and create an issue for that

File: packages/libs/client/src/signing/eckoWallet/eckoTypes.ts (hunk 5 above)
Comment (alber70g) [reply packages/libs/client/src/signing/eckoWallet/eckoTypes.ts:12]: I modified the interface and changed `eckoCommon` to use the types of the functions from the interface

File: common/changes/@kadena/client/feat-signWithEckoWallet_2023-08-08-11-29.json [hunk 6]
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@kadena/client",
+      "comment": "Adds signWithEckoWallet and quickSignWithEckoWallet",
+      "type": "none"
```
alber70g, line 6: Shouldn't this be minor?

alber70g (approved):

common/changes/@kadena/client/feat-signWithEckoWallet_2023-08-08-11-29.json
alber70g: I updated this one

LindaOrtega:
Do we still want to merge this pull request?

PR #752, opened 2023-08-07
[docs] Search: add google events to new searches

bearni95:
- Added `search:qa` and `search:docs` to google-analytics events

---
  - https://app.asana.com/0/0/1205091936807444

### DIFF
--- a/packages/apps/docs/src/components/Search/Search.tsx
+++ b/packages/apps/docs/src/components/Search/Search.tsx
@@ -2,6 +2,7 @@ import { SearchResults } from './components/SearchResults';
 
 import { useSearch } from '@/hooks';
 import { useSemanticSearch } from '@/hooks/useSearch/useSemanticSearch';
+import { analyticsEvent, EVENT_NAMES } from '@/utils/analytics';
 import React, { FC, useEffect, useState } from 'react';
 
 interface IProps {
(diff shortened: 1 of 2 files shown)


eileenmguo (approved):

packages/apps/docs/src/components/Search/Search.tsx
```diff
@@ -35,6 +36,14 @@ export const Search: FC<IProps> = ({ query, hasScroll, limitResults }) => {
       if (tabName === 'docs') {
         handleSemanticSubmit(query);
       }
+
+      analyticsEvent(
+        EVENT_NAMES[tabName === 'qa' ? 'search:qa' : 'search:docs'],
+        {
+          label: query,
```
realdreamer, line 43:
```suggestion
          query: query,
```
i think `query` makes it more sense instead of label.

bearni95: Unfortunately these values are the parameter names in the method `analyticsEvent`, not sure it makes sense to change it there:
https://github.com/kadena-community/kadena.js/blob/b100166cb29fa26d5d194d17be0321260ea23c16/packages/apps/docs/src/utils/analytics.ts#L21

realdreamer (approved):

PR #759, opened 2023-08-08
Chore/client integration tests

Ghislain89:
This PR adds

- the `test:integration` command. Invoke through `pnpm run test:integration`
- GitHub Actions to run Integration tests  when a change is made to @kadena/Client
- Updates the default Jest unit test profile to ignore integration tests.
- Integration tests to validate integration between kadena client and chainweb.

---
  - https://app.asana.com/0/0/1205201634350538


packages/libs/client/src/integration-tests/test-data/accounts.ts
```diff
@@ -0,0 +1,14 @@
+import { IAccount } from '../support/interfaces';
+
+export const sourceAccount: IAccount = {
+  account: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
+  publicKey: '5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
+  chainId: '0',
+  guard: '5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
+};
+export const targetAccount: IAccount = {
```
javadkh2, line 9: Why is this one the same as sourceAccount?

javadkh2:
Nice job Ghislain!
I will try to fix the typing issue, then we won't need to cast as Pact.modules to any

packages/libs/client/src/integration-tests/test-data/accounts.ts
Ghislain89: I opted for the usecase that an individual wants to transfer funds to his own account on another chain. I can easily update if it's more realistic to be a different account?

javadkh2 (changes requested):
We need to update the github workflow file based on the new monorepo setup.

packages/libs/client/src/integration-tests/helpers/account/keyFromAccount.ts
```diff
@@ -0,0 +1,4 @@
+export type Account = `${'k' | 'w'}:${string}` | string;
+export function keyFromAccount(account: Account): string {
+  return account.split(':')[1];
+}
```
alber70g, line 4: Taking the key from the account can only be done with `k:...` accounts. Would be good to do a runtime check whether the `account` is compatible. Even when it is, the key could've been rotated. 

```suggestion
export function keyFromAccount(account: Account): string {
  if (!account.includes('k:')) {
    throw new Error(`Not able to retrieve key from '${account}'`);
  }
  return account.split(':')[1];
}
```

alber70g (approved):
Minor comments. Only static review

packages/libs/client/src/integration-tests/helpers/account/keyFromAccount.ts
Ghislain89: @alber70g given that we provide a type for Account: `export type Account = `${'k' | 'w'}:${string}` | string;`
Perhaps we could just return the account if there is no `k:` prefix.

packages/libs/client/package.json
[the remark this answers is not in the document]
Ghislain89: Should be done now.

packages/libs/client/src/integration-tests/helpers/account/fund-account.ts
```diff
@@ -0,0 +1,56 @@
...
+): Promise<string | undefined> {
+  const transaction = Pact.builder
+    .execution(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (Pact.modules as any).coin['transfer-create'](
+        sender00Account.account,
+        receiver,
+        readKeyset('ks'),
+        amount,
+      ),
```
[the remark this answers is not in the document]
Ghislain89: I tried making the typings available by running the first mentioned script (I had to slightly update it) this seems to generate typings but then using pact.modules.coin doesn't appear to work. I might need a tiny little bit of help getting that to work.

packages/libs/client/TESTING.md
```diff
@@ -0,0 +1,35 @@
...
+
+Once devnet is up and running execute the following script to execute the integration tests.
+
+``` shell
+pnpm run test:integration
+```
+[1]: https://www.docker.com/
+[2]: /packages/tools/heft-rig/jest.config.json
+[3]: http://localhost:8080
+
```
webpro, line 35: Bonus points if you add `*.md` or `TESTING.md` to the `format:md` script in `package.json` (and run that).

packages/libs/client/jest.integration.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
webpro, line 1: Does it make sense to extend from `packages/tools/heft-rig/jest.config.json` (and overwrite as needed like below), or not (yet)?

packages/libs/client/package.json
```diff
@@ -33,7 +33,8 @@
     "pactjs:generate:contract": "pactjs contract-generate --file contracts/coin.contract.pact",
     "pactjs:retrieve:contract": "pactjs retrieve-contract --out contracts/coin.contract.pact --module coin",
     "start": "ts-node --transpile-only src/index.ts",
-    "test": "jest"
+    "test": "jest",
+    "test:integration": "jest --silent -c ./jest.integration.config.js"
```
webpro, line 37: Why go silent?

packages/libs/client/src/integration-tests/helpers/fp-helpers.ts
```diff
@@ -0,0 +1,19 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const head = (args: all[]): any => args[0];
+
+export const inspect =
+  (tag: string): (<T extends unknown>(data: T) => T) =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  <T extends any>(data: T): T => {
+    console.log(tag, data);
+    return data;
+  };
```
webpro, line 19: Idea: move to reusable helpers? Not necessarily in this PR.

webpro (approved):
Nitpicks only, good stuff ✨

packages/libs/client/src/integration-tests/helpers/account/fund-account.ts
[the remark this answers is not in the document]
javadkh2: We can't generate types for the contract here. The problem is that this file is in the client package, and we use interface merging for Pact types. This will break types in the pact.ts file itself. If we can somehow isolate this section, maybe with a separate tsconfig file, we might be able to fix it.

javadkh2 (approved):
I think this is good enough and very valuable to merge it and have the test in the pipeline. Fixing the typing issues might require some revision of how we export contract types. For example, if we don't use interface merging, then we can easily generate types here.

packages/libs/client/package.json
Ghislain89: Primarily to not output all the console logging done by Kadena Client.  It makes the logs generated by Jest very hard to read.

packages/libs/client/src/integration-tests/helpers/fp-helpers.ts
Ghislain89: can do!

packages/libs/client/jest.integration.config.js
Ghislain89: I thought of that, but there are quite a few settings in the central config should not be used for the integration config. It might make sense to eventually add a shared integration config if more packages perform similar tests.

packages/libs/client/TESTING.md
Ghislain89: It is done!

PR #762, opened 2023-08-08
Introduce knip & remove some unused dependencies

webpro:
With a root `package.json` in this monorepo we unlock more opportunities and different tooling.

Knip scours the project for unused files, dependencies and exports. This PR contains just the tip of the iceberg.


timoheddes (approved):

sanderlooijenga (approved):

.changeset/quick-fireants-build.md
alber70g, line 1: Is this relevant for the external consumer of our lib? 
Sounds like this belongs in the commit message rather than the changelog

webpro: No, you're right. We need another level behind `patch` so we can have `none` too w/ Changesets. Updated.

package.json
```diff
@@ -10,6 +10,7 @@
     "format:ci": "pnpm -r run format:src --check",
     "format:md": "remark *.md -o --use @kadena-dev/markdown",
     "preinstall": "npx --no-install only-allow pnpm",
+    "knip": "NEXT_TRANSLATE_PATH=packages/apps/tools knip",
```
webpro, line 13:
```
NEXT_TRANSLATE_PATH=packages/apps/tools 
```

This has nothing to do with Knip, but some dependency uses `process.cwd()` instead of properly resolving the current Next.js directory and this is the only way I see now. Another could be to `pnpm patch` the package (and/or open PR upstream), will look into that later.

alber70g (approved):

sstraatemans (approved):

PR #764, opened 2023-08-08
[react-ui] Remove subcomponents from <Link>

bearni95:
- Removed the icon subcomponent
- Export simplified, only component and interface
- Corrected lint warnings on `Table.Stories.tsx` and `ContentHeader.tsx` files

---
  - https://app.asana.com/0/0/1205171349208298


eileenmguo (changes requested):
This should match the component API of `Button` so it should accept an icon as a prop. I'm not sure if you were part of the discussions where we decided to change this, but let me know if you need some more context!

packages/libs/react-ui/src/components/Link/Link.tsx
```diff
@@ -1,17 +1,38 @@
...
+      {Icon && iconAlign === 'left' && <LinkIcon icon={Icon} />}
+      {children}
+      {Icon && iconAlign === 'right' && <LinkIcon icon={Icon} />}
+    </>
+  );
+
   return (
     <a href={href} target={target} className={linkContainerClass}>
-      {children}
+      {buttonChildren}
```
eileenmguo, line 35:
```suggestion
      {linkChildren}
```

packages/libs/react-ui/src/components/Link/LinkIcon.tsx
eileenmguo, line 1: can we remove this file all together to simplify?

packages/libs/react-ui/src/components/Link/index.ts
```diff
@@ -1,17 +1,8 @@
 import type { ILinkProps } from './Link';
 import { Link as LinkRoot } from './Link';
```
eileenmguo, line 2: I think it's better to rename the files so there is no reference to Root anymore. Should be able to make this file much smaller as well

eileenmguo (commented):
When looking at chromatic, the last example of the link is missing the icon

eileenmguo (approved):

PR #765, opened 2023-08-08
[react-ui] ProfileCard component implementation

bearni95:
…sub-components; lint warning fixes

For now it uses props, as it's composition seems quite static. Open to structuring as subcomponents, but sounds a bit messier to implement @eileenmguo @realdreamer.

### DIFF
--- a/packages/libs/react-ui/src/components/ProfileCard/ProfileCard.tsx
+++ b/packages/libs/react-ui/src/components/ProfileCard/ProfileCard.tsx
@@ -1,0 +1,64 @@
+import {
+  boldTextClass,
+  imageClass,
+  imageContainerClass,
+  linkClass,
+  linkContainerClass,
+  profileCardClass,
+  tagClass,
+  tagContainerClass,
+} from './ProfileCard.css';
+
+import { Grid } from '@components/Grid';
+import { Tag } from '@components/Tag';
+import React, { FC } from 'react';
+
+export default {};
+export interface IProfileCardProps {
+  name: string;
+  title: string;
+  tags: string[];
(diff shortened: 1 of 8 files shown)


packages/libs/react-ui/src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Tag } from '@components/Tag';
+import React, { FC } from 'react';
+
+export default {};
+export interface IProfileCardProps {
+  name: string;
+  title: string;
+  tags: string[];
+  imageSrc: string;
+  links: Record<string, string>;
```
eileenmguo, line 22: I think using configuration for this is fine since it's quite simple and links will be external, however I do think tags and links should be optional

packages/libs/react-ui/src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Grid.Root columns={12}>
+        <Grid.Item columnSpan={2}>
+          <div className={imageContainerClass}>
+            <img className={imageClass} src={imageSrc} alt={name} />
+          </div>
+        </Grid.Item>
+        <Grid.Item columnSpan={10}>
+          <p className={boldTextClass}>{name}</p>
+          <p>{title}</p>
+          <ul className={tagContainerClass}>
```
eileenmguo, line 43: you should only render these lists if they have children, otherwise you're cluttering the dom with empty elements

eileenmguo (approved):

PR #766, opened 2023-08-09
implement select chain and account

rezaerami:
- add select chain dropdown
- add select account dropdown
- render existing chains and accounts based on wallet connect accounts

image
image
image
image


sanderlooijenga:
Should we rename `Select Chain` to `Select Network`?

packages/apps/tools/src/components/Common/Layout/index.tsx
```diff
@@ -31,13 +50,91 @@ export const Layout: FC<IProps> = ({ children }: IProps) => {
     },
   ];
 
+  const getNetworks = (): Network[] => {
```
sanderlooijenga, line 53: This logic is quite specific and belongs to the connect wallet functionality, right? If so; wouldn't it maybe be better to extract this logic to a better location?

packages/apps/tools/src/components/Common/Layout/index.tsx
```diff
@@ -31,13 +50,91 @@ export const Layout: FC<IProps> = ({ children }: IProps) => {
...
+          const [, network] = account.split(':');
+          return network;
+        })
+        ?.filter((network) => network !== 'development'),
+    );
+
+    return Array.from(result ?? []) as Network[];
+  };
+
+  const getAccounts = (): string[] => {
```
sanderlooijenga, line 66: See comment above

packages/apps/tools/src/context/connect-wallet-context.tsx
```diff
@@ -72,14 +109,15 @@ export const WalletConnectClientContextProvider: FC<
   );
 
   const connect = useCallback(
-    async (pairing?: { topic: string }) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async (pairing: any) => {
```
sanderlooijenga, line 113: There is no way of typing this parameter?

packages/apps/tools/src/components/Common/Layout/index.tsx
rezaerami: not really, here we just want to get accounts that belong to a chain, wallet connect gives you all of the accounts with network as its prefix, so we want to visually separate them not logically

packages/apps/tools/src/components/Common/Layout/index.tsx
rezaerami: same for accounts, this is only for separate accounts of a network visually

packages/apps/tools/src/context/connect-wallet-context.tsx
rezaerami: not really, tbh i tried everything i knew and everytime it was giving me a different error

MRVDH (approved):
Checked the functionality etc and looks good. Toggle works well. The background colour of the select box changes when switching the theme still though, but let's keep that out for now until we have added the new Header component.

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
```diff
@@ -1,41 +1,135 @@
...
+          if (account.match(regex)) return account;
+        })
+        ?.map((account) => {
+          const [, , accountName] = account.split(':');
+          return accountName;
+        }),
+    );
+
+    return Array.from(result) ?? [];
+  };
```
alber70g, line 75: - you can use `"".includes()` (no need to use regex for that, esp when you have to ignore a lint rule)
- you can return boolean in a filter lambda (no need to re-return when you're in a filter)
- you can pick the 3rd element straight from the split action
- A `Array.from(<SET>)` always returns an array, even if the set is empty

```suggestion
  const getAccounts = (): string[] => {
    return Array.from(new Set(
      accounts
        ?.filter((account) => account.includes(`kadena\:${selectedNetwork}`))
        ?.map((account) => account.split(':')[3]),
    ));
  };
```

alber70g (approved):
Minor comment

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
rezaerami: valid point, all fixed

timoheddes (approved):

PR #773, opened 2023-08-10
refactor: Footer to match NavHeader

timoheddes:
After implementing the NavHeader component we noticed the Footer doesn't entirely meet our standards. This PR fixes that and also renames the Footer and its files to NavFooter.


packages/libs/react-ui/src/components/index.ts
```diff
@@ -24,7 +24,7 @@ export type { ITooltipProps } from './Tooltip';
 export type { ITrackerCardProps, ILabelValue } from './TrackerCard/TrackerCard';
 export type { ITreeProps } from './Tree';
 export type {
-  INavFooterRootProps,
+  INavFooterContainerProps,
```
eileenmguo, line 27: I think Root makes more sense since the subcomponent is named Root. What do you think?

timoheddes: Yeah that works, I'll have NavHeader match it then in this PR.

packages/libs/react-ui/src/components/NavFooter/NavFooterLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+  children,
+  href,
+  target,
+}) => {
+  return (
+    <div className={linkBoxClass} data-testid="kda-footer-link-item">
+      <span
+        className={classNames(linkClass, {
+          color: 'inherit',
+          textDecorationColor: 'inherit',
```
eileenmguo, line 23: can these just be part of linkClass?

packages/libs/react-ui/src/components/NavFooter/NavFooterLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+        })}
+      >
+        {href !== undefined ? (
+          <a
+            style={{
+              color: 'inherit',
+              textDecorationColor: 'inherit',
+            }}
```
eileenmguo, line 31: can we move this to a class?

eileenmguo: Actually I think these styles can be removed

packages/libs/react-ui/src/components/NavFooter/NavFooterLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+            href={href}
+            target={target}
+          >
+            {children}
+          </a>
+        ) : (
+          <span>{children}</span>
```
eileenmguo, line 38: you can remove the extra wrapping span

packages/libs/react-ui/src/components/NavFooter/NavFooter.stories.tsx
```diff
@@ -102,14 +103,8 @@ export const Primary: Story = {
...
-              <NavFooter.Link key={index}>
-                {item.href !== undefined ? (
-                  <a href={item.href} target={item.target}>
-                    {item.title}
-                  </a>
-                ) : (
-                  <span>{item.title}</span>
-                )}
+              <NavFooter.Link key={index} href={item.href} target={item.target}>
+                {item.title}
```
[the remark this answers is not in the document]
timoheddes: Ah ok, this is just a story though. But I'll change it!

eileenmguo (approved):

PR #782, opened 2023-08-14
Update Select component

timoheddes:
### Select Component updates

[Asana Link](https://app.asana.com/0/1204958723576385/1205164815746147/f)

**What changed?**
- Updated Select styling to match design and support dark and light theme
- Tweaked the size of the native HTML select element that we render to ensure you can click anywhere on the actual Select component to see its options (previously required to click in the center of our visual component)
- Added disabled styling (missing exact design, [also missing in e.g. Input component](https://app.asana.com/0/1205288273408141/1205332525545630/f)?)
- Add variant prop to allow switching between default form select and one with a solid border and transparent background
- Set Tooltip z-index as a temporary fix to ensure it's above form elements (created [task](https://app.asana.com/0/1205288273408141/1205307366740459/f) to fix this for all components)
- Removed style from option elements as these can't be styled (?? not sure what the idea was there, am I missing something?)
- Added typing to the Input component type to match valid HTML Input Types rather than any string and updated Storybook implementation to allow switching between input types

**Implementation preview in Storybook, showing Input and Select:**

https://github.com/kadena-community/kadena.js/assets/79908211/75a157ed-c58f-4001-8095-0162d07a50d5

### DIFF
--- a/packages/libs/react-ui/src/components/Input/Input.stories.tsx
+++ b/packages/libs/react-ui/src/components/Input/Input.stories.tsx
@@ -4,11 +4,37 @@ import { IInputProps, Input } from '@components/Input';
 import { Stack } from '@components/Stack';
 import type { Meta, StoryObj } from '@storybook/react';
 import { vars } from '@theme/vars.css';
-import React from 'react';
+import React, { HTMLInputTypeAttribute } from 'react';
+
+const HTMLInputTypes: HTMLInputTypeAttribute[] = [
+  'button',
+  'checkbox',
+  'color',
+  'date',
+  'datetime-local',
+  'email',
+  'file',
+  'hidden',
+  'image',
+  'month',
+  'number',
+  'password',
+  'radio',
+  'range',
+  'reset',
+  'search',
+  'submit',
+  'tel',
+  'text',
+  'time',
+  'url',
+  'week',
+];
 
 const meta: Meta<
   {
     leftIcon: keyof typeof SystemIcon;
+    type?: React.HTMLInputTypeAttribute;
   } & IInputProps
 > = {
   title: 'Components/Input',
@@ -21,6 +47,10 @@ const meta: Meta<
     },
   },
   argTypes: {
+    type: {
+      control: { type: 'select' },
+      options: HTMLInputTypes,
+    },
     disabled: {
       description: 'Disables the input and applies visual styling.',
       control: {
@@ -35,7 +65,7 @@ const meta: Meta<
       description:
         'Icon rendered inside the input to the left of the input text.',
       options: [
-        undefined,
+        '-',
         ...(Object.keys(SystemIcon) as (keyof typeof SystemIcon)[]),
       ],
       control: {
@@ -46,7 +76,7 @@ const meta: Meta<
       description:
         'Icon rendered inside the input to the right of the input text.',
       options: [
-        undefined,
+        '-',
         ...(Object.keys(SystemIcon) as (keyof typeof SystemIcon)[]),
       ],
       control: {
@@ -58,6 +88,7 @@ const meta: Meta<
       control: {
         type: 'text',
       },
+      description: 'Text to display before the input',
     },
     leadingTextWidth: {
       description:
@@ -66,7 +97,7 @@ const meta: Meta<
         type: 'select',
       },
       options: [
-        undefined,
+        '- Omit this property to auto-size the leading text',
         ...Object.keys(vars.sizes).map((key) => key as keyof typeof vars.sizes),
       ],
     },
@@ -89,6 +120,7 @@ type Story = StoryObj<
     leadingText: string;
     leftIcon: keyof typeof SystemIcon;
     rightIcon: keyof typeof SystemIcon;
+    type: React.HTMLInputTypeAttribute;
   } & Omit<IInputProps, 'leftIcon' | 'rightIcon'>
 >;
 
@@ -96,8 +128,9 @@ export const Dynamic: Story = {
   name: 'Input',
   args: {
     leftIcon: undefined,
+    type: 'text',
     rightIcon: undefined,
-    leadingText: 'Leading',
+    leadingText: '',
     leadingTextWidth: undefined,
     outlined: false,
   },
@@ -109,6 +142,7 @@ export const Dynamic: Story = {
     leadingTextWidth,
     onChange,
     disabled,
+    type,
   }) => (
     <Input
       id="inlineInputStory"
@@ -120,6 +154,7 @@ export const Dynamic: Story = {
       leadingTextWidth={leadingTextWidth}
       outlined={outlined}
       disabled={disabled}
+      type={type}
     />
   ),
 };
@@ -128,19 +163,22 @@ export const InlineWithButton: Story = {
   name: 'Inline with button',
   args: {
     leftIcon: undefined,
+    type: 'text',
   },
-  render: ({ leftIcon, onChange }) => (
-    <Stack gap="$xs" alignItems="stretch">
+  render: ({ leftIcon, onChange, type }) => (
+    <Stack spacing="$xs" alignItems="stretch">
       <Input
         id="inlineInputStory"
         leftIcon={SystemIcon[leftIcon]}
         onChange={onChange}
         placeholder="This is a placeholder"
         outlined
+        type={type}
       />
       <Button title="Submit" onClick={() => {}}>
         Submit
       </Button>
+      <input type="" placeholder="This is a placeholder" />
     </Stack>
   ),
 };
--- a/packages/libs/react-ui/src/components/Input/Input.stories.tsx
+++ b/packages/libs/react-ui/src/components/Input/Input.stories.tsx
@@ -88,7 +88,6 @@ const meta: Meta<
       control: {
         type: 'text',
       },
-      description: 'Text to display before the input',
     },
     leadingTextWidth: {
       description:
--- a/packages/libs/react-ui/src/components/Input/Input.stories.tsx
+++ b/packages/libs/react-ui/src/components/Input/Input.stories.tsx
@@ -178,7 +178,6 @@ export const InlineWithButton: Story = {
       <Button title="Submit" onClick={() => {}}>
         Submit
       </Button>
-      <input type="" placeholder="This is a placeholder" />
     </Stack>
   ),
 };
--- a/packages/libs/react-ui/src/components/Input/Input.stories.tsx
+++ b/packages/libs/react-ui/src/components/Input/Input.stories.tsx
@@ -166,7 +166,7 @@ export const InlineWithButton: Story = {
     type: 'text',
   },
   render: ({ leftIcon, onChange, type }) => (
-    <Stack spacing="$xs" alignItems="stretch">
+    <Stack gap="$xs" alignItems="stretch">
       <Input
         id="inlineInputStory"
         leftIcon={SystemIcon[leftIcon]}
--- a/packages/libs/react-ui/src/components/Select/Select.css.ts
+++ b/packages/libs/react-ui/src/components/Select/Select.css.ts
@@ -11,7 +11,6 @@ export const containerClass = style([
     overflow: 'hidden',
     lineHeight: '$lg',
     borderRadius: '$sm',
-    flexDirection: 'column',
     borderColor: {
       lightMode: '$white',
       darkMode: '$gray60',
--- a/packages/libs/react-ui/src/components/Select/Select.css.ts
+++ b/packages/libs/react-ui/src/components/Select/Select.css.ts
@@ -10,7 +10,7 @@ export const containerClass = style([
     display: 'flex',
     overflow: 'hidden',
     lineHeight: '$lg',
-    borderRadius: '$sm',
+    borderRadius: '$xs',
     borderColor: {
       lightMode: '$white',
       darkMode: '$gray60',
@@ -46,7 +46,7 @@ export const selectVariants = styleVariants({
       backgroundColor: 'transparent',
     }),
     {
-      border: `1px solid ${vars.colors.$gray30}`,
+      border: `1px solid ${vars.colors.$gray40}`,
     },
   ],
 });
--- a/packages/libs/react-ui/src/components/Select/Select.css.ts
+++ b/packages/libs/react-ui/src/components/Select/Select.css.ts
@@ -1,20 +1,28 @@
 import { vars } from '../../styles';
 
 import { sprinkles } from '@theme/sprinkles.css';
-import { style, styleVariants } from '@vanilla-extract/css';
+import { style } from '@vanilla-extract/css';
 
 export const containerClass = style([
   sprinkles({
     alignItems: 'stretch',
-    color: '$foreground',
-    display: 'flex',
-    overflow: 'hidden',
-    lineHeight: '$lg',
+    backgroundColor: {
+      lightMode: '$white',
+      darkMode: '$background',
+    },
     borderColor: {
       lightMode: '$white',
       darkMode: '$gray60',
     },
+    borderRadius: '$sm',
+    color: '$foreground',
+    display: 'flex',
+    lineHeight: '$lg',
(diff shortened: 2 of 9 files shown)


packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -42,26 +47,29 @@ export const Select: FC<ISelectProps> = forwardRef<
...
-          aria-label={ariaLabel}
-          ref={ref}
-          className={selectClass}
-          disabled={Boolean(disabled)}
-          {...rest}
-        >
-          {children}
-        </select>
-      </div>
+      <label>
```
eileenmguo, line 55: why are you wrapping this in a label?

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -22,6 +25,7 @@ export interface ISelectProps
   onChange: React.ChangeEventHandler<HTMLSelectElement>;
   ref?: React.ForwardedRef<HTMLSelectElement>;
   ariaLabel: string;
+  variant?: (typeof variants)[number];
```
eileenmguo, line 28: You can remove "variants" and type it with just the variant object 
```suggestion
    variant?: keyof typeof selectVariants;
```

eileenmguo (commented):
Looking at storybook, the form version of the select is very close but doesn't look aligned with the input component. The idea is that these would be used together, so any visual inconsistencies will be very apparent. I would suggest for testing purposes to just put them on top of each other in storybook to compare styling during development

packages/libs/react-ui/src/components/Select/Select.tsx
timoheddes: It's just so that clicking the (optional) icon will also highlight the select field.

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -1,23 +1,56 @@
...
+  default: [
+    {
+      borderBottom: `1px solid ${vars.colors.$gray30}`,
+    },
+  ],
+  solid: [
+    {
+      border: `1px solid ${vars.colors.$gray30}`,
+    },
+  ],
```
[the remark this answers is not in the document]
timoheddes: I guess it's between use case vs what to expect visually. I think it's the default one that would be used in a form though, right? The one I named solid shows up in the design as part of the header, so default would become form and the solid one would become default? But that doesn't make too much sense either. I initially used default and form but found it confusing. What do you think? Perhaps I'm wrong thinking that the current default one would be the form one..

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -1,23 +1,56 @@
...
-  sprinkles({ pointerEvents: 'none' }),
-  { opacity: 0.5 },
+  sprinkles({
+    backgroundColor: {
+      lightMode: '$gray30',
+      // darkMode: '$gray100',
+    },
+    color: {
+      lightMode: '$gray100',
+      // darkMode: '$gray40',
```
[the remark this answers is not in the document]
timoheddes: No :)

packages/libs/react-ui/src/components/Select/Select.tsx
timoheddes: I updated it now to make the original wrapping div the label, so we don't have another depth added to it.

timoheddes:
> Looking at storybook, the form version of the select is very close but doesn't look aligned with the input component. The idea is that these would be used together, so any visual inconsistencies will be very apparent. I would suggest for testing purposes to just put them on top of each other in storybook to compare styling during development

I changed it to match input but I noticed that with the input component, its height jumps when an icon is added. I assume this is not the intention? Do you know? Otherwise I'll check with Isa. I can update the input component to remain the same height with or without the icon as well. Seems strange to me to have variation in the input heights in a form depending on whether or not an icon is added.

MRVDH:
We have select boxes in the Tools project header: https://www.figma.com/file/A11XXPwJSHNSqrRBEFfRc8/K%3AApps?type=design&node-id=725-11387&mode=design&t=wcBXEna6ES40ftxc-0
image

But since the regular form select is different, should this select in the header be a custom component? Is this something we should check with @isaozler?

cc: @KristinaSpasevska

timoheddes:
@MRVDH part of this PR is adding a variant, which would be the one you're linking to there :)

packages/libs/react-ui/src/components/Select/Select.css.ts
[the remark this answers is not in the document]
eileenmguo: hmmm I see where you're coming from. So we can set the default variant prop to anything regardless of what we name it. I just think the form variant should be named form for clarity and then we could name the other one something descriptive as well. I do think you're right that it will most likely be used in a form so that can be the default variant in the end

packages/libs/react-ui/src/components/Select/Select.tsx
eileenmguo: oh interesting, I didn't realize it worked like that. Do you know how that effects accessibility? Also we use our `InputWrapper` to add an additional label?

timoheddes: Ahh ok, I didn't realise that. Then it's a label wrapped in a label now and we don't want that 😁 I'll change it back in that case!

My change came from the comment from Maarten who said the "Click listener only listens on text." which he explained to mean that if he clicks outside of the select box, like on the icon, it won't trigger the select box. But if we wrap the whole thing in a label through the inputWrapper, that shouldn't be an issue once it's implemented in the form as the label should trigger the corresponding form item.

eileenmguo: No, there would not be a label wrapped around it. It would be adjacent to it and connected via htmlFor. So like 
```
<label>
<label/>
<label>
  <select/>
</label>
```

eileenmguo: Can you achieve this by just having the select span the entire container, adding extra padding on the left, and positioning the icon absolutely with no pointerevents?

timoheddes: If the for tag on the label matches the id of the input, it works the same as wrapping it around

packages/libs/react-ui/src/components/Select/Select.css.ts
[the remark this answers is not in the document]
timoheddes: So now the default = form and we have flat as alternative. I'll resolve this conversation but we'll get approval from all parties involved before merging :)

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -1,53 +1,87 @@
...
+export const selectVariants = styleVariants({
+  default: [
+    {
+      borderBottom: `1px solid ${vars.colors.$gray30}`,
+    },
+  ],
+  flat: [
+    {
+      border: `1px solid ${vars.colors.$gray30}`,
+      background: 'transparent !important',
```
timoheddes, line 49: Had to add important to overwrite the dark theme default background styling. Please let me know if there's another better way.

eileenmguo: You should be able to use sprinkles for this. I think it might work without the !important that way

timoheddes: I tried just now but it continues to overwrite it without adding important.

eileenmguo: Maybe it's because you are applying the backgroundColor in the containerClass which has more specificity. I would also suggest moving any variant specific styling to the variants styling section only only keeping shared styling in the containerClass. I feel like this may fix the issue

packages/libs/react-ui/src/components/Select/Select.stories.tsx
```diff
@@ -33,6 +51,7 @@ export default meta;
 type Story = StoryObj<
   {
     icon: keyof typeof SystemIcon;
+    variant: SelectVariants;
```
eileenmguo, line 54:
```suggestion
    variant: keyof typeof SelectVariants;
```

packages/libs/react-ui/src/components/Select/Select.stories.tsx
```diff
@@ -41,7 +60,7 @@ export const Dynamic: Story = {
   args: {
     icon: undefined,
   },
-  render: ({ icon, disabled }) => {
+  render: ({ icon, disabled, variant = variants[0] }) => {
```
eileenmguo, line 63:
```suggestion
  render: ({ icon, disabled, variant = "form" }) => {
```

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -4,12 +4,16 @@ import {
...
+  selectVariants,
 } from './Select.css';
 
 import { SystemIcon } from '@components/Icon';
 import classNames from 'classnames';
 import React, { FC, forwardRef } from 'react';
 
+export type SelectVariants = 'form' | 'flat';
+export const variants: SelectVariants[] = ['form', 'flat'];
+
```
eileenmguo, line 16: I think this can be removed

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -34,16 +39,16 @@ export const Select: FC<ISelectProps> = forwardRef<
     disabled = false,
     children,
     ariaLabel,
+    variant = variants[0],
```
eileenmguo, line 42:
```suggestion
    variant = "form",
```

timoheddes: Wait what why do you suggest this? It would mean that you have to change it multiple places if we decide to change the variant name. Having this default to the first option within the variants makes sense I think?

packages/libs/react-ui/src/components/Select/Select.tsx
timoheddes: I did that so there's one source of truth and I can use it with the `variants[0]`, so we only have to change it in one location if we ever do.

packages/libs/react-ui/src/components/Select/Select.css.ts
timoheddes: Ah yes, ok so I changed it to not have the background set on the containerClass and that works.

packages/libs/react-ui/src/components/Input/Input.stories.tsx
```diff
@@ -103,19 +138,22 @@ export const InlineWithButton: Story = {
...
         leftIcon={SystemIcon[leftIcon]}
         onChange={onChange}
         placeholder="This is a placeholder"
         outlined
+        type={type}
       />
       <Button title="Submit" onClick={() => {}}>
         Submit
       </Button>
+      <input type="" placeholder="This is a placeholder" />
```
eileenmguo, line 156: Did you mean to leave this?

packages/libs/react-ui/src/components/Select/Select.css.ts
```diff
@@ -1,53 +1,88 @@
...
-  sprinkles({ pointerEvents: 'none' }),
-  { opacity: 0.5 },
+  sprinkles({
+    backgroundColor: {
+      lightMode: '$gray20',
+    },
+    color: {
+      lightMode: '$foreground',
+    },
+  }),
```
eileenmguo, line 30: What was the reason for changing the disabled state?

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -34,16 +38,16 @@ export const Select: FC<ISelectProps> = forwardRef<
...
   },
   ref,
 ) {
   return (
     <div
-      className={classNames(
-        containerClass,
-        disabled ? containerClassDisabled : '',
-      )}
+      className={classNames(selectVariants[variant], containerClass, {
```
eileenmguo, line 48: nit: I would order the classNames in order of specificity. So since containerClass is the most generic, it should go first so that it can be overridden (I think) by the following classnames. This might be why you needed to remove the backgroundColor from containerClass

Just a note for the future - might help with styling!

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -4,12 +4,15 @@ import {
...
   selectClass,
   selectContainerClass,
+  selectVariants,
 } from './Select.css';
 
 import { SystemIcon } from '@components/Icon';
 import classNames from 'classnames';
 import React, { FC, forwardRef } from 'react';
 
+export type SelectVariants = 'form' | 'transparent';
```
eileenmguo, line 14: I'm not opposed to having a selectVariant type, but as discussed I think it should be based on the actual variant object rather than something that you define yourself. I believe this would be `keyof typeof selectVariants`

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -22,6 +25,7 @@ export interface ISelectProps
   onChange: React.ChangeEventHandler<HTMLSelectElement>;
   ref?: React.ForwardedRef<HTMLSelectElement>;
   ariaLabel: string;
+  variant?: SelectVariants;
```
eileenmguo, line 28: If you write it this way, you can remove the SelectVariants type
```suggestion
  variant?: keyof typeof selectVariants;
```

packages/libs/react-ui/src/components/Select/Select.css.ts
timoheddes: I made a [task](https://app.asana.com/0/1205288273408141/1205332525545630/f) on the board but there appears to be no disabled state implemented, even though it appears to exist in the design. If you look in Storybook, you see that the input component currently has no visual indication that it's disabled.

eileenmguo: Right. So right now the `Input` doesn't have a disabled style (which is a bug), however the `InputWrapper` and subsequently the `TextField` do (because of the `InputWrapper`) which I believe is similar to the original implementation for this disabled state. I would look there to decide how to style this disable state

Note: I saw that the input component styles look off now (leading text is on top of the actual input?)

timoheddes: What is this? I figured that's disabled. Though the design isn't too clear about it. In any case, wouldn't it make a lot more sense to have the disabled styling on the actual element rather than managed by a wrapper element? Or is that not what you're saying about the current situation?

image

Ah yeah I see the leading text is off now, I'll fix that.

eileenmguo: Hmm to be honest, I'm not sure what that state is. I'm saying when the wrapper component is disabled, it applies the disabled style to everything, but also the input needs to be updated to have a disabled state. @isaozler is back so let's have him take a look and we can decide if we want to break this into a different ticket?

eileenmguo: Actually I think it's fine to already break this into a separate ticket since the input component also needs updates

timoheddes: Yeah I think that would be best!

EnoF (approved):

eileenmguo (commented):
border radius and colors are off on the `transparent` variant. I would suggest adding it to the nav header to see if it looks right

image

packages/libs/react-ui/src/components/Select/Select.tsx
```diff
@@ -40,10 +40,9 @@ export const Select: FC<ISelectProps> = forwardRef<
...
-      className={classNames(
-        containerClass,
-        disabled ? containerClassDisabled : '',
-      )}
+      className={classNames(containerClass, {
+        [containerClassDisabled]: disabled,
+      })}
       data-testid="kda-select"
     >
       <div className={selectContainerClass}>
```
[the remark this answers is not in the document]
timoheddes: Yeah 👍

eileenmguo (approved):

PR #784, opened 2023-08-15
Refactor/change NavHeader component

timoheddes:
Through many changes to the implementation of NavHeader we ended up with a less than ideal version of it.

This PR will fix these issues and also some other things that were forgotten in the original implementation <sup>1</sup>

**Fixes**
- allow consumer to be flexible with the children of `<NavHeader.Navigation />` so that we can properly support NextJS Link elements and anything else.
- allow consumer to set which link is active on load (`activeLink` prop on `<NavHeader.Navigation />`).

**<sup>1</sup> other things this PR adds:**
- we forgot to export the NavHeader from the react-ui components index barrel file.
- support custom onclick handler for children of `<NavHeader.Navigation />`.

### DIFF
--- a/packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
+++ b/packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
@@ -114,7 +114,11 @@ export const Dynamic: IStory = {
       <NavHeader.Root brand={brand}>
         <NavHeader.Navigation activeLink={navHeaderActiveLink}>
           {navItems.slice(0, linksCount).map((item, index) => (
-            <NavHeader.Link key={index} href={item.href}>
+            <NavHeader.Link
+              key={index}
+              href={item.href}
+              onClick={(event) => console.log(item.label, { event })}
+            >
               {item.label}
             </NavHeader.Link>
           ))}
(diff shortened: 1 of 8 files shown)


packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
```diff
@@ -66,6 +67,14 @@ const meta: Meta<StoryProps> = {
       description: 'Adjust sample navigation items count',
       if: { arg: 'useCustomNavigation', neq: true },
     },
+    navHeaderActiveLink: {
+      control: { disable: true },
```
timoheddes, line 71: Disabled because it's strictly to indicate how you can use this prop.

rezaerami (approved):

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
```diff
@@ -88,21 +97,28 @@ export const Dynamic: IStory = {
...
     return (
       <NavHeader.Root brand={brand}>
-        <NavHeader.Navigation>
+        <NavHeader.Navigation activeLink={navHeaderActiveLink}>
           {navItems.slice(0, linksCount).map((item, index) => (
-            <NavHeader.Link key={index} href={item.href}>
+            <NavHeader.Link
+              key={index}
+              href={item.href}
+              onClick={(event) => console.log(item.label, { event })}
```
eileenmguo, line 120: Should we treat this link like our other Link component and accept all link related props that don't affect the props we are using?

eileenmguo (approved):

PR #786, opened 2023-08-16
Migrate Ava to jest, to use one tool for all packages.

Ghislain89:
This PR removes AVA for pact CLI and replaces it with the current standard Jest for unit tests. This should make it easier for developers across the company to work on various repositories while limiting the need to learn another API in order to use different tooling.


webpro (approved):
Agreed with @EnoF to approve/merge

PR #793, opened 2023-08-17
Refactor Accordion [Composition pattern, optimisations, contributing standards]

timoheddes:
⚠️ **_Please read before review_** ⚠️

There are a lot of changes. A lot of them are just related to making type imports and exports explicit, e.g. instead of:

`import React, { FC } from 'react';`

we do:

```
import React from 'react';
```

etc.

Not doing this causes warnings in your console. This change doesn't really require review. So, that leaves the primary refactor which is the `Accordion` component. In short, please focus on reviewing:
- Any changes _outside of_ react-ui library
- `Accordion` in react-ui

---

**General PR description**

[Asana task](https://app.asana.com/0/1205288273408141/1205217537838862/f)

This PR primarily aims to refactor Accordion to follow our standards regarding composition structure and other things. Other components like it will be updated in similar fashion in the future so that they all follow the same structure.

**Updated Accordion preview**

https://github.com/kadena-community/kadena.js/assets/79908211/462b365e-46d4-432d-81c2-fea656284675


packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -1,40 +1,96 @@
...
+                onOpen: section?.props.onOpen,
+                onClose: section?.props.onClose,
+              })
+            }
+            className={classNames(
+              accordionTitleClass,
+              accordionTitleVariants[
+                openSections.includes(sectionIndex) ? 'opened' : 'closed'
+              ],
+            )}
```
eileenmguo, line 73: Why is this logic/styling in `Accordion.Root` instead of `Accordion.Section`?

eileenmguo (commented):
Thanks for fixing the type warnings!! It's really good to get these taken care of :) In the future I would suggest creating a separate PR for them for the sake of separating concerns so we can merge changes as soon as their ready

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
timoheddes: I kept the logic in Root that decides what is rendered and how by iterating over the children. Open to other ideas though! My focus was really how do I refactor the Accordion that takes a `sections` prop, array of sections, to an Accordion that takes subcomponents to render its sections.

eileenmguo: As discussed in previous PRs, I think it makes sense to colocate any styles and logic with the subcomponent that they are actually associated with. This helps with maintainability because it makes the code clearer and more predictable. Just how on links we would want the link itself to actually apply styles, not the parent - similar to why VE doesn't allow you to apply styles to other elements

eileenmguo: What do you think?

timoheddes: Yeah that's true. I'll have another look at it!

packages/libs/react-ui/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -1,59 +1,80 @@
...
+  parameters: {
+    controls: {
+      hideNoControlsWarning: true,
+      sort: 'requiredFirst',
+    },
+    docs: {
+      description: {
+        component:
+          'Note: optional property of `openSection` may be set to the Root component to set the initially opened section. This defaults to -1 and has only been explcitly set in the story code for demonstration purposes.',
+      },
```
eileenmguo, line 31: why set to this -1 by default rather than undefined? Also, based on conventions this should be prefixed with `initial`. So `initialOpenSection` or something similar

packages/libs/react-ui/src/components/Accordion/AccordionSection.tsx
```diff
@@ -3,55 +3,53 @@
...
         className={classNames(
-          accordionTitleClass,
+          accordionSectionHeadingClass,
           accordionTitleVariants[isOpen ? 'opened' : 'closed'],
         )}
       >
-        <span>{title}</span>
+        <span
+          data-testid="kda-accordion-section-title"
+          onClick={onClick}
```
eileenmguo, line 47: I think the whole title + icon should be the button which should receive the onClick

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -1,43 +1,54 @@
...
+            onClick: () => {
+              if (openSections.includes(sectionIndex)) {
+                setOpenSections(openSections.filter((i) => i !== sectionIndex));
+                section?.props.onClose?.();
+              } else {
+                setOpenSections(
+                  linked ? [sectionIndex] : [...openSections, sectionIndex],
+                );
+                section?.props.onOpen?.();
+              }
```
eileenmguo: I already think this is a huge improvement! Some more comments, I think it's better to keep this as simple as possible so I have the following suggestions: 
- Move the `onOpen` and `onClose` function calls to the `Accordion.Section` component. The component should already be aware of whether it is open, so I think it makes sense to have an onClick handler inside of the component that handles function calls that are only associated with that component. I feel like this would be more clear since `Accordion.Root` has no information about the section props
- Nit: This isn't super necessary, but could be nice to make a higher-order function that returns the onClick handler for these sections to keep the jsx cleaner

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -1,43 +1,53 @@
...
+          section as React.ReactElement<
+            HTMLElement | IAccordionSectionProps,
+            | string
+            | React.JSXElementConstructor<JSX.Element & IAccordionSectionProps>
+          >,
+          {
+            index: sectionIndex,
+            isOpen: openSections.includes(sectionIndex),
+            onOpen: () => section?.props.onOpen?.(),
+            onClose: () => section?.props.onClose?.(),
```
eileenmguo, line 39: Do you need to pass these?

packages/libs/react-ui/src/components/Accordion/AccordionSection.tsx
```diff
@@ -3,55 +3,61 @@
...
-
+    onClick?.();
+  };
   return (
-    <div className={accordionSectionClass}>
+    <section
+      className={accordionSectionClass}
+      data-testid="kda-accordion-section"
+    >
       <div
```
eileenmguo, line 48: should this be a button? and then you can wrap the icon in a span

timoheddes: The plus is already a button so then it would be a button in a button. But I'm not sure what would be semantically ideal..

eileenmguo: Well right now the icon is a button, the onClick handler is placed on the div wrapping the text so this can be semantically improved. Spoke with Isa and the whole text + icon should trigger the interaction so I think the html needs to be reworked a bit

timoheddes: Oh, currently the whole text + icon already triggers the interaction so that's not an issue but I agree we can improve the html structure. I'll do that :)

timoheddes: How about we turn the sections into [articles](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) and have the title and icon wrapped in a <header>? I'll push that change so you can see.

webpro:
Just saying, this is valid too:

```ts
import React, { type FC } from 'react';
```

PR #797, opened 2023-08-18
chore(monorepo): update README and CONTRIBUTING

alber70g:
After several questions from the community, it would be a good idea
to have the getting started in the README itself, instead of hidden
in the CONTRIBUTION guide.

After comments on issue resolution in the monorepo, it is reasonable
to have a basic guide on how to fix problems. Relevant steps have
been added to the CONTRIBUTION guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,12 +37,15 @@ tutorials" such as:
 
 ### Switch branches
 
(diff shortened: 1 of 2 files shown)


alber70g:
If we ever go back to Rush we need this back from Turborepo

PR #805, opened 2023-08-21
feat(pactjs-cli): let users combine --file and --contract

javadkh2:
the pactParser supports combining files and contracts name while generating types so I removed the restriction in the cli tool


common/changes/@kadena/pactjs-cli/feat-use-file-contract-together_2023-08-23-08-13.json
alber70g, line 1: Needs to be updated to the new changesets

alber70g (approved):

PR #806, opened 2023-08-21
Feat/kadena tools/sidebar account transactions

MRVDH:
Code for this page is not done yet. Midway PR.


packages/apps/tools/src/pages/account/account-transactions/results/index.tsx
```diff
@@ -121,8 +133,71 @@ const CheckTransactions: FC = () => {
...
+                      isAccount: true,
+                    },
+                    {
+                      label: 'to chain',
+                      value:
+                        transactionDetails?.crossChainId ||
+                        transactionDetails?.chain,
+                    },
+                  ]}
+                  icon={ProductIcon.ReceiverInactive}
```
nil-amrutlal-dept, line 194: The icon should change according to the transfer status. If the transfer is complete then it should show the `Receiver` icon instead of `ReceiverInactive`

packages/apps/tools/src/pages/account/account-transactions/results/index.tsx
```diff
@@ -3,15 +3,23 @@ import {
   Box,
   Breadcrumbs,
   Button,
+  Card,
```
nil-amrutlal-dept, line 6: Can be removed?

packages/apps/tools/src/pages/account/account-transactions/results/index.tsx
nil-amrutlal-dept: Oh ok I see. Then maybe just replace the Icon with `ProductIcon.Receiver`

nil-amrutlal-dept (approved):

packages/apps/tools/src/pages/account/account-transactions/results/index.tsx
MRVDH: Fixed!

KristinaSpasevska (approved):

PR #809, opened 2023-08-22
[docs] fix: analytics events

sstraatemans:


packages/apps/docs/src/components/Search/Search.tsx
```diff
@@ -28,7 +28,12 @@ export const Search: FC<IProps> = ({ query, hasScroll, limitResults }) => {
   } = useSemanticSearch(limitResults);
 
   useEffect(() => {
-    if (query !== undefined && query.trim() !== '') {
+    if (
+      query !== undefined &&
+      query.trim() !== '' &&
+      tabName !== undefined &&
```
sstraatemans, line 34: this makes sure that the event is only triggered 1 time, because the tabname is set in a useEffect

timoheddes (approved):

PR #811, opened 2023-08-22
Feat/add storybook setup

realdreamer:


sstraatemans (approved):

.github/workflows/docs-chromatic.yml
webpro, line 1: I think it makes sense to rename the files:

- `chromatic-docs.yml`
- `chromatic-react-ui.yml`
- etc.

webpro (approved):

PR #813, opened 2023-08-23
[@react-ui & implementation] Change 'spacing' prop to 'gap'

KristinaSpasevska:


timoheddes (approved):

MRVDH (approved):
I wonder if it's necessary to change it in react-components as well since it's legacy, but either way approved 😄

PR #814, opened 2023-08-23
No unused vars as errors

webpro:
As I think unused variables are not just warnings, but errors.


alber70g:
Does this now block the local build as well? Or is there a way to ignore that locally? It would be extremely annoying to be blocked by this during development. 

Let's say you want to try something out. You comment out a block of code, replace it with a `console.log`, and build and run it. Now you're faced with a blocking error that a variable is unused. That's undesirable

webpro:
> Does this now block the local build as well? Or is there a way to ignore that locally? It would be extremely annoying to be blocked by this during development.

This is why linting and building should be separate. Then you can still build. But I see we're not there yet. So let's wait with merging this one until #612 is merged first.

sstraatemans (approved):

timoheddes (approved):

.changeset/gorgeous-ducks-explode.md
alber70g, line 1: When we merge this file like this. Will all subsequent changesets not trigger a "missing" changeset, and inadvertently result in missing changeset entries when we release?

alber70g (commented):
One question about the changeset file

alber70g (approved):

PR #815, opened 2023-08-23
[@react-ui] Add initialSelectedPage to Pagination Component

KristinaSpasevska:

### DIFF
--- a/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
+++ b/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
@@ -18,7 +18,7 @@ const meta: Meta<
     docs: {
       description: {
         component:
-          '<i>Note: maximum navigation items is currently limited (not technically enforced).</i><br><br><strong>Label </strong><br> is used to allow screen readers to notify user that there is a pagination here.',
+          'A component used for pagination.',
       },
     },
   },
--- a/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
+++ b/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
@@ -17,8 +17,7 @@ const meta: Meta<
   parameters: {
     docs: {
       description: {
-        component:
-          'A component used for pagination.',
+        component: 'A component used for pagination.',
       },
     },
   },
--- a/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
+++ b/packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
@@ -27,7 +27,7 @@ const meta: Meta<
       control: {
         type: 'text',
       },
-      description: 'We use this to give info to screen readers',
+      description: 'Text that is passed to the Nav element as an aria-label for accessibility.',
     },
     totalPages: {
       control: {
@@ -36,7 +36,7 @@ const meta: Meta<
         max: 20,
         step: 2,
       },
-      description: 'Total number of pages',
+      description: 'Total number of pages.',
     },
     visiblePageLimit: {
       control: {
@@ -45,7 +45,7 @@ const meta: Meta<
         max: 7,
         step: 1,
       },
-      description: 'Number of pages that would be visible',
+      description: 'Number of pages that are visible and can be directly selected.',
     },
     initialSelectedPage: {
       control: {
@@ -54,7 +54,7 @@ const meta: Meta<
         max: 20,
         step: 1,
       },
-      description: 'Which page should be selected at start?',
+      description: 'The default selected page before any interaction.',
     },
   },
 };
@@ -76,13 +76,7 @@ export const Controlled: Story = {
     initialSelectedPage: 2,
   },
   render: ({ totalPages, label, visiblePageLimit, initialSelectedPage }) => {
-    const validInitialSelectedPage =
-      initialSelectedPage &&
-      initialSelectedPage <= totalPages &&
-      initialSelectedPage > 0;
-    const [page, setPage] = React.useState(
-      validInitialSelectedPage ? initialSelectedPage : 1,
-    );
+    const [page, setPage] = React.useState(initialSelectedPage ?? 1);
 
     return (
       <Stack direction="column" gap="$4">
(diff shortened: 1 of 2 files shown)


packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -65,6 +74,7 @@ export const Controlled: Story = {
           currentPage={page}
           label={label}
           visiblePageLimit={visiblePageLimit}
+          initialSetPage={initialSetPage}
```
eileenmguo, line 77: nit: Do you think `initialPage` would be more concise and still clear?

KristinaSpasevska: Should have been _initialSelectedPage_ , but _initialPage_ is fine too

eileenmguo: initialSelectedPage is fine with me as well! `initialSetPage` just felt a bit strange, but I'm ok with either of the other options 👍

KristinaSpasevska: Set to _initialSelectedPage_

eileenmguo (commented):
Looking at storybook. It doesn't have any selected page if the initialSelectedPage is not valid. Can you update it to set it to the first page if this is the case or implement another solution? 

Also unrelated to your changes, but I realized I forgot to wrap the `PageNum` component in an `li`. Could you add that in this PR?

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -53,9 +61,14 @@ export const Controlled: Story = {
     label: 'Label',
```
timoheddes, line 61: Another improvement that could be made, which isn't directly part of this PR, is to add a description to this prop/argument in storybook as it's not directly clear where or how this label is used.

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -53,9 +61,14 @@ export const Controlled: Story = {
...
     totalPages: 10,
     visiblePageLimit: 3,
+    initialSelectedPage: 2,
   },
-  render: ({ totalPages, label, visiblePageLimit }) => {
-    const [page, setPage] = React.useState(1);
+  render: ({ totalPages, label, visiblePageLimit, initialSelectedPage }) => {
+    const validInitialSelectedPage =
+      initialSelectedPage && initialSelectedPage <= totalPages;
```
eileenmguo, line 68: should also be greater than 0

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -50,12 +70,19 @@ type Story = StoryObj<IPaginationProps>;
...
-  render: ({ totalPages, label, visiblePageLimit }) => {
-    const [page, setPage] = React.useState(1);
+  render: ({ totalPages, label, visiblePageLimit, initialSelectedPage }) => {
+    const validInitialSelectedPage =
+      initialSelectedPage &&
+      initialSelectedPage <= totalPages &&
+      initialSelectedPage > 0;
+    const [page, setPage] = React.useState(
+      validInitialSelectedPage ? initialSelectedPage : 1,
+    );
```
eileenmguo, line 85: Can this logic be removed from the story? I think the component should handle it internally

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -35,6 +45,16 @@ const meta: Meta<
...
+      description: 'Number of pages that would be visible',
+    },
+    initialSelectedPage: {
+      control: {
+        type: 'number',
+        min: 1,
+        max: 20,
+        step: 1,
+      },
+      description: 'Which page should be selected at start?',
```
eileenmguo, line 57:
```suggestion
      description: 'The default selected page before any interaction.',
```

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -14,11 +14,20 @@ const meta: Meta<
...
+          '<i>Note: maximum navigation items is currently limited (not technically enforced).</i><br><br><strong>Label </strong><br> is used to allow screen readers to notify user that there is a pagination here.',
+      },
+    },
+  },
   argTypes: {
     label: {
       control: {
         type: 'text',
       },
+      description: 'We use this to give info to screen readers',
```
eileenmguo, line 30:
```suggestion
      description: 'Text that is passed to the Nav element as an aria-label for accessibility.',
```

packages/libs/react-ui/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -14,11 +14,20 @@ const meta: Meta<
   } & IPaginationProps
 > = {
   title: 'Components/Pagination',
+  parameters: {
+    docs: {
+      description: {
+        component:
+          '<i>Note: maximum navigation items is currently limited (not technically enforced).</i><br><br><strong>Label </strong><br> is used to allow screen readers to notify user that there is a pagination here.',
```
eileenmguo, line 21: What do you mean by this? `<i>Note: maximum navigation items is currently limited (not technically enforced)`

KristinaSpasevska: I took that from the NavHeader, but I thought it would be nice to notify the user that the limit we put in the storybook for the total number of pages is only for the usage of the component here (it's not a technical limitation)
Screenshot 2023-08-24 at 16 33 55

eileenmguo: The label description is duplicated in the props section. I do think that this text overall is more confusing than it is helpful right now. I would suggest just writing a a plain text description for what the component does/is used for

eileenmguo (approved):

PR #825, opened 2023-08-28
feat(docs): implement storybook for layout components

realdreamer:
Closes : https://app.asana.com/0/1205288273408141/1205310320676421/f

This PR adds the storybook for the following layout types
- Blog
- Landing
- Redocly
- Full

As part of this story figured out that having storybook for Home component doesn't makes sense. Since, as part of `Home` everything is just a `children`


packages/apps/docs/package.json
```diff
@@ -15,7 +15,7 @@
     "build:importReadMe": "node ./src/scripts/importReadme.mjs && prettier ./src/pages/docs --write",
     "build:next": "pnpm run build:scripts && next build",
     "build:scripts": "pnpm run build:importReadMe && pnpm run build:createtree && pnpm run build:createSpecJson && pnpm run build:brokenLinks && pnpm run build:checkForHeaders",
-    "build:storybook": "storybook build",
+    "build:storybook": "pnpm run build:createtree && storybook build",
```
sstraatemans, line 18: this will already run when the normal build will run.
Normal build is necessary in the monorepo to kick off dependency builds like `react-ui`

packages/apps/docs/src/components/Layout/Blog/styles.ts
```diff
@@ -1,6 +1,6 @@
 import { styled, StyledComponent } from '@kadena/react-components';
 
-import { BasePageGrid } from '../components';
```
sstraatemans, line 3: layout looks weird on small screen.
different as in browser

image

packages/apps/docs/src/components/Layout/Full/index.stories.tsx
```diff
@@ -0,0 +1,89 @@
+import { fullLayoutFrontMatter } from '../__fixtures__/frontmatter';
```
sstraatemans, line 1: layout onsmall screen is different as in browser.
header seems to have disapeared
image


in my browser:

image

packages/apps/docs/src/components/Layout/Blog/styles.ts
realdreamer: As we discussed IRL, https://app.asana.com/0/1205288273408141/1205384296981150/f created a separate bug ticket for fixing this

packages/apps/docs/src/components/Layout/Full/index.stories.tsx
realdreamer: As I mentioned, as part of storybook didn't implement complete page! this storybook contains only the layout part. So, it will not have header component.

packages/apps/docs/package.json
realdreamer: with turbo repo we can build all dependencies separately without running complete build. So, that's what we're doing as part of chromatic ci. So, we need this script separately.

sstraatemans (approved):

webpro (approved):
LGTM!

PR #832, opened 2023-08-29
chore(build): use turbo to execute all ci tasks

alber70g:
Turbo knows what tasks depend on each other. Using this
we can optimize all tasks that we need in ci and execute
them as one task


webpro (approved):
🚀

PR #834, opened 2023-08-29
chore(@kadena/react-ui): Updated the Link component to work with external links

eileenmguo:
There is been some confusion about how to use react-ui components with Next/Link. Previously many different solutions have shown up across the components, but this PR is the result of an effort to create one convention for this issue. 

This approach allows users to set an `asChild` prop on the Link component and add the external component as a direct child. When this prop is set, all props from the Link component are passed down to the child along with styles and icons. 

It can be used like so: 
```
<UILink icon="Account" asChild>
  <NextLink href="...">
    Link Label
  </NextLink>
</UILink>
```

The benefit of this is that the semantic html ends up with only one anchor and it keeps the styles and logic within the ui library's Link component

Caveats: 
- href prop on the UI's Link component is now optional so that users can set an href on the child. This means that users can potentially use the Link component without an href and there won't be a typescript error. I think this is still better than having to require the user to set an href on the `Link` component and the external component
- In the case of the same props being set on both the `Link` component and the external component, the external component props will override the ones applied to the `Link` component

The idea is that we update all link components to allow for this. Curious to know what you think!

UPDATE
It looked like no apps were using the `Link` component, so I updated the `Breadcrumb` component as well to check how it would look in other applications. I went ahead and removed the string interpolation on the href from the base react-ui `Breadcrumb` component to the docs and updated the implementation with this new approach. The result is cleaner html!


timoheddes:
I feel the best way to really review this and see if it suits our needs the way we hope, is through actually field testing it and implementing it. Should we update components that currently support Next Links so that they use this? I'm guessing that would be the idea anyway and then we can see if we run into other issues or not?

sstraatemans (approved):

timoheddes (approved):

webpro (approved):

PR #835, opened 2023-08-29
fix(@kadena/react-ui): Update Select props to align with Input props and tested with external lib

eileenmguo:
When the Select component was being used in the dev wallet, it wasn't working with react-hook-form. This PR adjusts props a little to require an `id` like the input and be compatible with react-hook-form. Tested this in the dev-wallet


timoheddes:
Perhaps `id` should be an optional prop on all components?

timoheddes (approved):

PR #837, opened 2023-08-29
fix(@kadena/react-ui): Refactored icon props to be consistent

eileenmguo:
Some components were accepting an Icon component as the icon prop and other were accepting a string for the key. I've updated all components to use the latter approach as it doesn't require the user to import the icon component.


webpro:
Not really tied to this PR, but using a string here also allows to start using the SVG sprite approach, I see we're still importing **all** the JSX now. Pages are close to 300KB, maybe good to see what we can do about this? Not in this PR obv. :)

sstraatemans (approved):

webpro:
I remember we've discussed changelogs briefly. Even though this lib is not published to npm, using Changesets allows to maintain a changelog. It will end up as `CHANGELOG.md` in each package dir. Or does this PR not contain a changelog item interesting enough to add it there?

Just in case, this is the syntax:

```
---
"@kadena/pkg1": none
"@kadena/pkg2": none
---

Description
```

webpro (approved):

PR #840, opened 2023-08-30
Shed some light on pnpm + turbo --filters

webpro:
The `--filter` argument can be confusing, this doc tries to clarify a bit. Still, the docs of pnpm and Turborepo are leading!

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -25,9 +25,14 @@ pnpm test --filter @kadena/client
 pnpm run --filter @kadena/client test
 ```
 
+There are also ways to include dependencies, select up to and/or including a
+certain package, packages changed since the `main` branch, etc. Refer to the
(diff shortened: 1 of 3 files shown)


docs/filters.md
```diff
@@ -0,0 +1,53 @@
...
+
+To run the test script of a specific package(s), bypassing Turborepo entirely:
+
+```sh
+pnpm --filter <package_selector> <command>
+```
+
+These commands all do the same, the `--filter` argument is provided to `pnpm`:
+
+```sh
```
eileenmguo, line 21: might be worth adding some documentation around the ellipses to select the package and it's dependencies https://pnpm.io/filtering#--filter-package_name-1

webpro: Good point. I'll add a note, but _really_ trying not to copy existing docs :)

PR #841, opened 2023-08-30
[DOCS]chore: small style changes in search

sstraatemans:


packages/apps/docs/src/utils/createSlug.ts
```diff
@@ -1,11 +1,20 @@
...
+  // if the length of the string is 0, return 0
+  if (str.length === 0) return `${hash}`;
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  str.split('').forEach((_char, idx) => {
+    const ch = str.charCodeAt(idx);
+    hash += ch;
+  });
+  return `h${hash}`;
+};
```
realdreamer, line 12: this change will introduce back the same issue we had before in the aside menus. https://alpha-docs-git-feat-docsimprovesearch2-kadena-js.vercel.app/docs/pact/reference/functions/operators#h94

Operators `!=` and `^` both shares the same hash now.
`and` `and?` shares the same hash etc..

realdreamer (approved):

timoheddes:
Side-note: I noticed that docs still has implementations of components that use an icon which is added as a string. This is no longer correct since [this change](https://github.com/kadena-community/kadena.js/pull/837). 

Example is `SearchResults.tsx` on line 147.

PR #844, opened 2023-08-31
Refactor Accordion component

timoheddes:
This PR updates the Accordion to meet our component standards and guidelines. It's a complete refactor as the previous implementation of the Accordion didn't match our standards nor matched the design.
 
**What changed?**
- The component is now composable with `Accordion.Root` as root and `Accordion.Section` as its children.
- The component has been updated to match the design.
- The component and its children now have a unidirectional data flow and logic and are properly typed.
- Consumers can now set the initially opened section through a new prop on the Root.
- The Sections no longer depend on/require useRef and useEffect.
- The story has been refactored with descriptive arguments and controls.

Finally, the implementation of the Accordion within the tools package has been updated to meet the new composition structure.

_Note that the Accordion has two versions. This is the content version that can be placed anywhere on a page. Another version will be made for navigation purposes._

[Asana task](https://app.asana.com/0/1205288273408141/1205217537838862/f)


.changeset/late-files-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+"@kadena/tools": patch
+"@kadena/react-ui": patch
+---
+
+Refactor Accordion component
```
eileenmguo, line 6: In this case I wouldn't add a changeset to @kadena/tools as (i think) there is not something substantial that changed in the tools site. For @kadena/react-ui I would also update it to `none` since it isn't published, but we can also have a discussion about it if you disagree.

Also would make the description of the change more descriptive as it is meant to help users understand what has changed

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
```diff
@@ -1,43 +1,53 @@
...
-          {section.children}
-        </AccordionSection>
-      ))}
+    <div data-testid="kda-accordion-sections">
+      {React.Children.map(children, (section, sectionIndex) =>
+        React.cloneElement(
+          section as React.ReactElement<
+            HTMLElement | IAccordionSectionProps,
+            | string
+            | React.JSXElementConstructor<JSX.Element & IAccordionSectionProps>
```
eileenmguo, line 35: Hmm I feel like you need to check if the `section` is a valid element or I think you'll get an error cloning it. For example what happens if it is a string?

eileenmguo (commented):
I feel like this is a great improvement on this component :)

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
timoheddes: Children are typed though, TS should show an error if you pass a string?

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -13,32 +16,49 @@ export const accordionSectionClass = style([
     },
   },
 ]);
-export const accordionTitleClass = style([
-  sprinkles({
-    display: 'flex',
+
+export const accordionSectionHeadingClass = style([
```
eileenmguo, line 20: Could some of these classes be consolidated into one now that they are all part of the same button?

packages/libs/react-ui/src/components/Accordion/AccordionSection.tsx
```diff
@@ -3,67 +3,74 @@
...
+          data-testid="kda-accordion-section-title"
+          className={accordionTitleClass}
         >
-          <SystemIcon.Close size="sm" />
-        </button>
-      </div>
+          {title}
+        </span>
+
+        <span className={toggleButtonClass}>
```
eileenmguo, line 61: Can these styles be cleaned up? Maybe even combined into the same span for cleaner css? 

nit: This is no longer a button so the classname could be updated

packages/libs/react-ui/src/components/Accordion/Accordion.tsx
eileenmguo: Well I think currently this is typed to allow a string? Maybe I'm wrong, just looks like it could break when passing in an element that is not an accordion section.

timoheddes: The effective type comes from the children as defined in the `IAccordionRootProps` but I understand the confusion as this part isn't necessary anymore. I'll change it to correctly reflect the expected type.

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -1,8 +1,11 @@
 import { sprinkles } from '@theme/sprinkles.css';
-import { style, styleVariants } from '@vanilla-extract/css';
+import { vars } from '@theme/vars.css';
+import { style } from '@vanilla-extract/css';
 
 export const accordionSectionClass = style([
   sprinkles({
+    display: 'block',
+    flexGrow: 1,
```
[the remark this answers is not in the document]
timoheddes: Ah yeah, forgotten left over 😁

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -13,32 +16,49 @@ export const accordionSectionClass = style([
...
     alignItems: 'center',
-    justifyContent: 'space-between',
+    background: 'none',
+    border: 'none',
+    borderBottom: `1px solid ${vars.colors.$gray20}`,
     cursor: 'pointer',
+    display: 'flex',
+    padding: 0,
+    textAlign: 'left',
+    transition: 'color 0.2s ease 0s',
```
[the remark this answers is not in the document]
timoheddes: I think that's the default when you omit it so it would end up being the same.

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -13,34 +15,50 @@ export const accordionSectionClass = style([
...
+    display: 'flex',
     fontSize: '$base',
     fontWeight: '$medium',
+    padding: 0,
+    textAlign: 'left',
+    transition: 'color 0.2s ease',
+    width: '100%',
+    selectors: {
+      '&.isOpen': {
+        border: 'none',
```
eileenmguo, line 35: can you use sprinkles?

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -13,34 +15,50 @@ export const accordionSectionClass = style([
...
-    borderBottom: '1px solid',
-  },
 ]);
-export const accordionTitleVariants = styleVariants({
-  closed: [sprinkles({ color: '$negativeContrast' })],
-  opened: [sprinkles({ color: '$foreground' })],
-});
 
-export const toggleButtonClass = style([
+export const toggleIconWrapperClass = style([
```
eileenmguo, line 50: I think you can combine the accordionTitleClass and toggleIconWrapperClass. You may even be able to combine it with the headingClass. I would suggest looking at ways to simplify the html and css to reduce the amount of code

eileenmguo (commented):
Some visual updates still need to be made:
- text should be bolder I think
- Icon looks a bit big
- and Icon doesn't show up in dark mode (I think simplifying html and css will fix this)
image

eileenmguo (commented):
It looks like the x icon changes colors when it is rotated, the line is the wrong color in dark mode, and the line is in the wrong place. It should be after the content

timoheddes:
@eileenmguo Can't really react to comments directly but the computed color for the x remains the same for me, did you check that and if so what color change does it show you?

Ah I misunderstood the line position change, I moved it now and updated its color :)

eileenmguo (approved):

PR #846, opened 2023-08-31
Consistent Type Imports

timoheddes:
This PR will implement the rules [consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports/) and [consistent-type-specifier-style](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md) to our ESlint configuration and update the currently existing imports/exports.

`consistent-type-imports` is set to `error` with the following options:

```
{
  fixStyle: 'separate-type-imports',
  prefer: 'type-imports',
},
```

`consistent-type-specifier-style` is set to `error` with `prefer-top-level` as option.


.changeset/sour-spiders-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kadena/react-ui": patch
```
eileenmguo, line 2: Can you update this to `none`? Right now the package isn't published so we don't need to bump versions. Might also be good to discuss this as a team though

.changeset/sour-spiders-reply.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Feat/explicit imports
```
eileenmguo, line 5: When adding changesets I think we should be more descriptive of the change, otherwise it's better to add empty changesets instead. Since this doesn't actually change anything for consumers I would actually add an empty changset instead.

I think of these descriptions as what you would see on npm for the version update

packages/libs/react-ui/.eslintrc.js
```diff
@@ -6,6 +6,15 @@ module.exports = {
...
   ],
+  rules: {
+    '@typescript-eslint/consistent-type-imports': [
+      'error',
+      {
+        fixStyle: 'inline-type-imports',
+        prefer: 'type-imports',
+      },
+    ],
+  },
```
eileenmguo, line 17: Should we move this to the react profile instead?

packages/libs/react-ui/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -1,6 +1,6 @@
-import { Accordion, IAccordionProps } from './Accordion';
+import { type IAccordionProps, Accordion } from './Accordion';
 
-import type { Meta, StoryObj } from '@storybook/react';
+import { type Meta, type StoryObj } from '@storybook/react';
```
eileenmguo, line 3: Is there a way to avoid this update when all the imports are types?

packages/libs/react-ui/.eslintrc.js
timoheddes: Yeah I'm in favour! Figured to at least implement it into our domain but I think it would be good to implement it as a rule everywhere.

packages/libs/react-ui/src/components/Accordion/Accordion.stories.tsx
timoheddes: Yes, you can just do `import type { ... }` actually. You think that would be better?

.changeset/sour-spiders-reply.md
timoheddes: I admit I've just been using the bot's suggested changesets! I'll have a look :D

packages/libs/react-ui/src/components/Accordion/Accordion.stories.tsx
ash-vd: I like the new change better as it's more explicit. Also, when another import from the same package is added in the future that isn't a type, the changes will be easier to spot

timoheddes: Agreed!

eileenmguo: Sounds good to me 👍

packages/libs/react-ui/.eslintrc.js
timoheddes: Done!

webpro (approved):
Awesome!

Is this also possible/auto-fixable only when **all** imports are types?

```ts
import { type ITransaction } from './types'; // ❌
import type { ITransaction } from './types'; // ✅
```

.changeset/sour-spiders-reply.md
alber70g, line 1: I'd expect a `@kadena/eslint-config` change here as this changes the lint rules for the consumer of this package.

packages/apps/docs/src/components/Analytics/Analytics.tsx
```diff
@@ -1,7 +1,7 @@
 import { analyticsPageView } from '@/utils/analytics';
 import { useRouter } from 'next/router';
 import Script from 'next/script';
-import React, { FC, useEffect } from 'react';
+import React, { type FC, useEffect } from 'react';
```
alber70g, line 4: I think there's a rule to separate `type` imports from "normal" imports. I'd like it to be separate and not inlined. Not sure what others think. 

Especially when you get things multiple imports mixed with type imports.

```ts
import React, { type FC, type ReactNode } from 'react';
```

should be 
```ts
import type { FC, ReactNode } from 'react';
import React from 'react';
```


To do so `fixStyle` should be `separate-type-imports`

See: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md#fixstyle

alber70g (changes requested):

packages/apps/docs/src/components/Analytics/Analytics.tsx
timoheddes: Yeah we did discuss this and personally I agree with @ash-vd : https://github.com/kadena-community/kadena.js/pull/846#discussion_r1311427795

What do you think?

alber70g: I'm in favour of separating imports between "types" and "regulars".
It adds cognitive load when reviewing. You'd have to keep track of two things in one line; types and non-types
It's easier to scan all non-types when they aren't mixed
Especially when re-exporting certain types that are for the consumer. In index.ts files this happens often, when working with @kadena/client for example

.changeset/sour-spiders-reply.md
javadkh2, line 1: Do we really need to bump version for this change?

alber70g: It changes the consumers' api. A new lint rule is added, changing the behaviour of the package

sstraatemans (approved):

Takadenoshi (approved):

alber70g (approved):

PR #848, opened 2023-08-31
[@kadena/graph] Implement confirmation depth in graph and client

nil-amrutlal-dept:


packages/apps/graph/src/graph/objects/Block.ts
```diff
@@ -66,3 +73,25 @@ async function getTransactionsRequestkeyByEvent(
...
+      UNION ALL
+      SELECT b.hash, b.parent, d.depth + 1 AS depth
+      FROM BlockDescendants d
+      JOIN blocks b ON d.hash = b.parent
+    )
+    SELECT MAX(depth) AS depth
+    FROM BlockDescendants;
+  `;
+
+  if (result && result[0] && result[0].depth) {
```
MRVDH, line 92: `result[0]` would throw an error if there isn't anything in the array, and `result` is always true. Consider doing `result.length && result[0].depth`

packages/apps/graph-client/.env.example
```diff
@@ -0,0 +1,2 @@
+# Depth necessary on a block to call it "confirmed", omit to use default value: 6
+REACT_APP_MAX_CALCULATED_CONFIRMATION_DEPTH=
```
MRVDH, line 2: Do we need the REACT_APP_ prefix?

MRVDH (approved):

packages/apps/graph-client/.env.example
nil-amrutlal-dept: Since I need to expose these variables to the client (browser), I need it to be there (it does not work without the prefix). 
Also, the prefix in this file was incorrect. The correct one is `NEXT_PUBLIC`, and I already updated the fiel. This information is available on Next.js [documentation](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser)

PR #859, opened 2023-09-04
Remove pre/postinstall scripts

webpro:
99.9% we run `pnpm install` we don't need any of this :)

### DIFF
--- a/packages/libs/chainweb-node-client/package.json
+++ b/packages/libs/chainweb-node-client/package.json
@@ -35,7 +35,7 @@
     "format:md": "remark README.md -o --use @kadena-dev/markdown",
     "format:src": "prettier config openapi src --write",
     "generate:openapi-types": "echo 'openapi specs needs fixes' # openapi-typescript \"./openapi/*.json\" --output ./src/openapi",
-    "postinstall": "pnpm run generate:openapi-types",
+    "_postinstall": "pnpm run generate:openapi-types",
     "lint": "eslint package.json src --ext .js,.ts --fix",
     "test": "heft test"
   },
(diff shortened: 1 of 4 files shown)


packages/libs/chainwebjs/package.json
```diff
@@ -52,7 +52,6 @@
     "format": "pnpm run /^format:.*/",
     "format:md": "remark README.md -o --use @kadena-dev/markdown",
     "format:src": "prettier config src --write",
-    "preinstall": "npx only-allow pnpm",
```
webpro, line 55: We define this in the root `package.json` already.

packages/libs/chainweb-node-client/package.json
```diff
@@ -35,7 +35,7 @@
     "format:md": "remark README.md -o --use @kadena-dev/markdown",
     "format:src": "prettier config openapi src --write",
     "generate:openapi-types": "echo 'openapi specs needs fixes' # openapi-typescript \"./openapi/*.json\" --output ./src/openapi",
-    "postinstall": "pnpm run generate:openapi-types",
```
alber70g, line 38: Shouldn't this be part of the build script too?

```
"build": "pnpm run generate:openapi-types && heft build --clean",
```

packages/libs/chainwebjs/package.json
alber70g: Does this have an effect on clients installing chainwebjs?

alber70g (approved):

packages/libs/chainwebjs/package.json
webpro: Yes, they are no longer forced to use pnpm

packages/libs/chainweb-node-client/package.json
webpro: Didn't bother too much, this isn't doing anything anyway (just `echo`)

PR #869, opened 2023-09-05
Prepare release workflow (fully automated w/ all tokens set up)

webpro:
- Automated workflow for versioning + publishing
- GitHub and npm tokens created and added as secrets (also see comments for access granted)
- Also see https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md
- For now just echo-ing the result to console (last step); could be bot(s) to send notification to Slack/Discord, anything
- We may need to give the `github-actions[bot]` user permission to push to the protected `main` branch

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,9 +32,8 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          # Node.js v18 gives error in changesets/action@v1 (can't find dist/index.js)
-          node-version: 16
-          cache: 'pnpm'
+          node-version: 18
+          cache: pnpm
 
       - name: Install dependencies
         run: pnpm install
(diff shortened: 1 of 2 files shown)


javadkh2 (approved):
💯

.github/workflows/release.yml
```diff
@@ -1,43 +1,63 @@
...
+        run: pnpm turbo build lint test
+
+      - name: Create Release PR or Publish
         id: changesets
         uses: changesets/action@v1
         with:
-          publish: pnpm publish-packages
+          # Note: pnpm install after versioning is necessary to refresh lockfile
+          version: pnpm run version
+          publish: pnpm run publish
```
realdreamer, line 51: `publish` command is default npm publish command or it triggers the publish command(`changeset publish`) from `package.json`

realdreamer (approved):

.github/workflows/release.yml
javadkh2: its juts an alias for "changeset publish", check the root package.json

alber70g (approved):

PR #871, opened 2023-09-05
chore(@kadena/react-ui): Update Button and IconButton components to accept asChild prop

eileenmguo:
Added the asChild prop to the following components:
- Button
- IconButton

Note: It pass all props to child rather than filtering as it does when the element type is chosen via the `as` prop


packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -7,10 +7,13 @@ import {
 import { SystemIcon } from '@components/Icon';
 import cn from 'classnames';
 import type { ButtonHTMLAttributes, FC } from 'react';
-import React from 'react';
+import React, { type ReactNode } from 'react';
```
timoheddes, line 10: With the latest change to our eslint rules, which has been merged by now, this will change slightly still so I recommend another format and save on this file after you rebase with main.

sstraatemans (approved):

PR #872, opened 2023-09-05
Upgrade to Node.js LTS/18 + current/20

webpro:
As [Next.js v16 is EOL at Sept 11](https://nodejs.org/en/blog/announcements/nodejs16-eol).


alber70g (approved):
Shouldn't we use the lowest version that we support? Generally newer versions are more backwards compatible. 

Unless we use specific features from the newer version

webpro:
> Shouldn't we use the lowest version that we support? Generally newer versions are more backwards compatible.
> 
> Unless we use specific features from the newer version

I guess you mean we use **only** the lowest version, then yeah would be a good idea, imho. Uses less resources too (looking at number of runs and remote caching for instance).

The irony is that just this week Node.js v20.6.0 came with a bug, impacting many popular projects. This is rare and usually fixed quickly, but it's not impossible. Still think one version is enough.

PR #875, opened 2023-09-06
Feat/page helpful

realdreamer:
This PR closes https://app.asana.com/0/1205288273408141/1205091936807459/f

Also made a couple of changes on react-ui
 - Adding onClose callback for Modal
 - Adding active state for the IconButton - https://www.figma.com/file/b95LxjVv5ztfWPkmPGNeeq/%3Ckda-button-%2F%3E?type=design&node-id=1250-361&mode=design&t=VMwsfo7fM4hVeqNk-0


packages/apps/docs/src/components/PageHelpful/useModal.tsx
```diff
@@ -0,0 +1,75 @@
+import {
```
sstraatemans, line 1: maybe make the name more specific for the pagehelp thing. because this hook is specific

packages/apps/docs/src/components/PageHelpful/useModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+      pagePath: pathname,
+      isPageHelpful: 'no',
+      comment: inputRef?.current?.value,
+    });
+    clearModal();
+  };
+
+  const renderModalComponent = useCallback((): void => {
+    renderModal(
+      <div className={modalWrapperClass}>
```
sstraatemans, line 39: can u check the design with @isaozler ?
I think the edit button is a bit in a weird position.

sstraatemans (approved):

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
```diff
@@ -55,3 +55,16 @@ export const colorVariants = styleVariants(colors, (color) => {
...
+    return [];
+  }
+
+  return [
+    {
+      outlineOffset: '2px',
+      outline: `2px solid ${vars.colors[`$${color as ColorType}Accent`]}`,
+    },
+  ];
+});
```
eileenmguo, line 70: I think there is potentially a cleaner way to implement this. You can make a var for the outline value of "none" and create a class `activeClass` that updates the var to the appropriate outline value. You might also need to make a var for `accentColor` as well, but I think this approach would still be cleaner since these two style variants are tightly coupled

packages/libs/react-ui/src/components/IconButton/IconButton.tsx
```diff
@@ -14,6 +18,7 @@ export interface IIconButtonProps
   color?: keyof typeof colorVariants;
   type?: 'button' | 'submit' | 'reset';
   asChild?: boolean;
+  state?: 'active';
```
eileenmguo, line 21:
```suggestion
  active?: boolean;
```

packages/libs/react-ui/src/components/IconButton/IconButton.tsx
```diff
@@ -24,25 +29,29 @@ export const IconButton: FC<IIconButtonProps> = ({
...
   children,
   asChild = false,
+  state,
   ...restProps
 }) => {
   const Icon = icon && SystemIcon[icon];
   const ariaLabel = restProps['aria-label'] ?? title;
+  const classNames = classnames(colorVariants[color], {
+    [stateVariants[color]]: state === 'active',
+  });
```
eileenmguo, line 39:
```suggestion
  const classNames = classnames(colorVariants[color], {
    [activeClass]: active,
  });
```

packages/libs/react-ui/src/components/Link/Link.tsx
```diff
@@ -12,17 +13,23 @@ export interface ILinkProps {
...
   asChild = false,
+  inline = false,
   ...restProps
 }) => {
   const Icon = icon && SystemIcon[icon];
 
+  const linkClasses = classnames(linkContainerClass, {
+    [inlineLinkClass]: inline,
+  });
+
```
eileenmguo, line 32: Should the link element be inline by default?

packages/libs/react-ui/src/components/IconButton/IconButton.css.ts
realdreamer: like we discussed over in slack, these styles are moved to default color variants styles. only show/hiding outline with non-active class

packages/libs/react-ui/src/components/Link/Link.tsx
realdreamer: ideally yes! but i made it as a separate prop so we could discuss and align and make changes

timoheddes (approved):

.changeset/wild-mangos-clap.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
webpro, line 2: Are you sure this should be empty? Seeing new things in react-ui

realdreamer: yes! it needs to be updated! It was initially for docs. I'll update it

PR #880, opened 2023-09-06
Separate linting (ESLint) and formatting (Prettier)

webpro:
This PR separate the linting from the formatting. In practice it means a bit more of the formatting work that eslint did has been moved to prettier. In general it's a bit more idiomatic towards how I think the tools are supposed to work.

- Add `eslint-config-prettier` to turn off all rules that are unnecessary or might conflict with Prettier.
- Remove `eslint-plugin-prettier` as we already have Prettier through `pnpm format` etc.
- This should not change much in your IDE as Prettier and ESLint are still both active.
- Adds `--cache` for fasterer runs.
- Unfortunately Prettier requires a `.prettierignore` in each workspace (unlike ESLint that does look up further for more `.gitignore` files). On the other hand, we do like self-contained packages.
- Turns out some files were missed before or need a bit of formatting, part of this PR.
- Adds formatting to the root workspace.

EDIT: One of the reasons for doing this is so we could _potentially_ use something else for either formatting or linting more easily. It should also be more clear locally or in CI what went wrong (is it an ESLint issue or just formatting, etc).


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -2,4 +2,7 @@ blank_issues_enabled: false
 contact_links:
   - name: Kadena Discord
     url: https://discord.io/kadena
-    about: Contact us on Discord if you're having trouble with the library. We have a specific channel for Kadena.js. "#kadena-js" can be found in the "Buidler Zone".
+    about:
+      Contact us on Discord if you're having trouble with the library. We have a
+      specific channel for Kadena.js. "#kadena-js" can be found in the "Buidler
```
eileenmguo, line 7:
```suggestion
      specific channel for Kadena.js. "#kadena-js" can be found in the "Builder
```

.vscode/extensions.json
```diff
@@ -1,13 +1,13 @@
...
-		"editorconfig.editorconfig",
-		"dbaeumer.vscode-eslint",
-		"folke.vscode-monorepo-workspace",
-		"esbenp.prettier-vscode",
-		"eamodio.gitlens",
-		"christian-kohler.path-intellisense",
-		"prisma.prisma",
-		"graphql.vscode-graphql"
-	]
+  "recommendations": [
```
eileenmguo, line 2: 😍

eileenmguo (approved):

.github/ISSUE_TEMPLATE/config.yml
webpro: I think this is correct/intentional 🤷

.vscode/extensions.json
webpro: _Prettier all the things!_

alber70g (approved):

.github/ISSUE_TEMPLATE/config.yml
ash-vd: https://www.coindesk.com/markets/2019/02/02/the-history-of-hodl/ :)

realdreamer (approved):

ash-vd (approved):

sanderlooijenga (approved):

packages/apps/docs/.prettierignore
```diff
@@ -0,0 +1,2 @@
+.next
+**/*.md
```
sstraatemans, line 2: why igone md files?

webpro: Because we have a separate formatter for Markdown files and also because we have quite some generated Markdown files (e.g. from API Extractor).

PR #885, opened 2023-09-07
[@kadena/graph-client] Request Key/Transaction Tracking

MRVDH:


packages/apps/graph-client/src/pages/transaction/[key].tsx
```diff
@@ -0,0 +1,283 @@
...
+import React from 'react';
+
+const StyledMain = styled('main', {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  flexDirection: 'column',
+  my: '5rem',
+});
+
```
Comment (nil-amrutlal-dept) [inline packages/apps/graph-client/src/pages/transaction/[key].tsx:19]: I can see that on the `src/components/loader/styles.css.ts` vanilla extract was used. Any reason to use stitches here?

nil-amrutlal-dept (approved):

packages/apps/graph-client/src/pages/transaction/[key].tsx
Comment (MRVDH) [reply packages/apps/graph-client/src/pages/transaction/[key].tsx]: Good catch, copy pasted

.changeset/chatty-pans-know.md
alber70g, line 1: Will this result in a catch-all for other PRs?

alber70g (approved):

PR #886, opened 2023-09-08
chore(@kadena/react-ui): Add minWidth and maxWidth to `Table.Th` props

eileenmguo:
- Added minWidth + all sizes to sprinkles
- Added minWidth and maxWidth as props to `Table.Th` props


barthuijgen (approved):

MRVDH (approved):

packages/libs/react-ui/src/components/Table/Th.tsx
```diff
@@ -6,12 +6,13 @@ import classNames from 'classnames';
 import type { FC } from 'react';
 import React from 'react';
 
-export interface IThProps extends Pick<Sprinkles, 'width'> {
+export interface IThProps
```
sanderlooijenga, line 9: Out of curiosity; why don't we (also) extend the props of the (React's) `th` element? (and then spreading those props on the element)

timoheddes (approved):

packages/libs/react-ui/src/components/Table/Th.tsx
eileenmguo: Hmm not sure. Maybe it just hasn't needed so it wasn't implemented, but I do think it would be valuable. I've added it to one of the tickets we are refining today

PR #889, opened 2023-09-11
feat: add nav header select component

KristinaSpasevska:


packages/libs/react-ui/src/components/NavHeader/NavHeaderContent.tsx
```diff
@@ -1,10 +1,11 @@
 import { childrenClass } from './NavHeader.css';
 
-import type { FC } from 'react';
+import type { INavHeaderSelectProps } from '@components/NavHeader/NavHeaderSelect';
+import type { FC, FunctionComponentElement } from 'react';
 import React from 'react';
 
 export interface INavHeaderContentProps {
-  children: React.ReactNode;
+  children: React.ReactNode | FunctionComponentElement<INavHeaderSelectProps>[];
```
timoheddes, line 8: I guess this is redundant? As we mainly use `FunctionComponentElement` to restrict but since it accepts any Node it won't actually restrict anything.

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -130,3 +130,72 @@ export const glowClass = style([
...
+    paddingRight: '$2',
+  }),
+  {
+    border: `1px solid ${vars.colors.$gray40}`,
+    selectors: {
+      '&::after': {
+        content: '',
+        position: 'relative',
+
+        backgroundImage: `url('data:image/svg+xml;charset=UTF-8, <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
```
timoheddes, line 156: Should we not use the ChevronDown icon here? Using the Icon component?

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelectOption.tsx
```diff
@@ -0,0 +1,13 @@
+import type { FC } from 'react';
+import React from 'react';
+
+export interface INavHeaderSelectOptionProps
+  extends Omit<React.HTMLAttributes<HTMLOptionElement>, 'as'> {
+  children: React.ReactNode;
+  value?: string[] | string | number;
+}
+
+export const NavHeaderSelectOption: FC<INavHeaderSelectOptionProps> = ({
```
eileenmguo, line 10: Do you think we need to expose this? The other Select component doesn't export the `Option` component even though it has one (which we should either remove or export), but right now we aren't adding anything to the options. What do you think @KristinaSpasevska @timoheddes

eileenmguo (changes requested):
I just pushed an update to update storybook to always show all elements (including select). Things that I think need to change:
- Currently you can't click on the arrow (disable pointer events)
- When clicking on the arrow, the other icon turns red on the NavHeader.Select
- I can't seem to tab to the item or there isn't a style for it. Should this have a focus state?
- When removing the button the arrow icon is off. It should always be centered
- I would suggest that you ask Isa to check how the normal select looks since the arrow looks different

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelectOption.tsx
timoheddes: I think it makes sense to match the other Select component and thus not export it.

KristinaSpasevska: Same goes for me, if that's the way the Select was built, this should be similar

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -130,3 +130,99 @@ export const glowClass = style([
...
   },
 ]);
+
+export const selectContainerClass = style([
+  sprinkles({
+    alignItems: 'stretch',
+    backgroundColor: {
+      lightMode: '$gray90',
+      darkMode: '$gray90',
+    },
```
eileenmguo, line 140: nite: If it will stay they same color, you don't need to define lightMode and darkMode separately

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+) {
+  const Icon = icon && SystemIcon[icon];
+  const ChevronDown = SystemIcon.ChevronDown;
+
+  return (
+    <div
+      className={classNames(selectContainerClass, {
+        [selectContainerClassDisabled]: disabled,
+      })}
+      data-testid="kda-select"
```
eileenmguo, line 44: is this being used?

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -130,3 +130,99 @@ export const glowClass = style([
...
+
+export const chevronIconClass = style([
+  sprinkles({
+    marginRight: '$2',
+    position: 'absolute',
+    right: '$1',
+    color: '$gray40',
+  }),
+  {
+    top: '7px',
```
eileenmguo, line 219: I don't think this should be absolutely. Rather I think it would be better to position the contents with flex to center. Right now I think if the height of the select changes for any reason, the alignment will be off

eileenmguo (commented):
The style looks off on focus as well. Can you check with Isa how this is supposed to work? 
image

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelect.tsx
KristinaSpasevska: If this can be disabled, yes. If we don't need the 'disabled' option at all, then I will remove this. 
@eileenmguo let me know what you think.

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -210,13 +207,15 @@ export const selectClass = style([
 
 export const chevronIconClass = style([
   sprinkles({
+    height: '$9',
```
eileenmguo, line 210: would it make more sense to position it to the top, bottom, and right? Right now if the height doesn't match the select component, it will still not be centered

packages/libs/react-ui/src/components/NavHeader/NavHeader.stories.tsx
```diff
@@ -107,9 +103,9 @@ export const Dynamic: IStory = {
     customNavigation,
     linksCount,
     navHeaderActiveLink,
-    renderSampleContent = false,
   }) => {
     const navItems = useCustomNavigation ? customNavigation : sampleNavItems;
+    const selectedNetwork = 'Mainnet';
```
timoheddes, line 108: nit but could be `sampleNetworkItems[0]`

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelect.tsx
eileenmguo: ahh yes, sorry i did mean the testid

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -207,15 +207,13 @@ export const selectClass = style([
...
-    height: '$9',
     marginRight: '$2',
     position: 'absolute',
     right: '$1',
     color: '$gray40',
-    display: 'inline-flex',
-    alignItems: 'center',
   }),
   {
+    top: '20%',
```
eileenmguo, line 216: What was the reason for removing the flex properties and making this 20% from the top all the time? Right now it looks like the select component grows to fit it's container, but the icon doesn't. So this could still result in the the icon not being centered

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -130,3 +130,92 @@ export const glowClass = style([
...
+  },
+]);
+
+export const selectClass = style([
+  sprinkles({
+    background: 'none',
+    border: 'none',
+    color: '$gray40',
+    flexGrow: 1,
+    outline: 'none',
```
timoheddes, line 189: I think we do want an outline for when you tab to the element? Currently the rest of the navHeader can be navigated by tabbing and all the items are visually highlighted when you do except this one.

Not sure though, @eileenmguo ?

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <span className={selectIconClass}>
+          <Icon size="md" />
+        </span>
+      )}
+      <select
+        aria-label={ariaLabel}
+        className={selectClass}
+        disabled={Boolean(disabled)}
+        ref={ref}
+        {...rest}
```
timoheddes, line 55: I don't think we should spread the rest here. I see that it's the same with the original Select component we have and think we should remove this line from both.

Also, if we do spread this in, check your console and you should see a warning related to setting the value without an onChange handler.

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
eileenmguo: Yeah, this will have an outline, but it will match the style of the input and select component that @isaozler is designing so that update will be included in that ticket

packages/libs/react-ui/src/components/NavHeader/NavHeaderSelect.tsx
eileenmguo: mmm I do think we would want to pass the native select props to the select components. Since this is a thin style wrapper around a native select element, I don't think it should take away any native functionality. The warning should be resolved though, I think maybe we can remove the onChange prop in the types or something since it will be included with the native select element.

KristinaSpasevska: I will remove the onChange prop in the type and use it from the native select

eileenmguo (approved):

PR #897, opened 2023-09-12
[DOCS]chore: content change

sstraatemans:


packages/apps/docs/src/pages/docs/build/support/index.md
sstraatemans, line 1: this was double content

alber70g (approved):

PR #899, opened 2023-09-12
feat(@kadena/react-ui): Add SelectField Component

eileenmguo:
This PR adds a SelectField component which is a composition of the `InputWrapper` and `Select` components.

Asana Ticket: https://app.asana.com/0/1205288273408141/1205375258118259/f


packages/libs/react-ui/src/components/SelectField/index.ts
```diff
@@ -0,0 +1 @@
+export { type ISelectFieldProps, SelectField } from './SelectField';
```
timoheddes, line 1: I think the linter should split these now?

packages/libs/react-ui/src/components/SelectField/SelectField.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface ISelectFieldProps extends Omit<IInputWrapperProps, 'htmlFor'> {
+  selectProps: Omit<ISelectProps, 'disabled' | 'children' | 'leadingTextWidth'>;
+}
+
+export const SelectField: FC<ISelectFieldProps> = ({
+  disabled = false,
+  selectProps,
+  children,
+  ...rest
```
timoheddes, line 16: Sometimes we spread props and sometimes we don't. Do we have clear rules or something to decide when we do?

eileenmguo: I think it's fine to use the spread operator since we know what those props will be using typescript. I think the distinction is more around when you extend our prop types to all of the available props for a specific element which depends on the use case. In the case of input elements I think we don't want to restrict anything since our input components should be a thin wrapper.

packages/libs/react-ui/src/components/SelectField/index.ts
eileenmguo: hmm I'm not sure. I don't see any errors so maybe this case isn't covered?

eileenmguo: I think it makes sense to have it work for exports too, but that would need to be a separate PR. I'll update this specific case for this PR

timoheddes (approved):

PR #900, opened 2023-09-12
Remove colors from Notification component

timoheddes:
As requested by Isa, removed the secondary and tertiary color variants from the Notification component.

[Asana Link](https://app.asana.com/0/1205407980256034/1205463910302470/f)


packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -1,16 +1,14 @@
 import { sprinkles } from '@theme/sprinkles.css';
-import type { ColorType } from '@theme/vars.css';
 import { vars } from '@theme/vars.css';
 import { style, styleVariants } from '@vanilla-extract/css';
 
-const colors: Record<ColorType, ColorType> = {
+export type NotificationColors = 'info' | 'positive' | 'warning' | 'negative' | 'primary';
+const colors: Record<NotificationColors, NotificationColors> = {
```
eileenmguo, line 6: I think it's better to use `Omit<ColorType, "secondary" | "tertiary">` since this ties this type to the original

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -27,7 +28,7 @@ const meta: Meta<
     color: {
       options: Object.keys(
         cardColorVariants,
-      ) as (keyof typeof cardColorVariants)[],
+      ).filter(e => e !== 'tertiary' && e !== 'secondary') as (NotificationColors)[],
```
eileenmguo, line 31: I think you can simplify this and just use Object.keys(color) to do this. I believe it will also already be typed for you

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -1,3 +1,5 @@
+import type {
```
eileenmguo, line 1: lint?

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -17,7 +19,7 @@ export interface INotificationProps {
   title?: string;
   children?: React.ReactNode;
   expanded?: boolean;
-  color?: keyof typeof cardColorVariants;
+  color?: NotificationColors;
```
eileenmguo, line 22: nit: I have a preference to keep it how it was for the sake of defining less types

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
timoheddes: Yeah my implementation of this was a bit rushed in between other things since Isa asked. My bad, I have refactored it to a different approach.

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
sanderlooijenga: It's not a type though, is it?

realdreamer (approved):

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
timoheddes: It's the linter that's causing that.. I'll add an eslint-disable-next-line

packages/libs/react-ui/src/components/Notification/NotificationContainer.tsx
```diff
@@ -1,10 +1,12 @@
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
```
timoheddes, line 1: Since I'm only using the colorVariants const in an interface, ESLint turns it into a type import while it's not so I added this line

sanderlooijenga (approved):

eileenmguo (approved):

PR #906, opened 2023-09-14
chore(monorepo): add workspaces in package.json to allow `bun i`

alber70g:


webpro (changes requested):
We can't just merge this without making sure everything actually works :)

Fast installation is awesome, but there's more... 🤷

alber70g:
Yes definitely. I only saw this working. I see that it doesn't install in the node_modules for the packages, so yes, there's a long road ahead

PR #913, opened 2023-09-15
[@kadena/client] add `signWithKeypair` functionality

alber70g:
- feat(client): adds `createSignWithKeypair` and allows `signWithKeypair`

This new function allows you to sign with:
- one keypair, one transaction
- one keypair, multiple transactions
- multiple keypairs, one transaction
- multiple keypairs, multiple transactions, multiple times

An example of the most comprehensive variant:

```typescript
const signWithKeystore = createSignWithKeypair([keyPair, keyPair2]);
const [signedTx1, signedTx2] = await signWithKeystore([tx1, tx2]);
const signedTx3 = await signWithKeystore(tx3);
```

### DIFF
--- a/packages/libs/client/src/signing/keypair/createSignWithKeypair.ts
+++ b/packages/libs/client/src/signing/keypair/createSignWithKeypair.ts
@@ -2,7 +2,7 @@ import { signHash } from '@kadena/cryptography-utils';
 import type { IUnsignedCommand } from '@kadena/types';
 
 import type { IPactCommand } from '../../interfaces/IPactCommand';
-import type { ISignFunction, ISingleSignFunction } from '../ISignFunction';
+import type { ISignFunction } from '../ISignFunction';
 import { addSignatures } from '../utils/addSignatures';
 import { parseTransactionCommand } from '../utils/parseTransactionCommand';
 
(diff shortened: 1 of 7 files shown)


ash-vd (approved):

packages/libs/client/src/signing/keypair/createSignWithKeypair.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Sign with public/private key-pair according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
+ *
+ * @public
+ */
+export const createSignWithKeypair: (
+  keyOrKeys: IKeypair | IKeypair[],
+) => ISignFunction & ISingleSignFunction = (
```
javadkh2, line 26: `ISingleSign` interface already extends `ISingleSignFunction` So you don need to do this again here

javadkh2: its better to use function overloads instead
```TS
interface ICreateSignWithKeypair {
    /** jsdoc for this variant */
   ( key: IKeypair) => ISingleSignFunction ,
   
    /** jsdoc for this variant */
   (keys: IKeypair[]) => ISignFunction
}

export const createSignWithKeypair: ICreateSignWithKeypair = ...
```

alber70g: I've modified it to your suggestion

packages/libs/client/src/signing/keypair/tests/signWithKeypair.test.ts
```diff
@@ -0,0 +1,230 @@
...
+
+const keyPair: IKeypair = {
+  publicKey: '09e82da78d531e2d16852a923e9fe0f80f3b67a9b8d92c7f05e4782222252e12',
+  secretKey: '76e1cabaa58a33321982e434f355dc7a4cfbee092a4ac1c7aac26302ba80d992',
+};
+
+const coin: ICoin = getModule('coin');
+
+describe('createSignWithKeypair', () => {
+  let signWithKeypair: ISignFunction & ISingleSignFunction;
```
javadkh2, line 16: `ISignFunction` is enough.
I think you only need to use `ISignFunction` in other places in this file as well instead of (`ISingleSignFunction` or `ISignFunction & ISingleSignFunction` ) because createSignWithKeypair always returns a function that accepts both.

packages/libs/client/src/signing/keypair/createSignWithKeypair.ts
```diff
@@ -0,0 +1,128 @@
...
+  keypairs: IKeypair[],
+): IKeypair[] {
+  const relevantKeypairs = keypairs.filter((keypair) =>
+    tx.signers.some(({ pubKey }) => pubKey === keypair.publicKey),
+  );
+  debug('relevant keypairs', relevantKeypairs);
+  return relevantKeypairs;
+}
+
+function signWithKeypairs(
```
javadkh2, line 118: Shouldn't we export this function as well?
Consider I have developed my way to manage the keypairs (for example, in the dev-wallet we already did) and just need a simple function that signs the transaction with a list of keypairs.
It gives users more flexibility.

alber70g: The `createSignWithKeypair` doesn't really change that, right? Even if you had a way to manage that you can still use `const sign = createSignWithKeypair(allMyKeys)(txs)`. Also this one considers only keypairs that are already checked whether they're needed for this tx. 

Secondly, exposing this one as extra API introduces more surface that we need to maintain

packages/libs/client/src/signing/keypair/tests/signWithKeypair.test.ts
alber70g: Fixed

javadkh2 (approved):

PR #915, opened 2023-09-15
fixes the test for verifySig

ash-vd:
The test for verifySig wasn't correct and using invalid data


LindaOrtega (approved):
LGTM!

One small nit:
Could we add a test where we use crypto's `sign` function directly? I think it could help show how the outputs of `sign` can be used with the `verifySig` function.

ash-vd:
> LGTM!
> 
> One small nit: Could we add a test where we use crypto's `sign` function directly? I think it could help show how the outputs of `sign` can be used with the `verifySig` function.

Unit tests are meant to test just the functions themselves. If we test `sign` here as well the tests become less isolated so I don't want to add that test here.

PR #918, opened 2023-09-18
Feat/tools website/add notification when mainnet is active

KristinaSpasevska:


packages/apps/tools/src/pages/faucet/existing/index.tsx
```diff
@@ -136,6 +140,16 @@ const ExistingAccountFaucetPage: FC = () => {
           }}
           body={requestStatus.message}
         />
+        {showNotification ? (
+            <FormStatusNotification
+              status={'erroneous'}
+              statusBodies={{
+                erroneous: t('You cannot use faucet on Mainnet.'),
```
sanderlooijenga, line 147: I'm not sure whether this is the appropriate message. Maybe it should be more of a warning, indicating to users that the Faucet will only fund users on testnet, therefor switching between networks has no effect on this call. Not sure what Isa or Albert thinks about it?

packages/apps/tools/src/pages/faucet/existing/index.tsx
```diff
@@ -136,6 +140,16 @@ const ExistingAccountFaucetPage: FC = () => {
...
+              }}
+              body={t('You cannot use faucet on Mainnet.')}
```
sanderlooijenga, line 149: Should an entry be added to the translation file?

packages/apps/tools/src/pages/faucet/existing/index.tsx
```diff
@@ -128,6 +133,16 @@ const ExistingAccountFaucetPage: FC = () => {
         <Breadcrumbs.Item>{t('Existing')}</Breadcrumbs.Item>
       </Breadcrumbs.Root>
       <Heading as="h4">{t('Add Funds to Existing Account')}</Heading>
+      <div className={notificationContainerStyle}>
+        {showNotification ? (
+          <Notification.Root
+            color="warning"
+            expanded={true}
+            icon="Information"
+            title={t('The faucet can not be used on Mainnet. You can switch networks in the top bar.')}
```
sanderlooijenga, line 142: I still think we should display a different message. This one is kind of incorrect, as in, the network toggle has no effect on the Faucet, it just always executes on Testnet. So perhaps better might be, when on Mainnet, to show a notification like;

`The Faucet only runs on Testnet. The Network toggle in the topbar has no effect on the Faucet. It will not fund accounts on Mainnet, only on Testnet.`

Perhaps something along those lines, what do you think?

KristinaSpasevska: Sound fine, yeah. I will change the text to this version

sanderlooijenga (approved):

PR #922, opened 2023-09-18
Fix `types` path and make linter yell

webpro:
The `"types"` field should point to what API Extractor generates in the `dist` folder.

There is one issue though, added a note.


packages/libs/pactjs/package.json
```diff
@@ -24,7 +24,7 @@
     }
   ],
   "main": "lib/index.js",
-  "types": "lib/index.d.ts",
+  "types": "dist/pactjs.d.ts",
```
webpro, line 27: This one causes trouble:

```
[typescript] Encountered 2 TypeScript issues:
[typescript] Error: src/tests/mkCap.test.ts:54:57 - (TS2339) Property 'toPactInteger' does not exist on type 'PactNumber'.
[typescript] Error: src/tests/mkCap.test.ts:55:62 - (TS2339) Property 'toPactDecimal' does not exist on type 'PactNumber'.
 ---- Compile encountered an error (1577ms) ---- 
-------------------- Finished (1.756s) --------------------
Encountered 2 errors:
  [typescript] src/tests/mkCap.test.ts:54:57 - (TS2339) Property 'toPactInteger' does not exist on type 'PactNumber'.
  [typescript] src/tests/mkCap.test.ts:55:62 - (TS2339) Property 'toPactDecimal' does not exist on type 'PactNumber'.
```

sanderlooijenga (approved):

packages/libs/pactjs/package.json
webpro: The issue is probably caused by the triple slash directive in `PactNumber.ts` that ends up in `lib/PactNumber.d.ts`, but not in `dist/pactjs.d.ts`.

webpro: API Extractor seems to not handle the triple slash directive:

```ts
/// <reference path="./extend-bignumber.ts" />
```

It's just gone (I think it should inline `extend-bignumber.ts`). To me this seems normal and valid usage of TypeScript. AE seems to [support only `types` and `lib` references](https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/src/generators/DtsRollupGenerator.ts#L95-L102).

I did try a few things with `files` in `tsconfig.json` and `bundledPackages` in `api-extractor.json`, but to no avail.

Not proud of it, but this does seem to do the trick:

```
"build": "heft build --clean && cat src/extend-bignumber.ts >> dist/pactjs.d.ts"
```

There are other ways around this, but making exceptions to the `types": "dist/[pkg].d.ts"` rule or other TypeScript hacks to me sound even worse.

Any ideas welcome, though!

If we don't come up with a real fix and merge this I could open an issue in the rushstack repo.

javadkh2:
Why? I'd rather to use a file generated by typescript itself. it also come with a source map file.
> The "types" field should point to what API Extractor generates in the dist folder.

webpro:
> Why? I'd rather use a file generated by typescript itself. it also come with a source map file.
> 
> > The "types" field should point to what API Extractor generates in the dist folder.

This was a request from @alber70g. You can look at https://api-extractor.com for some reasoning under "DTS rollup". It makes sense, until we hit issues like in this PR.

Tbh not a fan myself, just using `tsc` output is what everyone else does indeed, and would eliminate similar issues.

alber70g (approved):

packages/apps/graph/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "ISC",
   "author": "",
   "main": "index.js",
+  "types": "dist/graph.d.ts",
```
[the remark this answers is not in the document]
webpro: Right, I've also removed this `config/api-extractor.json`

packages/libs/pactjs/package.json
webpro: Unfortunately not, declaration gets stripped, not sure how to work around this.

packages/tools/cookbook/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "ISC",
   "author": "",
   "main": "index.js",
+  "types": "dist/cookbook.d.ts",
```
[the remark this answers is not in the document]
webpro: Right, I've also removed this `config/api-extractor.json`

alber70g (approved):

PR #924, opened 2023-09-19
Fix lint in CI workflow, divide `format` vs `lint` commands

webpro:
We've been mixing up the commands a bit. This PR includes:

1) a fix to actually throw (halt CI) when formatting is off
2) remove CI workflow from the `main` branch (it's in the release workflow too)
3) an attempt to divide this better:

- Use `pnpm format` to format (`prettier --write && eslint --fix`)
- Use `pnpm lint` to lint (`prettier --check && eslint`)

Thus, no auto-fixing during linting. Use format for that.

To check that this PR is actually doing 1) it's properly blocked by #923. The cause of this issue was that `prettier --write --check` doesn't halt the process but just `prettier --check` does.


packages/tools/cookbook/package.json
```diff
@@ -12,10 +12,10 @@
     "build:local": "pnpm run pactjs:coin && heft build --clean",
     "format": "pnpm run /^format:.*/",
     "format:md": "remark README.md -o --use @kadena-dev/markdown",
-    "format:src": "prettier . --cache --write",
+    "format:lint": "pnpm run lint:src --fix","format:src": "prettier . --cache --write",
```
sanderlooijenga, line 15: Commands should be on separate lines?

sanderlooijenga (changes requested):

packages/tools/cookbook/package.json
webpro: Definitely! Guess those were skipped because of other failing formatting (see desc.). Fixed.

packages/apps/docs/package.json
```diff
@@ -24,10 +24,12 @@
     "cypress:run": "percy exec -- cypress run",
     "dev": "pnpm run build:scripts && next dev",
     "format": "pnpm run /^format:.*/",
+    "format:lint": "pnpm run lint:src --fix",
     "format:src": "prettier . --cache --write",
     "lint": "pnpm run /^lint:.*/",
+    "lint:fmt": "prettier . --cache --check",
```
sanderlooijenga, line 30: Where does `fmt` stand for btw?

webpro: format (see e.g. https://dprint.dev/cli/#formatting-files)

sanderlooijenga: Right, so we got `format:lint` and `lint:format`, gotcha

sanderlooijenga (approved):

packages/apps/docs/package.json
webpro: lol, that's... odd/twisted

Ideas welcome! In practice it's "auto-save" in IDE and `pnpm lint` or `pnpm format` in the shell, so not that problematic I guess.

ash-vd (approved):

realdreamer (approved):

packages/apps/docs/package.json
sanderlooijenga: Wasn't meant sarcastically btw, just verifying :D Thanks for the clarification!

webpro: Not taken that way! I didn't see it myself, and it is odd. But not wrong :)

timoheddes: Get a room you guys.

timoheddes (approved):

PR #938, opened 2023-09-20
Remove workaround that was needed for the `allowWarningsInSuccessfulBuild: false` rush setting

webpro:
After moving away from Rush we no longer need this workaround.


barthuijgen (approved):
👍

alber70g (approved):

PR #939, opened 2023-09-20
Upgrade `typescript` and `@types/node` dependencies

webpro:
* Update typescript from 5.1.6 to 5.2.2 (6ef515ac6)
* Upgrade @types/node to latest v18 (ef5e7993b)
* Upgrade root dependencies (ef400984f)
* Update changelog (256a558db)

### DIFF
--- a/.changeset/swift-moons-act.md
+++ b/.changeset/swift-moons-act.md
@@ -1,0 +1,29 @@
+---
+'@kadena/docs': patch
+'@kadena/graph': patch
+'@kadena/graph-client': patch
+'@kadena/immutable-records': patch
+'@kadena/chainweb-node-client': patch
+'@kadena/chainweb-stream-client': patch
+'@kadena/chainwebjs': patch
+'@kadena/client': patch
+'@kadena/client-examples': patch
+'@kadena/cryptography-utils': patch
+'kadena.js': patch
+'@kadena/pactjs': patch
+'@kadena/pactjs-generator': patch
+'@kadena/react-components': patch
+'@kadena/react-ui': patch
+'@kadena/types': patch
+'@kadena/cookbook': patch
+'@kadena-dev/eslint-config': patch
+'@kadena-dev/eslint-plugin': patch
+'@kadena-dev/heft-rig': patch
+'@kadena/kda-cli': patch
+'@kadena-dev/lint-package': patch
+'@kadena/pactjs-cli': patch
+'@kadena-dev/markdown': patch
+'@kadena-dev/scripts': patch
+---
+
+Upgrade `typescript` and `@types/node` dependencies
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
   "devDependencies": {
     "@changesets/cli": "^2.26.2",
     "@kadena-dev/markdown": "workspace:*",
-    "knip": "^2.21.2",
+    "knip": "^2.25.2",
     "only-allow": "^1.1.1",
-    "syncpack": "^10.9.3",
-    "tsx": "^3.12.8",
+    "syncpack": "^11.2.1",
+    "tsx": "^3.12.10",
     "turbo": "^1.10.14"
   },
   "engines": {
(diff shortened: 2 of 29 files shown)


alber70g (approved):

timoheddes (approved):

webpro:
The lowest Node.js version we support is v18, so this PR updates `@types/node` to v18. This resulted in [the build to fail](https://github.com/kadena-community/kadena.js/actions/runs/6252296416/job/16994917693#step:7:100):

```
Project: @kadena/chainweb-stream-client@0.1.0
Heft version: 0.50.6
Node version: v18.17.1
Error: Unable to resolve a TypeScript compiler package
```

- We'll need [`@rushstack/heft` v0.59.0 or higher](https://github.com/microsoft/rushstack/blob/main/apps/heft/CHANGELOG.md) that also updates `@types/node` to v18
- We are at v0.50.6, breaking changes in v0.51.0, v0.52.0, v0.53.0, v0.58.0 (and v0.61.0)

So let's install v0.51.0 first and see what happens:

```
$ pnpm run build

[...]

@kadena-dev/eslint-plugin:build: Error: File does not exist: /Users/lars/p/kadena/kadena.js/packages/tools/eslint-plugin/config/heft.json
```

Cool, we can just clean up `@kadena-dev/eslint-plugin` as it actually has no build step or tests.

Then some error about the outdated `heft.json`. Let's see [the v0.51 migration guide](https://rushstack.io/blog/2023/06/16/heft-migration-guide/).

We'll update `packages/tools/heft-rig/profiles/lax/config/heft.json`, and we need the plugins are separate npm packages now, let's install them:

```
pnpm add @rushstack/heft-typescript-plugin @rushstack/heft-api-extractor-plugin @rushstack/heft-jest-plugin

[...]

packages/tools/heft-rig
├─┬ @rushstack/heft-typescript-plugin 0.2.2
│ └── ✕ unmet peer @rushstack/heft@0.61.0: found 0.51.0
├─┬ @rushstack/heft-api-extractor-plugin 0.2.2
│ └── ✕ unmet peer @rushstack/heft@0.61.0: found 0.51.0
└─┬ @rushstack/heft-jest-plugin 0.5.12
  └── ✕ unmet peer @rushstack/heft@^0.50.6: found 0.51.0
```

Indeed, trying `pnpm build` gives errors. We'll need to downgrade the plugins:

```
pnpm up @rushstack/heft-typescript-plugin@0.1.0 @rushstack/heft-api-extractor-plugin@ 0.1.0 @rushstack/heft-jest-plugin@0.6.0
```

Let's try again:

```
pnpm turbo build -F @kadena/chainweb-stream-client

[...]

@kadena/chainweb-stream-client:build: [build:typescript] Error: Unable to resolve "typescript". For more information on riggable dependency resolution, see https://rushstack.io/pages/heft/rig_packages/#3-riggable-dependencies
@kadena/chainweb-stream-client:build: Encountered 1 error
@kadena/chainweb-stream-client:build:   [build:typescript] Unable to resolve "typescript". For more information on riggable dependency resolution, see https://rushstack.io/pages/heft/rig_packages/#3-riggable-dependencies
```

- Following the link I ended up on https://heft.rushstack.io/pages/plugins/typescript/, only to find that their example also uses `"typescript"`. I don't understand what `"typescript"` the error message refers to. The `typescript` dependency is installed (only) in the `@kadena-dev/heft-rig` workspace.
- Updated the plugins to be `dependencies` (not `devDependencies`).
- Turns out adding `typescript` to the `@kadena/chainweb-stream-client` workspace removes the error. Yay, we have a build! But the idea is that this would not be necessary, looking at the first line of the documentation they link to in the error:

> A rig package can also provide NPM dependencies, to avoid having to specify
> them as "devDependencies" for your project.

At this point I'm done. No idea how much work would be left to do. But with the `tsconfig.json` and `typescript` dependency in the local workspace, what's the point of Heft? The plugin offers `additionalModuleKindsToEmit` and `staticAssetsToCopy`, which we're not using or would be trivial to do. We are not using the ESLint or Storybook plugins at all, and Jest is something we want to move away from. That leaves API Extractor. So when removing Heft/Rig from the monorepo now we'll need to take care of Jest and API Extractor.

Apart from the fact that I only wanted to upgrade `@types/node` to v18, having to do those tedious migrations is a waste of time. It's supposed to save time. Shared configuration is just an `"extends"` property away anyway.

That's why I'm closing this PR for now and will open a new one, so we can see what removing Heft looks like. In the end it comes down to the same reason we abandoned Rush: Heft/Rig is a leaky abstraction.

PR #942, opened 2023-09-21
Decrease padding on "compact" button variant

timoheddes:


sstraatemans (approved):

packages/libs/react-ui/src/components/Button/Button.css.ts
```diff
@@ -108,6 +108,7 @@ export const compactVariant = styleVariants(colorVariants, (variant) => {
       color: `$${variant}ContrastInverted`,
     }),
     {
+      padding: `${vars.sizes.$1} ${vars.sizes.$1}`,
```
eileenmguo, line 111: you can add this as a sprinkle

PR #946, opened 2023-09-21
[@kadena/graph] feat: Simulate traffic on devnet v1

nil-amrutlal-dept:
**Devnet simulate traffic v1**
- creates accounts on devnet and funds these
- creates traffic within these accounts, both transfers and crosschain-transfers
- randomly does regular or crosschain-transfers and randomly selects to which account to send the funds to
- uses `sender00` to fund gas costs and also to fund new chain accounts (to be changed in future versions)

Simulate v1 is an initial implementation of devnet simulated traffic. In this version, the main goal was to get the multiple accounts to transfer funds to each other to have transactions, chain accounts, and account balances.

The next iteration will rely less on `sender00` and will check balances to make sure there are funds instead of just adding funds.

Note: `readme.md` has been updated and scripts to fund accounts and simulate traffic have been added

---
  - https://app.asana.com/0/0/1205530098605072


packages/apps/graph/README.md
```diff
@@ -108,3 +113,19 @@ docker run --rm -it --add-host=devnet:host-gateway enof/devnet:l2-latest --task=
...
+
+### Simulate traffic on the devnet
+
+```sh
+npm run simulate -- -a <accounts> -i <timeInterval> -t <transferAmount>
+```
+
+- accounts - number of accounts to be created in the devnet
+- timeInterval - frequency of transactions
+- transferAmount - amount to be transfered in each transaction
```
MRVDH, line 131: Does this support floats or only ints?

packages/apps/graph/package.json
```diff
@@ -8,28 +8,38 @@
   "author": "",
   "main": "index.js",
   "scripts": {
-    "build": "pnpm prisma:generate && heft build --clean",
+    "prebuild": "pnpm run pactjs:generate:contract:chain",
+    "build": "pnpm prebuild && pnpm prisma:generate && heft build --clean",
```
MRVDH, line 12: If pnpm doesn't support pre/post commands, then we might as well just do `"build": "pnpm run pactjs:generate:contract:chain && pnpm prisma:generate && heft build --clean",` and remove the prebuild command

packages/apps/graph/src/scripts/devnet/config.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  NETWORK_ID: 'fast-development',
+  CHAIN_ID: '1',
```
MRVDH, line 3: Maybe default to chain 0, since that is the first chain

MRVDH (changes requested):
Some minor comments. Looks good though!

packages/apps/graph/README.md
```diff
@@ -108,3 +113,19 @@ docker run --rm -it --add-host=devnet:host-gateway enof/devnet:l2-latest --task=
 
 > **Windows alternative**  
 > replace `$(pwd)` with `%cd%`
+
+### Fund an account on the devnet
+
+```sh
+npm run fund -- -k <publickey>
```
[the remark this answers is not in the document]
nil-amrutlal-dept: In this case is `<publickey>`. So for example, for account `k:123456789` you'd run `npm run fund -- -k 123456789`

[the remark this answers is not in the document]
MRVDH: Is there a way to also fund a 8named) account?

packages/apps/graph/README.md
```diff
@@ -108,3 +113,19 @@ docker run --rm -it --add-host=devnet:host-gateway enof/devnet:l2-latest --task=
...
+```
+
+### Simulate traffic on the devnet
+
+```sh
+npm run simulate -- -a <accounts> -i <timeInterval> -t <transferAmount>
+```
+
+- accounts - number of accounts to be created in the devnet
+- timeInterval - frequency of transactions
```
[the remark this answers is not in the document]
nil-amrutlal-dept: Miliseconds. Update file to clarify this

packages/apps/graph/README.md
nil-amrutlal-dept: Floats. Updated file to clarify this

packages/apps/graph/package.json
nil-amrutlal-dept: prebuild removed 👍🏽

packages/apps/graph/README.md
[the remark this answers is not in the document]
nil-amrutlal-dept: Yep, I think we could get the `publickey` of an `accountName` by doing `Pact.modules.coin.details(<accountName>)`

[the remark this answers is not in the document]
MRVDH: Let's keep it as it is for now then 👍

MRVDH (approved):

packages/apps/graph/src/scripts/devnet/config.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  NETWORK_ID: 'fast-development',
+  CHAIN_ID: '0',
+  PORT: 8080,
+} as const;
```
alber70g, line 5: Can we make this a named export?

```suggestion
export const devnetConfig = {
  NETWORK_ID: 'fast-development',
  CHAIN_ID: '0',
  PORT: 8080,
} as const;
```

packages/apps/graph/src/scripts/devnet/get-balance.ts
```diff
@@ -0,0 +1,35 @@
+import { Pact } from '@kadena/client';
+import type { IPactDecimal } from '@kadena/types';
+
+import config from './config';
+import type { IAccount } from './helper';
+import { dirtyRead } from './helper';
+
+export async function getBalance(account: IAccount): Promise<IPactDecimal> {
+  console.log(`Getting balance for ${account.account}`);
```
alber70g, line 9: Instead of console.log, we could use a more generic solution. That can also be used to log to file etc. Maybe we can use a logging interface that, based on env vars or otherwise logs in different ways.

I made something 😅 😉 
https://github.com/kadena-community/kadena.js/blob/feat/ag/generic-logger/packages/apps/graph/src/utils/logger.ts

packages/apps/graph/src/scripts/devnet/simulate.ts
```diff
@@ -0,0 +1,61 @@
+import { crossChainTransfer } from './crosschain-transfer';
+import type { IAccount } from './helper';
+import { createAccount, getRandomChainId } from './helper';
+import { transfer } from './transfer';
+
+export async function simulate(
+  noAccounts: number = 2,
```
alber70g, line 7:
```suggestion
  numberOfAccounts: number = 2,
```

packages/apps/graph/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "./node_modules/@kadena-dev/heft-rig/profiles/default/tsconfig-base.json",
   "compilerOptions": {
-    "types": ["node"],
+    "types": [".kadena/pactjs-generated", "node"],
+    "skipLibCheck": true,
```
alber70g, line 5: Why is this necessary?

alber70g (approved):

PR #949, opened 2023-09-21
Update README.md

EnoF:
Update `senderAccount` to use the `k:your-pubkey` instead of just the `pubkey`

### DIFF
--- a/packages/libs/client/README.md
+++ b/packages/libs/client/README.md
@@ -185,7 +185,7 @@ const unsignedTransaction = Pact.builder
       decimal: '231',
     }),
   ])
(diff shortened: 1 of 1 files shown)


packages/libs/client/README.md
```diff
@@ -185,7 +185,7 @@ const unsignedTransaction = Pact.builder
       decimal: '231',
     }),
   ])
-  .setMeta({ chainId: '1', senderAccount: 'your-pubkey' })
+  .setMeta({ chainId: '1', senderAccount: 'k:your-pubkey' })
```
javadkh2, line 188: Maybe its better to use "gas-payer-account"

alber70g: Agreed :)

PR #950, opened 2023-09-22
Feat/module explorer redesign

sanderlooijenga:
Initial redesign of the module explorer page. UI/UX still needs to be completed, so parts will be improved/added later.

---
  - https://app.asana.com/0/0/1205486725990454


packages/apps/tools/src/components/Global/ModuleExplorer/SidePanel/styles.css.ts
```diff
@@ -0,0 +1,14 @@
+import { style } from '@vanilla-extract/css';
+
+export const containerStyle = style([
+  { display: 'flex', flexDirection: 'column', height: '85vh' },
```
KristinaSpasevska, line 4: We can use sprinkles here for all the values that are already defined

packages/apps/tools/src/components/Global/ModuleExplorer/SidePanel/styles.css.ts
```diff
@@ -0,0 +1,14 @@
...
+]);
+
+export const modulesContainerStyle = style([{ flex: 1, overflow: 'scroll' }]);
+
+export const moduleTitle = style({
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  maxWidth: '18rem',
```
KristinaSpasevska, line 13: Is it possible to use some of the already defined values for sizes from the sprinkles?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const ModuleExplorer = ({
+  modules,
+  openedModules,
+  onModuleClick,
+}: IModuleExplorerProps) => {
+  const results = getModulesMap(modules);
+  return (
+    <div className={containerStyle}>
```
KristinaSpasevska, line 29: Is it possible here to use the Grid from the react-ui lib (Grid.Root & Grid.Item) instead of a regular div?

sanderlooijenga: Unfortunately, I can't, since the Grid only supports columns of equal sizes, which is not the case for me 😁

packages/apps/tools/src/components/Global/ModuleExplorer/SidePanel/styles.css.ts
sanderlooijenga: Will take a look, thanks!

sanderlooijenga: I've refactored it to use values coming from `vars`. The UI/UX still needs to be done for this page, so the values used will most definitely change in the near future.

packages/apps/tools/src/components/Global/ModuleExplorer/SidePanel/styles.css.ts
sanderlooijenga: I've refactored it to use values coming from `vars`. The UI/UX still needs to be done for this page, so the values used will most definitely change in the near future.

KristinaSpasevska (approved):

PR #954, opened 2023-09-25
React UI: Navigation Header Accordion component

timoheddes:
This PR introduces the NavAccordion component as a variant of the Accordion component, using similar structure and logic. Both components have been refactored and are using context to manage state and pass down props to their subcomponents.

[Asana task](https://app.asana.com/0/1205288273408141/1205386230598870/f)

The component consists of four sub components:

**<NavAccordion.Root>** as the parent container
**<NavAccordion.Section>** for each navigational section (collapsable)
**<NavAccordion.Group>** to create sub sections (lists) that group related navigational links (collapsable)
**<NavAccordion.Link>** to create links that can be used to navigate (using our Link component and supporting the `asChild` prop)

_Design and colours have been discussed with and checked by Isa._


packages/libs/react-ui/src/components/NavAccordion/NavAccordion.tsx
```diff
@@ -0,0 +1,54 @@
...
+                ? setOpenSections(
+                    openSections.filter((i) => i !== sectionIndex),
+                  )
+                : setOpenSections(
+                    linked ? [sectionIndex] : [...openSections, sectionIndex],
+                  ),
+          },
+        ),
+      )}
+    </nav>
```
eileenmguo, line 52: Is this the same as the normal accordion component? If so, I think it would be best to reuse that component

packages/libs/react-ui/src/components/NavAccordion/NavAccordionSection.tsx
```diff
@@ -0,0 +1,73 @@
...
+            navAccordionContentClass,
+            navAccordionContentListClass,
+          ])}
+        >
+          {children}
+        </ul>
+      )}
+    </section>
+  );
+};
```
eileenmguo, line 73: Also look the same as the other accordion component. Can we reuse that one?

eileenmguo (commented):
Main feedback is that it looks like we are copying a lot of the logic from the accordion component when I believe we can just use those components instead. Additionally, this is using a lot of cloneElement, but there is a ticket on the board to update the accordion logic to use useContext instead based on recommendations from react. If they are using the same components, we can update this logic in one place.

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -53,8 +56,20 @@ export const accordionContentClass = style([
...
     fontSize: '$base',
+    margin: 0,
     overflow: 'hidden',
+    padding: 0,
     paddingBottom: '$2',
-    paddingTop: '$2',
+  }),
+]);
+
+export const accordionCollapse = style([
```
timoheddes, line 66: This is in preparation for [animating the opening and closing](https://app.asana.com/0/1205693948735844/1205693670996249/f). Rather than only rendering the Accordion content when a section is open, we need to have the content immediately available so that we can calculate its height. For now, we just switch between display `block` and `none` to show and hide.

packages/libs/react-ui/src/components/Link/Link.tsx
```diff
@@ -7,21 +7,21 @@ import type { FC, ReactNode } from 'react';
 import React from 'react';
 
 export interface ILinkProps {
-  href?: string;
-  target?: '_blank' | '_self' | '_parent' | '_top';
+  asChild?: boolean;
+  block?: boolean;
```
eileenmguo, line 11: I see this was added before, but this is actually inconsistent with how we apply styles. I've made a ticket for it: https://app.asana.com/0/1205693948735844/1205687963086439

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.css.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export const navAccordionWrapperClass = style({});
+
+export const navAccordionListClass = style([
+  sprinkles({
+    margin: 0,
+    padding: 0,
+  }),
+  {
+    listStyleType: 'none',
```
eileenmguo, line 13: this can be a part of sprinkles now. It was added recently

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.css.ts
```diff
@@ -0,0 +1,151 @@
...
+]);
+
+export const navAccordionGroupButtonClass = style([
+  sprinkles({
+    alignItems: 'center',
+    marginBottom: 0,
+    textAlign: 'left',
+  }),
+  {
+    justifyContent: 'flex-start',
```
eileenmguo, line 55: you can move this to sprinkles

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.css.ts
```diff
@@ -0,0 +1,151 @@
...
+    paddingBottom: '0 !important',
```
eileenmguo, line 56: Is this what we want? It reduces the interactive area as well

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.css.ts
```diff
@@ -0,0 +1,151 @@
...
+  {
+    selectors: {
+      '&:active, &:hover': {
+        color: `${vars.colors.$negativeSurface} !important`,
+      },
+      '&:hover': {
+        textDecoration: 'underline',
+      },
+      // Top level links
+      'nav > &': {
```
eileenmguo, line 121: nit: I would recommend using the actual className that's used on the nav element if possible to make it more specific

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.tsx
```diff
@@ -0,0 +1,36 @@
...
+  linked?: boolean;
+}
+
+export const NavAccordionRoot: FC<INavAccordionRootProps> = ({
+  children,
+  linked = false,
+}) => {
+  const [openSections, setOpenSections] =
+    useState<NavAccordionState>(initialOpenSections);
+
```
eileenmguo, line 28: I think you can put this in the context file so that this file is simpler and its collocated with the context

packages/libs/react-ui/src/components/NavAccordion/NavAccordionGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const handleClick = (): void => {
+    if (isOpen) {
+      onClose?.();
+    } else {
+      onOpen?.();
+    }
+    setIsOpen(!isOpen);
+    onClick?.();
```
eileenmguo, line 47: I forget. Why do we need a separate onClick handler in addition to the onOpen and onClose?

packages/libs/react-ui/src/components/NavAccordion/NavAccordionLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface INavAccordionLinkProps {
+  active?: boolean;
+  asChild?: boolean;
+  children: ReactNode;
+  deepLink?: boolean;
+  href?: string;
+}
+
+export const NavAccordionLink: FC<INavAccordionLinkProps> = ({
+  asChild,
```
eileenmguo, line 14: This  asChild prop is not being utilized

packages/libs/react-ui/src/components/NavAccordion/NavAccordionSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+      onClose?.();
+    } else {
+      setOpenSections(linked ? [sectionId] : [...openSections, sectionId]);
+      onOpen?.();
+    }
+    onClick?.();
+  };
+
+  return (
+    <section className={classNames([accordionSectionClass])}>
```
eileenmguo, line 60: nit: don't need to use classNames here

packages/libs/react-ui/src/components/NavAccordion/NavAccordionSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+      />
+
+      {children && (
+        <ul
+          className={classNames(navAccordionListClass, [
+            isOpen ? accordionExpand : accordionCollapse,
+          ])}
+        >
+          {Children.map(children, (child) => {
+            const Tag = child.type === NavAccordionGroup ? 'ul' : 'li';
```
eileenmguo, line 75: nit: Tag confused me because I thought this was adding the Tag component here. Maybe `Element` would be a clearer name. Also, could we move these tags into the components themselves to avoid mapping through children?

packages/libs/react-ui/src/components/Accordion/AccordionHeading.tsx
```diff
@@ -0,0 +1,45 @@
...
+  isOpen: boolean;
+  onClick?: () => void;
+  title: string;
+  titleClassName?: string;
+}
+
+export const AccordionHeading: FC<IAccordionHeadingProps> = ({
+  className,
+  icon,
+  iconClassName = '',
```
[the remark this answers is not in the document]
timoheddes: This was old, forgot to change this heading component

packages/libs/react-ui/src/components/NavAccordion/NavAccordionGroup.tsx
timoheddes: We don't. It's more of an extra to offer more flexibility. What do you think, shall I remove it?

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
timoheddes: Alright, let's tackle this problem in the task for it then. I'll remove this for now.

packages/libs/react-ui/src/components/Accordion/AccordionSection.tsx
```diff
@@ -1,21 +1,21 @@
...
 
-import { SystemIcon } from '@components/Icon';
 import classNames from 'classnames';
 import type { FC } from 'react';
-import React from 'react';
+import React, { useContext } from 'react';
 
 export interface IAccordionSectionProps {
   children?: React.ReactNode;
   index?: number;
```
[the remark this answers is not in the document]
timoheddes: It isn't used at all actually, I'll remove it

packages/libs/react-ui/src/components/Link/Link.tsx
timoheddes: I think I only re-ordered this list. Unrelated to this PR otherwise.

packages/libs/react-ui/src/components/NavAccordion/NavAccordion.tsx
timoheddes: I think this is the usual way of doing it unless you want to move to using actions and reducers which seems overly complex for what we're doing here.

eileenmguo: ahh ok, I think I'm thinking of that indeed. This looks good 👍

packages/libs/react-ui/src/components/Accordion/AccordionHeading.tsx
```diff
@@ -0,0 +1,37 @@
...
+  accordionToggleIconClass,
+} from './Accordion.css';
+
+import { SystemIcon } from '@components/Icon';
+import classNames from 'classnames';
+import type { FC } from 'react';
+import React from 'react';
+
+export interface IAccordionHeadingProps {
+  icon: keyof typeof SystemIcon;
```
eileenmguo, line 13: In what instance would the icon change? The rotation only works for the arrow or the + so I don't think we want to give them the option to use any icon. I would just stick with a fixed + for now

timoheddes: It's because I originally wanted to re-use the heading for the NavAccordion group headings but too many little things differ. I think I'll just get rid of the whole Heading component

timoheddes: Done

packages/libs/react-ui/src/components/NavAccordion/NavAccordionGroup.tsx
```diff
@@ -0,0 +1,76 @@
...
+import classNames from 'classnames';
+import type { FC, FunctionComponentElement } from 'react';
+import React, { Children, useState } from 'react';
+
+export interface INavAccordionGroupProps {
+  children: FunctionComponentElement<INavAccordionLinkProps>[];
+  index?: number;
+  onClose?: () => void;
+  onOpen?: () => void;
+  onClick?: () => void;
```
eileenmguo, line 24: cleanup?

packages/libs/react-ui/src/components/NavAccordion/NavAccordionGroup.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (isOpen) {
+      onClose?.();
+    } else {
+      onOpen?.();
+    }
+    setIsOpen(!isOpen);
+  };
+
+  return (
+    <span>
```
eileenmguo, line 45: should this be a div?

packages/libs/react-ui/src/components/NavAccordion/NavAccordionSection.tsx
```diff
@@ -0,0 +1,86 @@
...
+      );
+      onClose?.();
+    } else {
+      setOpenSections(linked ? [sectionId] : [...openSections, sectionId]);
+      onOpen?.();
+    }
+  };
+
+  return (
+    <section className={accordionSectionClass}>
```
eileenmguo, line 58: nit: Looking into accessibility, I'm not sure this is the right element to use here. I think a div might make more sense, but I don't have a strong preference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

timoheddes: I added a comment to this ticket: https://app.asana.com/0/1205693948735844/1205687963086449/f. I think when we pick that up it would be a good one to also check this and change it if necessary. What do you think?

eileenmguo (approved):

PR #961, opened 2023-09-26
Move `.kadena/pactjs-generated` one level up, out of `node_modules`

webpro:
- Generated files shouldn't be hidden in `node_modules`, but be more visible instead + eligible for (different kind of) caching
- Might caused race conditions during format/build tasks

Breaking change in `tsconfig.json`:

Before:

```json
{
  "compilerOptions": {
    "types": [".kadena/pactjs-generated"]
  },
}

```

After:

```json
{
  "compilerOptions": {
    "types": []
  },
  "include": [".kadena/pactjs-generated"]
}
```


alber70g:
> Note that TypeScript checks both the current dir and `node_modules` for the `types` so I expect no issues here.

With this comment you're saying that it's not a breaking change?

Can we add a check for the user that they have to add the `"include": ...` section to the `tsconfig.json` when currently the `tsconfig.json#compilerOptions.types.includes('.kadena/pactjs-generated')`

alber70g (changes requested):
Instead of duplicating generated code, in both `./.kadena` and `./node_modules/.kadena/pactjs-generated`, I recommend adopting a fix-forward strategy. 

We should empower the user with the ability to designate a custom directory by adding a `--out <directory>` option into the CLI.

This approach ensures that individuals encountering issues with the `./node_modules/.kadena/pactjs-generated` package can gracefully choose to opt out, while allowing others to continue using the base functionality. 

An additional advantage is that this adjustment won't needlessly disrupt established workflows.

Randynamic:
Why Is it a bad thing to introduce a breaking change?

alber70g:
> Why Is it a bad thing to introduce a breaking change?

It's not, but when it's _unnecessary_ it's better to _improve_ the current state by adding a feature.

Current situation:
default output directory `./node_modules/.kadena/pactjs-generated`

New situation:
customizable output directory with `--out`, when not supplied, use the default output directory `./node_modules/.kadena/pactjs-generated`

webpro:
The point of this PR, as described in the initial two bullet points, is getting rid of that default behavior altogether. It's not a good practice and it's most likely causing race conditions. So a breaking change is necessary. Adding a feature would make the situation even worse.

alber70g:
> The point of this PR, as described in the initial two bullet points, is getting rid of that default behavior altogether. It's not a good practice and it's most likely causing race conditions. So a breaking change is necessary. Adding a feature would make the situation even worse.

I agree, a breaking change for this is necessary, and it's the right thing to do, but not right now. It doesn't benefit the community to introduce a breaking change this early in the release of 1.0.0. 

We should take the time to iron out the bugs that can be fixed in minor and patch releases, so the consumers _automatically_ enjoy them (considering that most consumers use `^` as their semver indicator), and after a while we can introduce it, with other breaking changes that we want to make, as one major version upgrade.

Randynamic:
@alber70g when are we going to merge this one?

Randynamic:
https://app.asana.com/0/1206899464980558/1208226679158102

PR #962, opened 2023-09-26
fix: remove vscode workspace settings to allow custom definitions

barthuijgen:
The workspace guidelines suggest setting up your editor environment yourself, but having this settings file checked into git goes against that. Unless someone has a good solutions I'm afraid vscode does not give it to you both ways, either you provide a well rounded configuration that works for everyone, or you have no configuration checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,9 @@ packages/*/*/temp
 .idea
 .idea/*
 
+# Vscode files
+.vscode/settings.json
+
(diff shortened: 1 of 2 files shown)


barthuijgen:
Closed with the alternative solution to use a git ignore path from git global config.
Also suggested changes to the settings in another PR: https://github.com/kadena-community/kadena.js/pull/964

PR #964, opened 2023-09-27
feat: added vscode settings for format on save

barthuijgen:
These are the settings I personally use to try and match the repositories formatting rules.

It uses a vscode extension `format-code-action` to be able to format as part of `editor.codeActionsOnSave` which enables both eslint fix and prettier format. The alternative way to doing this would be the prettier plugin for eslint, but that is slower and requires extra configuration per project.

We could limit format on save per file type, but currently it will run for everything prettier supports, and also run eslint-fix for typescript files.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,16 @@
 {
+  "eslint.format.enable": true,
+  "editor.formatOnSave": true,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "graphql-config.load.rootDir": "packages/apps/graph",
+  "[typescript][typescriptreact]": {
+    "editor.formatOnSave": false,
+    "editor.codeActionsOnSave": [
+      "source.formatDocument",
+      "source.fixAll.eslint"
+    ]
+  },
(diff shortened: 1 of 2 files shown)


barthuijgen:
tl;dr: Don't put .idea and .vscode directories to project's .gitignore. You should configure a global .gitignore for your machine instead.

.vscode/settings.json
```diff
@@ -1,6 +1,16 @@
...
+  "editor.formatOnSave": true,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "graphql-config.load.rootDir": "packages/apps/graph",
+  "[typescript][typescriptreact]": {
+    "editor.formatOnSave": false,
+    "editor.codeActionsOnSave": [
+      "source.formatDocument",
+      "source.fixAll.eslint"
+    ]
+  },
```
alber70g, line 12: I'm not a fan of interrupting workflow for everyone. This is something you can add yourself to your own settings.

I suppose configuring extensions like the graphql one is fine?

PR #969, opened 2023-09-27
[@kadena/graph] feat: Simulate traffic on devnet v2

nil-amrutlal-dept:
**Devnet simulate traffic v2**
- creates accounts on devnet and funds these an equal amount taking into consideration the `tokenPool`
- creates traffic within these accounts, both transfers and cross-chain transfers
- all traffic is random and can be replicated due to the use of seeded random generation
- balance checking
- auto refunding of accounts (to guarantee simulation can run indeterminately)
- error handling
- file record with comprehensive logs
- added option to run fund script without a key (in this case, it will generate one and fun it)

Simulate v2 is a second iteration of the simulated traffic on the devnet. In this version, the main goal was to create a simulation that was possible to replicate (by using a seeded random number) and that depended less on `sender00`, as well as checked balances. The transfer amounts and chains are completely randomized. 

---
  - https://app.asana.com/0/0/1205582598928051


packages/apps/graph/src/scripts/devnet/config.ts
```diff
@@ -2,4 +2,5 @@ export default {
   NETWORK_ID: 'fast-development',
   CHAIN_ID: '0',
   PORT: 8080,
+  NO_CHAINS: 20,
```
alber70g, line 5:
```suggestion
  NUMBER_OF_CHAINS: 20,
```

packages/apps/graph/src/scripts/devnet/helper.ts
```diff
@@ -107,30 +107,12 @@ export const sender00: IAccount = {
...
-  const randomIndex = Math.floor(Math.random() * chainIds.length);
-  return chainIds[randomIndex];
+export const getRandomNumber = (
+  randomNumber: number,
+  maxNumber: number,
+): number => {
+  if (randomNumber > 1 || randomNumber < 0)
+    throw new Error('randomNumber must be less than 1 and greater than 0');
+  const generatedNumber = Math.floor(randomNumber * maxNumber);
+  return generatedNumber === 0 ? 1 : generatedNumber;
```
alber70g, line 117: We might need some docs to understand what `randomNumber` arg means and what it's used for.

alber70g (approved):

packages/apps/graph/src/scripts/devnet/helper.ts
nil-amrutlal-dept: Just added some comments on it to explain its functionality

MRVDH (approved):

PR #970, opened 2023-09-27
feat(tools): add custom network functionality

KristinaSpasevska:


alber70g:
Something's broken on the preview link. Is that only here?

packages/apps/tools/locales/en/common.json
```diff
@@ -161,5 +161,7 @@
   "Retry": "Retry",
   "Search": "Search",
   "XChain transfer has been successfully finalized!": "XChain transfer has been successfully finalized!",
-  "The Faucet only runs on Testnet. The Network toggle in the topbar has no effect on the Faucet. It will not fund accounts on Mainnet, only on Testnet.": "The Faucet only runs on Testnet. The Network toggle in the topbar has no effect on the Faucet. It will not fund accounts on Mainnet, only on Testnet."
+  "The Faucet only runs on Testnet. The Network toggle in the topbar has no effect on the Faucet. It will not fund accounts on Mainnet, only on Testnet.": "The Faucet only runs on Testnet. The Network toggle in the topbar has no effect on the Faucet. It will not fund accounts on Mainnet, only on Testnet.",
```
alber70g, line 164: It should contain something along the lines of not running on Mainnet but on other networks where the faucet is installed

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { zodResolver } from '@hookform/resolvers/zod';
+import useTranslation from 'next-translate/useTranslation';
+import type { FC } from 'react';
+import React, { useState } from 'react';
+import { useForm } from 'react-hook-form';
+import * as z from 'zod';
+
+// @see; https://www.geeksforgeeks.org/how-to-validate-a-domain-name-using-regular-expression/
+const DOMAIN_NAME_REGEX: RegExp =
+  /^(?!-)[A-Za-z0-9-]+([\-\.]{1}[a-z0-9]+)*\.[A-Za-z]{2,6}$/;
```
alber70g, line 19: The network host can be anything. If someone hosted it on a local machine, they could use a hostname, ip-address, or ipv6 as well, which all aren't FQDN. I don't think it should be limited in any way.

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
```diff
@@ -0,0 +1,114 @@
...
+              }}
+            />
+            <TextField
+              label={'Network api'}
+              inputProps={{
+                id: 'api',
+                ...register('api'),
+                onChange: (e) => setApi(e.target.value),
+                value: api,
+                placeholder: 'localhost:8080'
```
alber70g: Even the placeholder doesn't match the regex ;)

packages/apps/tools/src/constants/kadena.ts
```diff
@@ -1,16 +1,15 @@
...
 export type DevOption = 'BASIC' | 'BACKEND' | 'DAPP';
 
 type KadenaConstants = {
-  [K in Network]: {
+  [K in DefinedNetwork]: {
     label: string;
     API: string;
     apiHost: (params: { networkId: string; chainId: string }) => string;
-    estatsHost: () => string;
+    estatsHost?: () => string;
```
alber70g, line 12: Shouldn't this include the arguments?
```suggestion
    estatsHost?: ({ networkId: string, chainId: ChainId }) => string;
```

Why would this be optional? In the devnet you can reach it on `<host>/txs/account/k:554754f48b16df24b552f6832dda090642ed9658559fef9f3ee1bb4637ea7c94?limit=1000`

packages/apps/tools/src/utils/network.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const isNetwork = (x: any): x is Network =>
+  ['mainnet01', 'testnet04'].includes(x);
+
+export const getConfigNetworkNames = (): DefinedNetwork[] => {
+  return Object.keys(kadenaConstants).filter((key) =>
+    isNetwork(key),
+  ) as DefinedNetwork[];
+};
```
alber70g, line 23: Does it only include predefined networks or should it include custom networks here?

alber70g (approved):
Minor comments. I did only do a static code review

KristinaSpasevska:
> Something's broken on the preview link. Is that only here?

I updated a fix, you should be able to see the preview now

packages/apps/tools/src/utils/network.ts
KristinaSpasevska: Here I'm only fetching the predefined config networks and in the /utils/network.ts all networks are handled

packages/apps/tools/src/services/accounts/get-transactions.ts
```diff
@@ -26,17 +26,22 @@ export async function getTransactions(options: {
   account: string;
 }): Promise<ITransaction[]> {
   debug(getTransactions.name);
-
   const { network, chain, account } = options;
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
ash-vd, line 30: This can be removed

packages/apps/tools/src/services/accounts/get-transactions.ts
```diff
@@ -26,17 +26,22 @@ export async function getTransactions(options: {
...
-    const result: ITransaction[] = (await fetch(
-      `https://${getKadenaConstantByNetwork(
-        network,
-      ).estatsHost()}/txs/account/${account}?token=coin&chain=${chain}&limit=10`,
-    ).then((res) => res.json())) as ITransaction[];
+  const networkDto = networksData.find((item) => item.networkId === network);
 
-    return result;
+  if (!networkDto) {
+    // @ts-ignore
```
ash-vd, line 36: Can we add `null` as a returntype and explicitly return `null` here?

packages/apps/tools/src/services/accounts/get-transactions.ts
```diff
@@ -26,17 +26,22 @@ export async function getTransactions(options: {
...
+    return;
+  }
+
+  try {
+    // @ts-ignore
```
ash-vd, line 41: I don't think this is needed anymore

packages/apps/tools/src/services/modules/describe-module.ts
```diff
@@ -24,18 +21,27 @@ export const describeModule = async (
   ttl: number = kadenaConstants.API_TTL,
 ): Promise<ILocalCommandResult> => {
   debug(describeModule.name);
-  const networkId = chainNetwork[network].network;
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
ash-vd, line 24: Can be removed

packages/apps/tools/src/services/modules/list-module.ts
```diff
@@ -26,18 +23,27 @@ export const listModules = async (
   ttl: number = kadenaConstants.API_TTL,
 ): Promise<IModulesResult> => {
   debug(listModules.name);
-  const networkId = chainNetwork[network].network;
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
ash-vd, line 26: Can be removed

packages/apps/tools/src/services/modules/list-module.ts
```diff
@@ -26,18 +23,27 @@ export const listModules = async (
...
+  const { networksData } = useWalletConnectClient();
+
+  const networkDto = networksData.find((item) => item.networkId === network);
+
+  if (!networkDto) {
+    // @ts-ignore
```
ash-vd, line 32: Return `null` and change the type

packages/apps/tools/locales/en/common.json
KristinaSpasevska: @alber70g can you please let me know what the correct message should be?

packages/apps/tools/src/constants/kadena.ts
KristinaSpasevska: estatsHost is now included as a prop and is not optional 👍

packages/apps/tools/locales/en/common.json
alber70g:
```suggestion
  "The Faucet is not available on Mainnet. On other networks, the Faucet smart contract must be deployed to fund accounts. In the Module Explorer you can see if it's deployed: https://tools.kadena.io/transactions/module-explorer?module=user.coin-faucet&chain=1": "The Faucet is not available on Mainnet. On other networks, the Faucet smart contract must be deployed to fund accounts. In the Module Explorer you can see if it's deployed: https://tools.kadena.io/transactions/module-explorer?module=user.coin-faucet&chain=1",
```

Can we add a URL here? If so, can we add a link to the module-explorer: The Faucet is not available on Mainnet. On other networks, the Faucet smart contract must be deployed to fund accounts. In the Module Explorer you can see if it's deployed: https://tools.kadena.io/transactions/module-explorer?module=user.coin-faucet&chain=1

alber70g (approved):

alber70g:
I looked at the preview URL, but it doesn't show the predefined networks anymore

KristinaSpasevska:
> > Something's broken on the preview link. Is that only here?
> 
> I see messages in the console that the origin isn't allowed, the WalletConnect modal loads but it takes some time for me.
> 
> @KristinaSpasevska What WalletConnect Project ID are we using here?

Previously we were using Reza's personal project ID, but I have no idea whether that is changed. If you have the correct one to be used here, we can change it and make sure we're using the correct one.

packages/apps/tools/src/services/accounts/get-transactions.ts
KristinaSpasevska: This still gives me this error: 
ESLint: React Hook "useWalletConnectClient" is called in function "describeModule" that is neither a React function component nor a custom React Hook function.

In order not to use this here this way, maybe I can import the `networksData` into the component and give it as a parameter to this function? Or just supress the warning as I have done? Please let me know
@ash-vd

sanderlooijenga:
I think we now also need to update the `showNotification` [check](https://github.com/kadena-community/kadena.js/blob/main/packages/apps/tools/src/pages/faucet/existing/index.tsx#L125) in the Faucet. I guess invert the condition to "not equals testnet"?

packages/apps/tools/src/components/Global/AddNetworkModal/styles.css.ts
```diff
@@ -0,0 +1,25 @@
+import { sprinkles } from '@kadena/react-ui/theme';
+
+import { style } from '@vanilla-extract/css';
+
+export const modalOptionsContentStyle = style([
+  sprinkles({
+    width: '100%',
+    fontSize: '$xs',
```
sanderlooijenga, line 8: It looks kind of small to me, not sure if this is by design? 😁

packages/apps/tools/src/constants/kadena.ts
```diff
@@ -1,12 +1,11 @@
-import type { ChainwebNetworkId } from '@kadena/chainweb-node-client';
-
 import { env } from '@/utils/env';
 
-export type Network = Exclude<ChainwebNetworkId, 'development'>;
+export type DefinedNetwork = 'mainnet01' | 'testnet04';
```
sanderlooijenga, line 3: Out of curiosity; why can't we use `ChainwebNetworkId` anymore?

packages/apps/tools/src/pages/faucet/existing/index.tsx
sanderlooijenga, line 1: `const showNotification = selectedNetwork !== 'tesnet04';`?

packages/apps/tools/src/services/accounts/get-transactions.ts
sanderlooijenga: The error is correct, I would do as you suggested Kristina, provide it as a parameter 👍

packages/apps/tools/src/utils/network.ts
```diff
@@ -0,0 +1,79 @@
+import type { DefinedNetwork, Network } from '@/constants/kadena';
+import { kadenaConstants } from '@/constants/kadena';
+
+interface IApiHostData {
+  networkId: string;
+  chainId: string;
+}
+
+export interface INetworkDto {
```
sanderlooijenga, line 9: What does `dto` stand for? Can't we just be explicit and write it out fully? 😁

packages/apps/tools/src/services/cross-chain-transfer-finish/get-transfer-data.ts
```diff
@@ -65,10 +65,22 @@ export async function getTransferData({
...
-      const host = getKadenaConstantByNetwork(network).apiHost({
-        networkId: chainNetwork[network].network,
+      const { networksData } = useWalletConnectClient();
+
+      const networkDto = networksData.find(
+        (item) => item.networkId === network,
+      );
+
+      if (!networkDto) {
+        // @ts-ignore
```
sanderlooijenga, line 75: What does TS otherwise complain about?

packages/apps/tools/src/utils/network.ts
```diff
@@ -0,0 +1,79 @@
...
+export const getAllNetworks = (
+  localStorageNetworks: INetworkDto[],
+): INetworkDto[] => {
+  const allNetworkObjects: INetworkDto[] = [];
+  const configNetworks = getConfigNetworkNames();
+
+  const configNetworksAdded =
+    Boolean(localStorageNetworks) &&
+    localStorageNetworks.find(
+      (item) => item.label === 'Mainnet' || item.label === 'Testnet',
```
sanderlooijenga, line 35: Maybe we can export the `label`s from the kadena constants file and compare it with that (to make it somewhat more resilient)?

packages/apps/tools/src/constants/kadena.ts
KristinaSpasevska: I think @alber70g would know the answer to this :)

trendzetter (commented):
`* feat: add estats as a required prop`

@KristinaSpasevska @alber70g Making estats a requirement raises the the requirements a lot, running chainweb-data is requires much resources on any significant network besides an empty devnet. I think this hasn't been well considered from a practical point of view. Plus, I think this probably would work also as an optional property. At least the underlying network doesn't require chainweb-data so it would be a requirement purely from the lib. Would it work with just inputting a dummy non-existing chainweb-data node?
See also earlier similar featback for chainweb-stream: https://github.com/kadena-io/chainweb-stream/issues/2

PR #976, opened 2023-10-03
Improve tooling config + add docs

webpro:
Improve, fix and document a few things around build, lint + format:

* Fix eslint-config build output config (caching) (a495ce2c)
* Remove duplicate config (it's already in base config) (719c73b1)
* Run format scripts sequentially (to prevent race condition during file writes) (a0a34122)
* Add some docs about linting and formatting (d820cfaa)
* Format a file (c7c1262c)
* Keep formatting docs/*.md (515564e0)


docs/lint-vs-format.md
```diff
@@ -0,0 +1,87 @@
...
+> Although lint rules can enforce some formatting preferences in code, they're
+> slower, less consistent, and more prone to missing edge cases when compared to
+> dedicated formatters.
+
+Another reason is that it should be possible to replace either ESLint or
+Prettier without them being too intermingled.
+
+This separation of concerns leads to clearer objectives and faster execution.
+For example, if you want to know whether your code is formatted properly you can
+do `pnpm lint` and get your answer faster compared to doing everything at once.
```
JoshuaKGoldberg, line 23: 😛 

```suggestion
do `pnpm format` and get your answer faster compared to doing everything at once.
```

alber70g (approved):

docs/lint-vs-format.md
```diff
@@ -0,0 +1,87 @@
...
+
+When looking into the related scripts in `package.json` you'll find something
+like this:
+
+```json
+{
+  "name": "@kadena/pkg",
+  "scripts": {
+    "format": "pnpm run --sequential /^format:.*/",
+    "format:lint": "pnpm run lint:src --fix",
```
JoshuaKGoldberg, line 51: Just popping in un-asked-for: I personally have always found it confusing when commands mix "format" and "lint". A lot of devs already have a hard time disambiguating between the two. I'd suggest either calling this `lint:src:fix` or just omitting it (because all it does is add `--fix`.

docs/lint-vs-format.md
```diff
@@ -0,0 +1,87 @@
...
+_tl/dr; Use `pnpm format` for formatting and `pnpm lint` for linting._
+
+When looking into the related scripts in `package.json` you'll find something
+like this:
+
+```json
+{
+  "name": "@kadena/pkg",
+  "scripts": {
+    "format": "pnpm run --sequential /^format:.*/",
```
JoshuaKGoldberg, line 50: Ooh TIL `pnpm run` takes in regexes, that's awesome!

docs/lint-vs-format.md
webpro: Thanks for popping in, it's appreciated. Naming is hard 🥲

- Renaming to `lint:src:fix` means losing the `format:*` convention, and with that the `/^format:.*/`, in turn resulting in having to manually keep track of them in each package separately.
- Omitting it would mean duplication of something that could change, and lead to subtle bugs as the other script includes two arguments that might change, namely the dir(s) to scan and the extensions (which unfortunately can't be configured elsewhere, like in ESLint config files). They're different across the workspaces. And change rarely, but still.

Since developers only really need to remember `format` and `lint` I thought the existing "mechanical" solution is easier/safer to maintain long-term. Still happy to be convinced otherwise, but these were my considerations.

sanderlooijenga (approved):

PR #977, opened 2023-10-03
[@tools-website] Fixes in module-explorer and estats host in accounts

KristinaSpasevska:


packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
```diff
@@ -25,8 +25,8 @@ export const getQueryValue = (
 export const getCookieValue = (
   needle: string,
   haystack: NextApiRequestCookies,
-  defaultValue?: string,
-): string | null => {
+  defaultValue?: string | [any],
```
sanderlooijenga, line 28: Why does the `defaultValue` need to be changed?

packages/apps/tools/src/utils/network.ts
```diff
@@ -1,17 +1,25 @@
+import type { ChainwebChainId } from '@kadena/chainweb-node-client';
+
 import type { DefinedNetwork, Network } from '@/constants/kadena';
 import { kadenaConstants } from '@/constants/kadena';
 
 interface IApiHostData {
+  api: string;
   networkId: string;
   chainId: string;
```
sanderlooijenga, line 9: Also should to be of type `ChainwebChainId`?

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
```diff
@@ -15,9 +15,9 @@ import { useForm } from 'react-hook-form';
 import * as z from 'zod';
 
 const schema = z.object({
-  label: z.string().trim(),
-  networkId: z.string().trim(),
-  api: z.string().trim(),
+  label: z.string().trim().min(1),
+  networkId: z.string().trim().min(1),
+  api: z.string().trim().min(1),
```
sanderlooijenga, line 20: You could perhaps reuse [this](https://github.com/kadena-community/kadena.js/blob/main/packages/apps/tools/src/pages/transactions/cross-chain-transfer-finisher/index.tsx#L65) for validating this field?

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
```diff
@@ -15,9 +15,9 @@ import { useForm } from 'react-hook-form';
 import * as z from 'zod';
 
 const schema = z.object({
-  label: z.string().trim(),
-  networkId: z.string().trim(),
-  api: z.string().trim(),
+  label: z.string().trim().min(1),
```
sanderlooijenga, line 18: Do we have any more specific requirements for the input (instead of just minimal 1 character)?

packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
KristinaSpasevska: We return the array of networks here, so if they're not in the cookie, the default will be an empty array

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
KristinaSpasevska: it will fail for custom networks (localhost:8080 or something)

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
KristinaSpasevska: Not than I know of, let me know if you think some more should be added

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
sanderlooijenga: Ah you're right, that wouldn't work, my bad

packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
sanderlooijenga: But the `defaultValue` is not being given as a parameter for the networks cookie retrieval, right?

packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
```diff
@@ -25,8 +25,8 @@ export const getQueryValue = (
...
+): string | [any] | [] | null => {
```
sanderlooijenga, line 29: Btw what is the expected allowed return type here? An array (tuple) with one item of type `any`, or should it be `Array<any>` (or `any[]`).

Sidenote; I would just like for TS to infer these types, so ESLint giving us an error for not explicitly declaring the return type. What do you think?

packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
KristinaSpasevska: Ah, yeah... I removed the default value in the params. I will remove it here as well

packages/apps/tools/src/pages/transactions/module-explorer/utils.ts
KristinaSpasevska: Do you think it would be better to add the type here Array<INetworkData> so we avoid the TS errors? We can add this as a possible return type

sanderlooijenga: You know what might be better; since the defaultValue could be `any`thing, maybe it's better to make it a generic and let TS infer the return type itself. What do you think?

KristinaSpasevska: I agree and it will better be a generic

sanderlooijenga: FYI; made it generic 👌

sanderlooijenga (approved):

PR #980, opened 2023-10-04
chore: add github action for ingesting docs data to Algolia

realdreamer:
This PR adds the workflow to push the docs website datasets to publish into Algolia search to perform semantic/keywords search in docs website.


sstraatemans (approved):

webpro:
Two remarks:

- This incurs costs every merge to main with a change in `packages/apps/docs/**`. I believe the costs are low and it's probably not worth further optimizations, but probably good to have some rough idea of costs x frequency?
- Is the Algolia service fine with re-sending the same data multiple times? Does it basically replace everything properly to prevent (near) duplicates?

sstraatemans:
Looking at the costs of algolia. there are no costs for sending the data

realdreamer:
> * This incurs costs every merge to main with a change in `packages/apps/docs/**`. I believe the costs are low and it's probably not worth further optimizations, but probably good to have some rough idea of costs x frequency?
Looking at the Algolia pricing model, it doesn't have any rate limit for updates. It's mainly depends on number of records and search requests

> * Is the Algolia service fine with re-sending the same data multiple times? Does it basically replace everything properly to prevent (near) duplicates?
Algolia updates relies on `objectID` property this property we're generating based on file name in 7 docs. if the ids are already exists in the index then it only performs partial/no updates

webpro:
> > * This incurs costs every merge to main with a change in `packages/apps/docs/**`. I believe the costs are low and it's probably not worth further optimizations, but probably good to have some rough idea of costs x frequency?
> 
> Looking at the Algolia pricing model, it doesn't have any rate limit for updates. It's mainly depends on number of records and search requests

Sorry, I wasn't clear. I was mainly referring to using the [OpenAI create embedding endpoint](https://platform.openai.com/docs/guides/embeddings/what-are-embeddings) to create the vectors: https://openai.com/pricing says $0.0001 / 1K tokens 

Thinking about this... do we even need those? We could skip that part and only use the parsed documents but not the embeddings?

.github/workflows/docs-ingestion.yml
```diff
@@ -0,0 +1,33 @@
...
+      - packages/apps/docs/**
+
+jobs:
+  ingestion:
+    name: Ingestion docs website data to Algolia
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history, needed to determine diffs.
```
Ghislain89, line 17: Do you really need fetch depth 0? This means all history is included, while you likely only need the latest state?

realdreamer: yeah, only needs the latest state

webpro (approved):

PR #982, opened 2023-10-04
React UI: allow nested arrays of restricted types as children for Table subcomponents

timoheddes:
The subcomponents of the Table component (Body, Head and TR) are restricting the type of children you can pass to them. This PR adds React Fragments to be valid children.

[Asana Link](https://app.asana.com/0/1205553603119741/1205648128915723/f)


MRVDH (approved):

eileenmguo (commented):
I think this allows users to wrap anything in a fragment to pass in as a child. I would recommend trying to fix the type instead of allowing fragments. 

Maybe updating the type to something like this would work? 
```
export type CompoundType<T> = Tree<FunctionComponentElement<T>>;

type Tree<T> = T | Array<Tree<T>>;
```

Essentially allowing users to pass in nested arrays of the same type

eileenmguo (approved):

PR #984, opened 2023-10-04
feat: immutable records according to designs (wip)

barthuijgen:
This pull request adds work in progress on immutable records app. I want to make this PR now to avoid it becoming even larger.


packages/apps/immutable-records/src/screens/Home/Home.tsx
```diff
@@ -1,7 +1,77 @@
...
+                key={day}
+                day={day}
+                available={availableNfts.includes(day)}
+                active={day === activeBidding}
+                progress={50}
+              />
+            ))}
+          </div>
+        </main>
+        <div className={yearsSidebar}>
```
sstraatemans, line 61: een <aside> tag?

packages/apps/immutable-records/src/screens/Home/Home.css.ts
```diff
@@ -2,11 +2,122 @@ import { sprinkles } from '@kadena/react-ui/theme';
...
+    fontWeight: 400,
+    letterSpacing: '-0.025rem',
+    lineHeight: '1.5rem',
+    margin: '0 0 1.3rem 0',
+  },
+]);
+
+export const main = style([
   sprinkles({
     margin: '$4',
```
sstraatemans, line 82: hm..als je overal gewoon 'px' gebruikt. zou ik dat bij deze denk ik ook doen.
je zit zover van de react-ui af, dan zou ik het eigenlijk helemaal niet gebruiken.
anders veranderd opeens deze variabele en klopt er niets van je design.

ik zou wel aan vanilla extract vasthouden, want dat gebruiken we nu overal

packages/apps/immutable-records/public/fonts/KodeMono.woff2
sstraatemans, line 1: isa heeft een npm package '@kadena/fonts' ofzo. daar kun je de kodemono mee gebruiken in css.
het zit ook al in react-ui (font-family: 'mono')

sstraatemans (approved):

packages/apps/immutable-records/public/fonts/KodeMono.woff2
barthuijgen: I checked out trying to use `@kadena/fonts` together with Isa, but it does not work well with next.js appdir. It is not able to load the font until javascript is executed causing it to pop-in after pageload. Using nextjs's font helper gives the best performance, but it was not compatible with the `@kadena/fonts` package.

packages/apps/immutable-records/.eslintrc.cjs
```diff
@@ -5,6 +5,20 @@ module.exports = {
...
+    '@rushstack/no-new-null': 'off',
+    // The official docs recommend against it: https://typescript-eslint.io/rules/typedef/
+    // Typescript is better at inferring and using `noImplicitAny` is sufficient
+    '@typescript-eslint/typedef': 'off',
     '@kadena-dev/typedef-var': 'off',
+    // This rule is too strict for our use case
+    '@typescript-eslint/strict-boolean-expressions': 'off',
+    // TODO: eslint in vscode does not understand tsconfig paths and gives errors
+    //       even though in the build it works fine
+    'import/no-unresolved': 'off',
```
webpro, line 22: Looks like a good opportunity to have something of a shared `app` ESLint config, or extend the existing `next` config.

Those profiles are here: https://github.com/kadena-community/kadena.js/tree/main/packages/tools/eslint-config/profile

This specific `import/no-unresolved` should not need to be disabled, this works well in the other packages too?

packages/apps/immutable-records/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "rootDir": "./",
     "types": ["node", "@types/gtag.js"],
+    "lib": ["ES2022", "DOM"],
```
webpro, line 6: See comment about shared profiles for ESLint.

webpro (commented):
Small-ish PRs are great, but maybe still give a short description what is included?

packages/apps/immutable-records/.eslintrc.cjs
barthuijgen: But I'm already extending `next` via `@kadena-dev/eslint-config/profile/next` right?
Also `next` extends `react` and `react` extends `lib`, so when I use `next` I automatically opt-in to all the `lib` rules too

barthuijgen: Found the solution, you have to configure `"eslint.workingDirectories": ["./packages/apps/immutable-records"]` in `./vscode/settings.json` 
Where I just have this app now, but ideally any directory that has a `.eslintrc.*` is added, really don't know why they don't just resolve these configs automatically but okay, this does work.

packages/apps/immutable-records/package.json
```diff
@@ -39,6 +47,8 @@
     "@vanilla-extract/next-plugin": "2.3.1",
     "eslint": "^8.45.0",
     "eslint-config-next": "13.4.5",
+    "lokijs": "^1.5.12",
```
webpro, line 50: What is this dependency being used for?

packages/apps/immutable-records/package.json
```diff
@@ -39,6 +47,8 @@
...
+    "pino-pretty": "^10.2.0",
```
webpro, line 51: What is this dependency being used for?

packages/apps/immutable-records/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  open: boolean;
+  onClose: () => void;
+}
+
+export const Modal: FC<PropsWithChildren<ModalProps>> = ({
+  children,
+  open,
+  onClose,
+}) => {
+  useEventListener('keydown', (e) => {
```
webpro, line 16: Do we really need another dependency to use `addEventListener`?

packages/apps/immutable-records/package.json
barthuijgen: `lokijs` and `pino-pretty` were both required by `@walletconnect/modal` and `@walletconnect/sign-client`, it gives build errors when I did not add these, seems like a bug on their side. It's dev-dependencies anyway, I don't think its a big deal right now.

packages/apps/immutable-records/package.json
barthuijgen: see comment on `lokijs`, these two have the same reason.

packages/apps/immutable-records/src/components/Modal/Modal.tsx
barthuijgen: `usehooks-ts` provides many useful utilities for react apps at a 4kb gzipeed size. I don't think there should be much reason to have to justify a package of that size if it makes us have to write and maintain less code.

webpro (approved):

PR #988, opened 2023-10-05
fix(docs): side menu sticky on desktop + mobile

realdreamer:
- Make side menu sticky on inner pages
- Landing page make it scrollable until header after that make it sticky
- Fix mobile view menu scroll only menu view not the body of the page


packages/apps/docs/src/components/Layout/components/Menu/menu.css.ts
```diff
@@ -1,19 +1,21 @@
...
-    position: 'absolute',
-    paddingBottom: '$40',
+    position: 'fixed',
     height: '100%',
     width: '100%',
     background: '$background',
     overflow: 'hidden',
   }),
   {
+    top: vars.sizes.$17,
```
sstraatemans, line 16: these are not sprinkles?

realdreamer: yes, I think so! but I'll double check now

sstraatemans (approved):

packages/apps/docs/src/components/Layout/components/Menu/menu.css.ts
realdreamer: sorry! my bad! it's part of sprinkles. updated it

PR #989, opened 2023-10-05
[@kadena/graph] , [@kadena/graph-client] Fix - Eventparameters fix and small adjustments

nil-amrutlal-dept:
Due to uncertainty on the type of data each parameter can have, the `EventParameters` as an `array` were causing errors. This was changed and now a `json` string, named `paramtext` is the output. 
It is now up to the API consumer to parse the data

---
  - https://app.asana.com/0/0/1205649860913253


MRVDH (approved):

packages/apps/graph/src/graph/Subscription/newBlocks.ts
```diff
@@ -28,13 +29,14 @@ async function* iteratorFn(
...
   ),
+  context: IContext,
 ): AsyncGenerator<Block[], void, unknown> {
   let lastBlock = (await getLastBlocks(chainIds))[0];
 
   yield [lastBlock];
   log('yielding initial block with id %s', lastBlock.id);
 
-  while (true) {
+  while (!context.req.socket.destroyed) {
```
nil-amrutlal-dept, line 39: Noticed that the loop continues if no block has been fetched and the subscription ended

webpro (approved):

PR #996, opened 2023-10-05
First setup for Tools E2E tests

Ghislain89:
This PR adds
* Shared Playwright Configuration, to be consumed in dApps
* E2E tests with limited mocking for tools dApp. 
* Shared Page Objects for re-usable components.
* GitHub Actions for these tests


.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,55 @@
...
+        run: pnpm install
+
+      - name: Install browsers
+        run: pnpm exec playwright install chromium
+        working-directory: packages/apps/tools
+
+      - name: Build Tools dApp
+        run: pnpm --filter @kadena/tools...  build
+        env:
+          FAUCET_PUBLIC_KEY: 782127638ab9cc8fa8598ff0120464ceef6f367ddcd616b47afffbdb175dcc5e
```
Ghislain89, line 41: We can potentially fetch these from github environment variables, but they are dummy values so it might be too much overhead.

packages/tools/heft-rig/tsconfig.playwright.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "baseUrl": "packages",
+    "paths": {
+      "@kadena/*": ["packages/tools/e2e-components"] // This mapping is relative to "baseUrl".
+    }
+  }
+}
```
webpro, line 8: Why do we need this path mapping here? Mapping like looks odd, since we have packages with the same name. Maybe they should just be (dev/peer) dependencies?

packages/tools/e2e-components/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@kadena/e2e-components",
```
webpro, line 2: This should probably be `@kadena-dev/e2e-components` and `private: true`?

webpro (commented):
The structure looks a bit odd to me: I see `e2e-tests` inside `apps/tools`, but `e2e-components/shared-components` in tools. And, "tools"?

.github/workflows/e2e-tests.yaml
Ghislain89: I'll add a comment. Unfortunately the _values_ need to be syntactically correct so I can't just put 'dummy-value' as a value.

packages/tools/heft-rig/tsconfig.playwright.json
Ghislain89: I could not get that to work fully, I'll revisit and ask for help if I get stuck.

packages/tools/e2e-tests/README.MD
```diff
@@ -0,0 +1 @@
+TODO
```
webpro, line 1:
```suggestion
<!-- genericHeader start -->

<!-- genericHeader end -->
```

This + `pnpm format:md` is a start (not necessarily in this PR)

packages/tools/e2e-tests/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@kadena-dev/e2e-tests",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "format": "pnpm run --sequential /^format:.*/",
+    "format:lint": "pnpm run lint:src --fix",
+    "format:md": "remark README.md -o --use @kadena-dev/markdown",
+    "format:src": "prettier . --cache --write",
+    "install:chromium": "npx playwright install chromium",
```
webpro, line 10: I think you can omit `npx`, it should be available as a binary in `$PATH`. Package managers add `node_modules/.bin` to `$PATH` automatically.

packages/tools/e2e-tests/src/fixtures/mocks/send.ts
```diff
@@ -0,0 +1,3 @@
+export const send: object = {
```
webpro, line 1:
```suggestion
export const send: Record<string, string[]> = {
```

Might not be good/specific enough, but `object` is usually too loose. Oh well, this is just a mock. And then we are forced to type stuff when declaring vars, lol. This can be turned off in ESLint settings (maybe just override for mock folders). Less work to just import the actual type if we have it?

packages/tools/e2e-tests/src/page-objects/react-ui/card.component.ts
```diff
@@ -0,0 +1,24 @@
+import type { Page } from '@playwright/test';
+
+export default class CardComponent {
+  private readonly _page: Page;
+
+  public constructor(page: Page) {
+    this._page = page;
+  }
+
+  public async setValueForTextbox(label: string, value: string): Promise<void> {
```
webpro, line 10: lol, reminds me of https://github.com/kadena-community/kadena.js/blob/90a96dd0e9a0ec6db7c5784a261ccc0e99bd91e4/packages/tools/lint-package/.eslintrc.cjs#L9-L10

packages/tools/e2e-tests/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "incremental": true,
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "declaration": false, // fixes pnpm issue: https://github.com/microsoft/TypeScript/issues/42873
+    "declarationMap": false,
+    "jsx": "preserve",
+    "allowJs": true
+  },
+  "include": ["/**/*.ts", "/**/*.tsx"],
```
webpro, line 18: This workspace also has its files in `src` dir? You can omit this `include`? No need to ignore that `playwright.config.ts` and `node_modules`.

packages/tools/e2e-tests/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "lint": "pnpm run /^lint:.*/",
+    "lint:fmt": "prettier . --cache --check",
+    "lint:pkg": "lint-package",
+    "lint:src": "eslint src --ext .js,.ts",
+    "test:e2e": "npx playwright test"
```
webpro, line 15: I think you can omit `npx`, it should be available as a binary in `$PATH`. Package managers add `node_modules/.bin` to `$PATH` automatically.

webpro (approved):
A few minor things. Overall I like how readable the tests are with the test helpers, already setting a great example going forward 👍

packages/tools/e2e-tests/src/page-objects/react-ui/card.component.ts
Ghislain89: There are quite a few functions that simply don't _need_ to return anything. I personally like to make this explicit though.

packages/tools/e2e-tests/tsconfig.json
Ghislain89: I'll change this. I moved all folders in to a src directory fairly late on because I felt it was quite messy otherwise.  The above was required to get it working previously.

packages/tools/e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,36 @@
...
+    channel: 'chromium',
+    trace: `retain-on-failure`,
+  },
+  expect: {
+    timeout: 10 * 1000,
+  },
+  webServer: {
+    command: `pnpm --filter ${process.env.TESTOBJECT} run start`,
+    url: 'http://127.0.0.1:3000',
+    reuseExistingServer: !process.env.CI,
```
alber70g, line 27: What happens when this is an empty string? For clarity could be `process.env.CI === undefined`

packages/tools/e2e-tests/src/helpers/mock.helper.ts
```diff
@@ -0,0 +1,18 @@
+import type { Page } from '@playwright/test';
+
+export default class MockHelper {
```
alber70g, line 3: Can we use named exports, for easy importing and maintainability

packages/tools/e2e-tests/src/page-objects/react-ui/navHeader.component.ts
```diff
@@ -0,0 +1,15 @@
+import type { Locator, Page } from '@playwright/test';
+
+export default class NavHeaderComponent {
```
alber70g, line 3: Can we use named exports, for easy importing and maintainability

packages/tools/e2e-tests/src/page-objects/react-ui/notificationContainer.component.ts
```diff
@@ -0,0 +1,19 @@
+import type { Locator, Page } from '@playwright/test';
+
+export default class NotificationContainerComponent {
```
alber70g, line 3: Can we use named exports, for easy importing and maintainability

packages/tools/e2e-tests/src/page-objects/tools-app/components/aside.component.ts
```diff
@@ -0,0 +1,13 @@
+import type { Page } from '@playwright/test';
+
+export default class AsideComponent {
```
alber70g, line 3: Can we use named exports, for easy importing and maintainability

packages/tools/e2e-tests/src/page-objects/tools-app/components/toolsHeader.component.ts
```diff
@@ -0,0 +1,33 @@
+import CardComponent from '../../react-ui/card.component';
+import NavHeaderComponent from '../../react-ui/navHeader.component';
+
+import type { Locator, Page } from '@playwright/test';
+
+export default class ToolsHeaderComponent extends NavHeaderComponent {
+  public networkCard: CardComponent;
+
```
alber70g, line 8: Can we use named exports, for easy importing and maintainability

(Applies to all the others as well. I stopped adding the comment from here)

packages/tools/e2e-tests/package.json
Ghislain89: Confirmed, removed.

packages/tools/e2e-tests/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./node_modules/@kadena-dev/heft-rig/tsconfig-base.json",
+  "compilerOptions": {
+    "types": ["node"],
+    "rootDir": "./",
```
alber70g, line 5: Why we wouldn't have `src` as `rootDir`?

alber70g (approved):

packages/tools/e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,36 @@
...
+  retries: process.env.CI ? 1 : 0,
+  workers: process.env.CI ? 4 : 1,
+  reporter: [['github'], ['list'], ['html', { open: 'never' }]],
+  use: {
+    headless: !!process.env.CI,
+    baseURL: process.env.PLAYWRIGHT_BASE_URL
+      ? process.env.PLAYWRIGHT_BASE_URL
+      : 'http://127.0.0.1:3000',
+    channel: 'chromium',
+    trace: `retain-on-failure`,
```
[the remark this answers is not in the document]
Ghislain89: fixed.

packages/tools/e2e-tests/playwright.config.ts
Ghislain89: Empty string assume false, but i'll change it either way to make it more clear.

packages/tools/e2e-tests/tsconfig.json
Ghislain89: It wasn't before. Good Catch.

packages/tools/e2e-tests/src/fixtures/mocks/send.ts
Ghislain89: These objects are representations of the response objects made by the blockchain. I'm not entirely sure we can make these types much more specific.

PR #998, opened 2023-10-05
[@kadena/graph] Feat: Update prisma schema to follow naming convention

nil-amrutlal-dept:
Updated prisma schema fields to follow naming convention `camelCase` and changed necessary files accordingly

---
  - https://app.asana.com/0/0/1205582598928047


packages/apps/graph/prisma/schema.prisma
```diff
@@ -12,124 +12,124 @@ datasource db {
...
+  creationTime    DateTime      @db.Timestamptz(6) @map("creationtime")
+  epoch           DateTime      @db.Timestamptz(6)
+  flags           Decimal       @db.Decimal(20, 0)
+  hash            String        @id @db.VarChar
+  height          BigInt
+  minerAccount    String        @db.VarChar @map("miner")
+  nonce           Decimal       @db.Decimal(20, 0)
+  parentBlockHash String        @db.VarChar @map("parent")
+  payload         String        @db.VarChar
+  powhash         String        @db.VarChar
```
MRVDH, line 25: powHash

packages/apps/graph/src/graph/objects/Block.ts
```diff
@@ -10,12 +10,11 @@ export default builder.prismaNode('Block', {
   fields: (t) => ({
     // database fields
     hash: t.exposeID('hash'),
-    chainid: t.expose('chainid', { type: 'BigInt' }),
-    creationtime: t.expose('creationtime', { type: 'DateTime' }),
+    chainid: t.expose('chainId', { type: 'BigInt' }),
+    creationtime: t.expose('creationTime', { type: 'DateTime' }),
     epoch: t.expose('epoch', { type: 'DateTime' }),
     height: t.expose('height', { type: 'BigInt' }),
     powhash: t.exposeString('powhash'),
```
MRVDH, line 17: same here

packages/apps/graph/src/graph/objects/Block.ts
```diff
@@ -10,12 +10,11 @@ export default builder.prismaNode('Block', {
   fields: (t) => ({
     // database fields
     hash: t.exposeID('hash'),
-    chainid: t.expose('chainid', { type: 'BigInt' }),
```
MRVDH, line 13: The GraphQL fields should reflect the changes as well, so this should also be chainId in the JS. (same for all other changed properties as well)

MRVDH (changes requested):

packages/apps/graph/src/graph/objects/Block.ts
nil-amrutlal-dept: I believe that has already been reflected (it gives comp error otherwise). This comment is on a removed piece of code.

MRVDH (approved):

webpro (approved):

PR #1012, opened 2023-10-09
feat(@kadena/react-ui): added textarea and textareafield to the react ui lib

ferreroltd:
This PR is including 2 new components, the textArea and the TextAreaField, which wraps the former with the InputWrapper, similar to the TextField


.changeset/tiny-weeks-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kadena/react-ui': minor
+---
+
+adding new text area component and the field for it
```
eileenmguo, line 5: These end up being the release notes, so they should be written as if other developers are reading it. I would suggest something like this 
```suggestion
Added the TextArea and TextAreaField components
```

packages/libs/react-ui/src/components/TextArea/TextArea.css.ts
```diff
@@ -0,0 +1,107 @@
...
+]);
+
+export const outlinedClass = style([
+  sprinkles({
+    borderRadius: '$sm',
+  }),
+  {
+    border: `1px solid ${vars.colors.$neutral3}`,
+  },
+]);
```
eileenmguo, line 107: I believe this came from the input component, but this should actually be part of a variant since it is a style. Also this would align better with other components. I've added it as a task on the asana board to update in the future https://app.asana.com/0/1205553603119741/1205687963085321/f

packages/libs/react-ui/src/components/TextAreaField/TextAreaField.stories.tsx
```diff
@@ -0,0 +1,68 @@
+import type { ITextAreaFieldProps } from './TextAreaField';
```
eileenmguo, line 1: In stories, I usually import from @component because it acts as an additional check that the export was done correctly. This essentially tests it while importing it like a consuming application

packages/libs/react-ui/src/components/TextAreaField/TextAreaField.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const TextAreaField: FC<ITextAreaFieldProps> = ({
+  disabled = false,
+  textAreaProps,
+  ...rest
+}) => {
+  const { id } = textAreaProps;
+
+  return (
+    <InputWrapper htmlFor={id} disabled={disabled} {...rest}>
+      <Textarea disabled={disabled} {...textAreaProps} />
```
eileenmguo, line 22: You still need to apply the `id` to the text area for the label to be applied to it

eileenmguo (commented):
Just some minor comments. Good job on your first PR!

packages/libs/react-ui/src/components/TextArea/TextArea.css.ts
ferreroltd: this is actually outdated as part of the new story im doing, which will make it always outlined and the status colors will apply to this border as well

packages/libs/react-ui/src/components/InputWrapper/InputWrapper.tsx
```diff
@@ -39,15 +39,17 @@ export const InputWrapper: FC<IInputWrapperProps> = ({
...
+          ? React.Children.map(children, (child) => {
+              if (!React.isValidElement(child)) return null;
+              const props = {
+                ...child.props,
+                leadingTextWidth,
+              };
 
-          return React.cloneElement(child, props);
-        })}
+              return React.cloneElement(child, props);
```
timoheddes, line 50: We're likely switching to using the Context API for passing props into subcomponents. We could refactor this later or, if you like, you can have a look at how I did it here: https://github.com/kadena-community/kadena.js/pull/954

eileenmguo: Here is the ticket for it: https://app.asana.com/0/1205693948735844/1205693948735848

packages/libs/react-ui/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,58 @@
...
+  HTMLTextAreaElement,
+  ITextareaProps
+>(function TextArea({ outlined, disabled = false, fontFamily, ...rest }, ref) {
+  const [value, setValue] = useState<string>('');
+
+  const handleValueChange = ({
+    target,
+  }: React.ChangeEvent<HTMLTextAreaElement>): void => {
+    setValue(target.value);
+  };
```
eileenmguo, line 37: This should be coming from the props, not internal state, otherwise users of the TextArea component will not able to control the value or in this case, get the updated value

packages/libs/react-ui/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export interface ITextareaProps
+  extends Omit<
+      TextareaHTMLAttributes<HTMLTextAreaElement>,
+      'as' | 'disabled' | 'children' | 'className' | 'id'
+    >,
+    Partial<Pick<Sprinkles, 'fontFamily'>> {
+  id: string;
+  value: string;
+  onValueChange: (event: ChangeEvent<HTMLTextAreaElement>) => void;
```
eileenmguo, line 23: I believe onChange will be included in the TextareaHTMLAttributes so you shouldn't need to add it here or on the textarea element, it should be part of `rest`

eileenmguo (approved):

timoheddes:
It's still using cloneElement rather than context but since Eileen approved it already, I guess we'll change that later. Looks good otherwise! I'll also approve

timoheddes (approved):

PR #1013, opened 2023-10-09
[@kadena/graph] Fix minor bugs on simulate

nil-amrutlal-dept:
Encountered bugs:

- [x] Fix the 'Debit Rejected' bug, due to calculation round ups.
- [x] Create the logs folder before trying to add log files to it.
- [x] ~~Hide or ignore the SPV error: An error occurred while calling spv API: Error: SPV target not reachable: target chain not reachable. Chainweb instance is too young~~ -> this one will be fixed in the client itself. Already talked with @javadkh2 about this
- [x] Git ignore the logs folder and its files.
- [x] Sometimes transfer to own accoutn still happen 

---
  - https://app.asana.com/0/0/1205678544389911


packages/apps/graph/src/scripts/devnet/crosschain-transfer.ts
```diff
@@ -93,9 +93,11 @@ export async function crossChainTransfer({
   const status = await listen(submittedTx);
   inspect('Transfer Result')(status);
   if (status.result.status === 'failure') {
-    throw new Error('DEBIT REJECTED');
+    throw new Error('Transfer failed');
   }
 
+  const originalConsoleLog = console.log;
+  console.log = () => {};
```
MRVDH, line 100: Is this so the spv log doesn't show up anymore? I'd rather have it show up than overwriting system functions

MRVDH (approved):

packages/apps/graph/src/scripts/devnet/crosschain-transfer.ts
nil-amrutlal-dept: This was a mistake, shouldn't be there. I'll remove it. The spv log will be replaced in the client itself and we will not overwrite system functions

packages/apps/graph/src/scripts/devnet/file.ts
```diff
@@ -21,7 +29,9 @@ export function createFile(filename: string): string {
     requestKey: '',
     type: undefined,
   };
-  const filepath = path.join(__dirname, 'logs/', filename);
+  const directory = path.join(__dirname, `${devnetConfig.LOG_FOLDERNAME}/`);
```
webpro, line 32: Minor: just `path.join(__dirname, devnetConfig.LOG_FOLDERNAME)` should work

webpro (approved):

PR #1014, opened 2023-10-09
feat(ui): implement non-outlined and inline style notification component

realdreamer:
- Introduced the variant as `standard` and `outlined`. But I'm open for discussion here whether we wanna go with outline as a boolean prop. 
- Added inline boolean prop to make the inline notification component for medium and larger screens. 

If we're agreeing to the implementation then I'll update all app/libs which are already consuming notification component

~TODO:~
- [x] Update apps/libs already consume notification component.

### DIFF
--- a/packages/libs/react-ui/src/components/Notification/Notification.css.ts
+++ b/packages/libs/react-ui/src/components/Notification/Notification.css.ts
@@ -1,4 +1,4 @@
-import { breakpoints, sprinkles } from '@theme/sprinkles.css';
+import { sprinkles } from '@theme/sprinkles.css';
 import type { ColorType } from '@theme/vars.css';
 import { vars } from '@theme/vars.css';
 import { style, styleVariants } from '@vanilla-extract/css';
@@ -52,16 +52,13 @@ export const inlineVariants = styleVariants({
   true: [
     sprinkles({
       display: 'flex',
-      flexDirection: 'column',
-    }),
-    {
-      '@media': {
-        [`screen and ${breakpoints.md}`]: {
-          alignItems: 'flex-start',
-          flexDirection: 'row',
-        },
+      alignItems: {
+        md: 'flex-start',
       },
-    },
+      flexDirection: {
+        md: 'row',
+      },
+    }),
   ],
   false: [sprinkles({ display: 'flex', flexDirection: 'column' })],
 });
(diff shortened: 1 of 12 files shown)


packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -44,6 +43,29 @@ export const expandVariants = styleVariants({
...
+    sprinkles({
+      display: 'flex',
+      flexDirection: 'column',
+    }),
+    {
+      '@media': {
+        [`screen and ${breakpoints.md}`]: {
+          alignItems: 'flex-start',
+          flexDirection: 'row',
+        },
```
eileenmguo, line 62: I believe you can put this in sprinkles

eileenmguo (commented):
I think using the variant makes sense for the actual style. Do you think expanded and inline could be a variant as well since it seems like they can't both be applied?

packages/libs/react-ui/src/components/Notification/Notification.css.ts
eileenmguo: you can use this syntax ```
```suggestion
sprinkles({
alignItems: {
md: "flex-start",
}), 
flexDirection: {
md: "row"
}
})
```

packages/libs/react-ui/src/components/Notification/Notification.stories.tsx
```diff
@@ -21,6 +21,12 @@ const meta: Meta<
     },
   },
   argTypes: {
+    variant: {
+      options: ['standard', 'outlined'],
```
timoheddes, line 25: Can be `options: Object.keys(displayVariants) as (keyof typeof displayVariants)[],`

realdreamer:
> Do you think expanded and inline could be a variant as well since it seems like they can't both be applied?

It can applied. expanded makes it full width. the `inline` makes the content inside of the `Notification` block to be next to each other. 

image

packages/libs/react-ui/src/components/Notification/Notification.css.ts
realdreamer: Thank you! I'll try that syntax :)

timoheddes (approved):

eileenmguo (approved):

PR #1016, opened 2023-10-09
[DOCS]fix: render hydration issues for sidemenu

sstraatemans:
there was a warning when rendering a page.

---
  - https://app.asana.com/0/0/1205677110602458


packages/apps/docs/src/components/Layout/components/Template/Template.tsx
```diff
@@ -26,7 +26,7 @@ export const Template: FC<IProps> = ({
   hideSideMenu = false,
 }) => {
   const { isMenuOpen, closeMenu } = useMenu();
-  const isMediumDevice = useMedia(breakpoints.md);
+  const isMediumDevice = useMedia(breakpoints.md, true);
```
sstraatemans, line 29: for SSR UseMedia needs a default value

packages/apps/docs/src/hooks/useWindowScroll.ts
```diff
@@ -11,7 +11,7 @@ export function useWindowScroll(): [
     y: 0,
   });
 
-  React.useLayoutEffect((): (() => void) => {
```
sstraatemans, line 14: useLayoutEffect can have sideeffects with SSR
only use for components that are only rendered on clientside

realdreamer (approved):
Good catch. Thank you for taking care of it.

PR #1018, opened 2023-10-10
Fix up some config etc. (attempt 2)

webpro:
This PR improves/fixes some configuration and dependencies in our monorepo.

_(Duplicating #1008, as CI runs kept getting cancelled (devops cancel culture?). Eventually traced it down to a combo of adding the root workspace (`- '.'`) to `pnpm-workspace.yaml` and `"extends": ["//"]` to the root `turbo.json` (which turbo documents to do). Oh well, let's move on.)_

Here's the list of commits:

* Fix pretter/prettier-plugin-packagejson location (77d18c790)
* Sync versions across workspaces (dcf49b625)
* Fix turbo config to include root workspace (4e16c1432)
* Temp fix for knip + next-translate-plugin issue (64eb4dc59)
* Upgrade some deps (e717af64f)
* Update changelog (e33b17408)
* Trial & error: upgrade @microsoft/api-extractor (80907c029)
* Trial & error: abandon root workspace idea in pnpm-workspace.yaml (301ce968d)
* Add linter + fix issues in root workspace (7eede98a1)


.changeset/twenty-bottles-enjoy.md
```diff
@@ -0,0 +1,27 @@
+---
```
Takadenoshi, line 1: do we need 2 changesets?

packages/apps/tools/package.json
```diff
@@ -76,16 +76,15 @@
     "@vanilla-extract/jest-transform": "1.1.1",
     "@vanilla-extract/next-plugin": "2.3.1",
     "@walletconnect/types": "~2.8.1",
+    "@webpro/next-translate-plugin": "^2.6.3",
```
Takadenoshi, line 79: ?

webpro: - Ugly: https://github.com/kadena-community/kadena.js/pull/1018/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L14
- Fix: https://github.com/aralroca/next-translate-plugin/pull/55

tl/dr; that plugin doesn't support custom base path cq monorepos so I fixed it. We'll go back to the original once it's sorted downstream.

packages/apps/docs/package.json
```diff
@@ -57,7 +57,7 @@
     "mdast-util-to-markdown": "~1.5.0",
     "mdast-util-to-string": "~3.2.0",
     "mobx": "~6.9.0",
-    "next": "13.5.3",
+    "next": "13.4.5",
```
realdreamer, line 60: what is the reason for downgrading next version here? 

cc @sstraatemans

packages/apps/docs/package.json
```diff
@@ -110,8 +110,7 @@
     "micromark-extension-frontmatter": "~1.1.0",
     "mini-css-extract-plugin": "2.7.6",
     "openapi-types": "~12.1.3",
-    "prettier": "~3.0.0",
-    "prettier-plugin-packagejson": "^2.4.5",
+    "prettier": "~3.0.3",
```
realdreamer, line 113: if we're allowing prettier at root level dependency, then why are we keeping prettier again here?

EnoF (approved):

packages/apps/docs/package.json
webpro: We have a policy in the repo to make sure the version of a package is the same across the monorepo. First I tried to upgrade the others, but got some error, so then downgraded this one.

packages/apps/docs/package.json
webpro: To keep them as much self-contained/isolated as possible.

packages/apps/docs/package.json
sstraatemans: hm the new version was helping us with some performance.
not really happy with this one

barthuijgen: I think it makes sense for packages like `next` to have an exception, this is something close to the app itself, you don't want to force the whole repo to upgrade it at once every time we upgrade in the future. Already proven by the fact that you had troubles doing it now.

`next` isn't some bundled dependency that is going to give issues in builds or increase bundle sizes.

webpro: Next.js has a dependency on React. Component libs too. So to prevent any issues here (we've had them before) we should stay in sync, _especially_ libs like Next.js + React.

alber70g (approved):

PR #1019, opened 2023-10-10
fix(docs): update cookie consent ui changes

realdreamer:


packages/apps/docs/src/components/CookieConsent/CookieConsent.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleReject = useCallback(() => {
+    setCookieConsent(false);
+  }, []);
+
+  if (cookieConsent !== null) return null;
+
+  return (
+    <div className={notificationClass}>
+      <Notification.Root
```
sstraatemans, line 35: can u add the wrapper (i think we use in Header)
the consent content is way to wide atm.
so the notification background is good. but the content needs to be as wide as the header content

packages/apps/docs/src/components/Layout/components/Menu/menu.css.ts
```diff
@@ -73,13 +73,7 @@ export const menuInLayoutVariants = styleVariants({
 });
 
 export const menuLayoutVariants = styleVariants({
-  landing: {
```
sstraatemans, line 76: the menu now moves below the consent notification. is this what we want?

packages/apps/docs/src/components/Layout/components/TitleHeader/style.css.ts
```diff
@@ -73,7 +73,7 @@ export const headerClass = style([
   {
     gridArea: 'main',
     zIndex: 3,
-    backgroundColor: $$backgroundOverlayColor,
+    backgroundColor: 'rgba(250,250,250, .8)',
```
sstraatemans, line 76: the variable is already this color

sstraatemans: this will not work in darkmode

realdreamer: ah okay! But actually we can't use this same global variable import from this file because the context has changed now after moving TileHeader out of the Main content area. But I'll have a look and define it in variable.

sstraatemans (approved):

PR #1039, opened 2023-10-13
[@tools-website] Disable button when 'mainnet' is selected in faucet

KristinaSpasevska:


sanderlooijenga (approved):

packages/apps/tools/src/pages/faucet/existing/index.tsx
```diff
@@ -122,7 +122,9 @@ const ExistingAccountFaucetPage: FC = () => {
     [chainID, t],
   );
 
-  const showNotification = selectedNetwork !== 'testnet04';
+  const testnetNotSelected: boolean = selectedNetwork !== 'testnet04';
```
Randynamic, line 125: I would rather don't have a negative selector. I would go for `mainnetSelected` as that is also stated in the name of the PR. This will also open up the possibility to later on connect to a different non-mainnet network.

sanderlooijenga (approved):

PR #1052, opened 2023-10-16
Feat/textarea and field borders

ferreroltd:
Title: Enhance UI Consistency: Centralize Styling Logic

Description:
I want to propose a significant improvement to our "fields" and "inputs". Currently, our select, input, and textArea components have different styles based on our statuses. To keep a more unified and developer-friendly experience, I suggest centralizing the styling logic within the inputWrapper component.

Why?
From my point of view, to have a status we usually stay in the context of a form where things need to "make sense" maybe with a (BE) contract or similar set of rules, we can always use this inputs separatedly but I dont see that we would evaluate the content as standalone inputs.
By consolidating the styling logic in inputWrapper, we not only achieve consistent visuals but also reinforce the idea that these components are primarily meant to be used within the structured environment of a form. This approach maintains clarity, "enforcing" this behaviour from our side


eileenmguo:
Should we close this PR?

PR #1071, opened 2023-10-17
fix(@kadena/react-ui): Fix NavHeader glow, refactored to useContext, and added accessibility improvements

eileenmguo:
- Updated the NavHeader to useContext instead of cloneElement
- Fixed the glow positioning when using the asChild prop
- Removed the glow when no path is selected
- Made the nav links active when the activeHref had the link Href as a subPath (we can make this configurable in the future if needed, but I updated this since the docs website works this way, and the tools site never showed active nav links)
- Added Accessibility improvements NOTE: the Select is not accessible, but will be checked/updated when the other select components are reviewed


packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
```diff
@@ -61,13 +61,18 @@ const Header: FC<IHeaderProps> = () => {
     setSelectedNetwork((e.target as HTMLSelectElement).value as Network);
   };
 
+  console.log({ pathname });
```
timoheddes, line 64: Cleanup?

packages/libs/react-ui/src/components/BrandLogo/index.tsx
```diff
@@ -1,12 +1,13 @@
...
+import { KadenaLogo } from './svgs/Kadena';
+import { KadenaDevToolsLogo } from './svgs/KadenaDevTools';
+import { KadenaDocsLogo } from './svgs/KadenaDocs';
 
 import type { FC, SVGProps } from 'react';
 import React from 'react';
 
-export type LogoVariant = 'Kadena' | 'DevTools' | 'Docs';
-export const logoVariants: LogoVariant[] = ['Kadena', 'DevTools', 'Docs'];
+// eslint-disable-next-line @kadena-dev/typedef-var
```
timoheddes, line 8: Do we need this? Works fine without it for me.

eileenmguo: This is just a minor change that allows us to add the logos to the constant and generates the types based on that array instead of having to edit both the type and array. It's not likely that these would get out of sync since they are right next to each other, but I just have a preference to make updates in as few places as possible to avoid accidental mistakes

timoheddes: Yeah I know what it does, I meant the disable. My IDE seems fine with it, without the eslint-disable?

eileenmguo: I get a warning when running the lint script

timoheddes (approved):

PR #1077, opened 2023-10-17
streamline usage of turbo and node versions across workflows

Ghislain89:
With this PR, the vast majority of workflows have been consolidated to either run as part of the CI workflow or as a seperate workflow if strictly required. 

This has a few benefits:
* easier maintenance of workflows, reducing duplication
* Only run specific jobs (e.g. integration tests) if build & unit tests succeed
* Fully benefit from Turbo's Remote Caching resulting in faster builds 
* Visualisation of the relationship between various jobs.


.github/workflows/visual-regression-tests.yaml
```diff
@@ -0,0 +1,65 @@
...
+        include:
+          - package: packages/libs/react-ui
+            projectToken: 'REACT_UI_CHROMATIC_TOKEN'
+          - package: packages/apps/docs
+            projectToken: 'DOCS_CHROMATIC_TOKEN'
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history, needed to determine diffs.
```
alber70g, line 27: Should we have emoji's here? Doesn't really add value imo

alber70g (approved):

PR #1084, opened 2023-10-18
chore: update build scripts with turbo build

realdreamer:
The pipe started to fails after removing build script from root


webpro (approved):
🚀

PR #1092, opened 2023-10-19
[@kadena/graph-client] Prefill module field and add labels

MRVDH:
- https://app.asana.com/0/1205553603119741/1205755088427583/f
- https://app.asana.com/0/1205553603119741/1205739032590541/f

### DIFF
--- a/packages/apps/graph-client/src/pages/index.tsx
+++ b/packages/apps/graph-client/src/pages/index.tsx
@@ -28,7 +28,7 @@ const Home: React.FC = () => {
 
   const [searchType, setSearchType] = useState<string>('request-key');
   const [searchField, setSearchField] = useState<string>('');
-  const [moduleField, setModuleField] = useState<string>('');
+  const [moduleField, setModuleField] = useState<string>('coin');
 
   const search = (): void => {
     switch (searchType) {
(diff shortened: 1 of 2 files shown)


nil-amrutlal-dept:
In the cases of Request Key and Block is it better to leave an empty placeholder or to still have something like `Enter block hash`?
I think for the block especially it would be beneficial to have a placeholder since it's not self-explanatory like Request Key.

packages/apps/graph-client/src/pages/index.tsx
```diff
@@ -51,13 +51,20 @@ const Home: React.FC = () => {
...
     block: 'Block Hash',
   };
 
+  const searchTypePlaceholders: Record<string, string> = {
+    'request-key': '',
+    account: 'k:1234...',
+    event: 'coin.TRANSFER',
+    block: '',
+  };
+
```
nil-amrutlal-dept, line 67: In the cases of Request Key and Block is it better to leave an empty placeholder or to still have something like Enter block hash?
I think for the block especially it would be beneficial to have a placeholder since it's not self-explanatory like Request Key.

```suggestion
  const searchTypePlaceholders: Record<string, string> = {
    'request-key': 'Request Key',
    account: 'k:1234...',
    event: 'coin.TRANSFER',
    block: 'Block Hash',
  };

```

nil-amrutlal-dept (approved):

packages/apps/graph-client/src/pages/index.tsx
MRVDH: Fixed!

PR #1093, opened 2023-10-19
[@kadena/graph-client][@kadena/graph] Show recent and all transactions (Feature)

nil-amrutlal-dept:
- Implemented recent transactions in the block overview page
- Implemented the `/transactions` page which shows the most recent transactions
- Adapted `graph` transactions query to make all parameters not required and dynamically add them if present
- Created `extended-transactions-table` and replaced multiple pages code with reusable component
- Fixed broken links in the transaction page (it was using block ID instead of block hash)

### DIFF
--- a/packages/apps/graph-client/src/components/extended-transactions-table/extended-transactions-table.tsx
+++ b/packages/apps/graph-client/src/components/extended-transactions-table/extended-transactions-table.tsx
@@ -1,9 +1,9 @@
-import { Box, Button, Grid, Link, Select, Table } from '@kadena/react-ui';
+import { Box, Button, Link, Table } from '@kadena/react-ui';
 
 import type { GetTransactionsQuery } from '@/__generated__/sdk';
 import routes from '@/constants/routes';
-import type { FetchMoreOptions,FetchMoreQueryOptions  } from '@apollo/client';
-import React, { useState } from 'react';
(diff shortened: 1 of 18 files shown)


packages/apps/graph-client/src/components/extended-transactions-table/extended-transactions-table.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+                if (fetchMoreResult.transactions.edges.length < 10) {
+                  return {
+                    ...prev,
+                    transactions: {
+                      ...fetchMoreResult.transactions,
+                      edges: [
+                        ...fetchMoreResult.transactions.edges,
+                        ...prev.transactions.edges,
+                      ].slice(0, 10),
```
MRVDH, line 77: Shouldn't this '10' and the 10 above be 20 instead, since you get the first and last 20 int he fetch more?

MRVDH (approved):

packages/apps/graph-client/src/components/extended-transactions-table/extended-transactions-table.tsx
nil-amrutlal-dept: Yep, it should. I was already proceeding with the changes for selectability on table size and pagination but thought I'd leave that for another PR. I'll leave it 10 for this one

PR #1095, opened 2023-10-19
Animate Accordion component content height on toggle

timoheddes:
---
  - https://app.asana.com/0/0/1205693670996249


Randynamic:
@eileenmguo are we still going to do something with this PR

PR #1105, opened 2023-10-23
Migrate react-ui & tools from jest to vitest

webpro:
This PR migrates react-ui and the tools app to Vitest (the docs app is cleaned up, does not have such tests).

This was an interesting migration, since the combo of Vitest + happy-dom results in actually parsing the stylesheets, ~which led me to find bugs (regarding the dark theme class) that also have been fixed in this PR.~

The ESLint plugins we can reconsider when migrations are finished. For now it's removed as it wasn't doing anything.

Some mocks have been removed, an a test is un-skipped, resulting in increased coverage.


packages/apps/docs/src/components/PageHelpful/styles.css.ts
```diff
@@ -23,7 +23,7 @@ export const textAreaClass = style([
     '::placeholder': {
       color: vars.colors.$gray40,
     },
-    [`${darkThemeClass} &::placeholder`]: {
+    [`.${darkThemeClass} &::placeholder`]: {
```
sstraatemans, line 26: this is not needed.

sstraatemans (changes requested):
the darkthemeclass does not need the '.'

packages/apps/docs/src/components/PageHelpful/styles.css.ts
webpro: The class name variable does not and should not contain a dot (a selector might contain one) and the stylesheet parser in our tests complains about it. Maybe there's something that fixes it in the chain eventually, but when parsing the stylesheet happy-dom rightfully complained about it.

webpro: Well, this is embarrassing. I've been analyzing stylesheets spitted out by the stylesheet parser from happy-dom, and found those dots missing. Now I can no longer reproduce this, so the dots are back. Apologies!

packages/libs/react-ui/src/styles/sprinkles.css.ts
```diff
@@ -33,7 +33,7 @@ const systemProperties = defineProperties({
 const colorProperties = defineProperties({
   conditions: {
     lightMode: {},
-    darkMode: { selector: `.${darkThemeClass} &` },
+    darkMode: { selector: `${darkThemeClass} &` },
```
eileenmguo, line 36: We should leave the `.` in front of the classname. The variable only includes the class name

sstraatemans (approved):

packages/apps/docs/src/components/PageHelpful/styles.css.ts
eileenmguo: Yeah, the variable shouldn't include the dot, so adding the dot is necessary

packages/libs/react-ui/src/styles/sprinkles.css.ts
webpro: OK restored this too. It's the same variable, no? 🤷  Oh, well.

eileenmguo: Yes, it is the same variable haha

timoheddes (approved):

alber70g (approved):

packages/libs/react-ui/src/styles/sprinkles.css.ts
webpro: So this PR no longer has changes related to this, and the migrated tests are passing, so let's leave it at this for now.

packages/apps/docs/src/components/PageHelpful/styles.css.ts
webpro: So this PR no longer has changes related to this, and the migrated tests are passing, so let's leave it at this for now.

eileenmguo (approved):

PR #1109, opened 2023-10-23
fix(inputs): unify outline prop in stories and added it to select

ferreroltd:


eileenmguo:
I don't believe this is the correct style

image
image

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -0,0 +1,22 @@
+import { sprinkles } from '@theme/sprinkles.css';
+import { darkThemeClass, vars } from '@theme/vars.css';
+import { style } from '@vanilla-extract/css';
+// use this in all our inputs
```
timoheddes, line 4: This comment is supposed to be and stay here or was it just for you while working on it?

packages/libs/react-ui/src/components/Form/TextField/TextField.stories.tsx
```diff
@@ -1,7 +1,8 @@
+import type { ITextFieldProps } from '@components/Form';
+import { TextField } from '@components/Form';
+// what do we want this to look like
```
timoheddes, line 3: Should we answer and decide before this PR is merged? 😄

packages/libs/react-ui/src/components/Form/Form.css.ts
ferreroltd: this one ill remove it, forgot to cleanup

packages/libs/react-ui/src/components/Form/TextField/TextField.stories.tsx
ferreroltd: this is an open question, i saw the import bellow that looks odd and wanted to know how do we want to import it like so we make consistent imports, so what do u think

.changeset/many-feet-taste.md
eileenmguo, line 1: very unfortunate file name 😅

packages/libs/react-ui/src/components/Form/TextField/TextField.stories.tsx
eileenmguo: It's not the prettiest import, but I think it's fine. If you want, you could move styles that are shared to the Form.css file

packages/libs/react-ui/src/components/index.ts
```diff
@@ -74,11 +77,18 @@ export { Button } from './Button';
 export { Card } from './Card';
 export { ContentHeader } from './ContentHeader';
 export { Divider } from './Divider/Divider';
+export {
+  Input,
+  InputWrapper,
```
eileenmguo, line 82: Would it make sense to update this component to be named FormFieldWrapper? Right now it feels very tied to the Input. What do you think?

eileenmguo (commented):
Visually it looks a little bit off. It looks like the border padding (in this case outline-offset) is not the same on every side. I would suggest updating the offset to 2px and trying to create the border bottom effect with a box shadow instead

packages/libs/react-ui/src/components/index.ts
ferreroltd: yeah you are right! ill do it

eileenmguo: I think it's better to do that as part of a separate PR to keep things separated. I'll add a ticket to the board and assign you to it!

eileenmguo: https://app.asana.com/0/1205693948735844/1205783514269055

We can work on this in the next sprint

eileenmguo (approved):
Approved under the condition that the styles will be refined in the PR with focus states

PR #1113, opened 2023-10-24
Feat/tools home page

sanderlooijenga:
---
  - https://app.asana.com/0/0/1205770207666323


Randynamic (approved):

packages/apps/tools/src/pages/home/index.tsx
```diff
@@ -6,11 +6,62 @@ import {
...
 import useTranslation from 'next-translate/useTranslation';
+import Head from 'next/head';
+import Link from 'next/link';
+
 import type { FC } from 'react';
 import React, { useRef } from 'react';
 
+const faqs: Array<{ title: string; body: React.ReactNode }> = [
+  {
+    title: 'What can I do with the Faucet?',
```
KristinaSpasevska, line 27: Will we add translation option for this?

packages/apps/tools/src/pages/home/index.tsx
```diff
@@ -72,15 +126,105 @@ const Home: FC = () => {
...
+        <Heading bold={false} variant="h3">
+          Kadena Developer Tools
+        </Heading>
+        <Heading bold={false} as="h2" variant="h5" color="default">
+          We&apos;re constantly adding new Developer Tools to make it easier for
+          our builders to utilize all Kadena has to offer.
+        </Heading>
+        <Card fullWidth>
+          <Grid.Root columns={2}>
+            <Grid.Item>
```
KristinaSpasevska, line 139: Would it be nicer to crate a const that will keep objects with the values of each GridItem and have a render function to generate these in order to avoid repetition in the file and keep things tidier? Let me know what do you think about it

KristinaSpasevska (approved):

packages/apps/tools/src/pages/home/index.tsx
sanderlooijenga: I prefer to keep it this way, it's now simple to add items and doesn't have too many. Otherwise, we would need to add logic (e.g. dealing with internal/external links) which isn't worth it at the moment, in my opinion.

packages/apps/tools/src/pages/home/index.tsx
sanderlooijenga: Added translations for these (by using [Trans](https://github.com/aralroca/next-translate#trans-component) component)

PR #1116, opened 2023-10-24
Add width and height props to layout components

timoheddes:
- Added height, minHeight, maxHeigh, width, minWidth, and maxWidth to the Box, Stack, and Grid components
- Added a className prop to the Box, Stack, and Grid components to allow for additional styling since those components do not have any visual styling that we wouldn't want the user to override. This makes those components configurable by sprinkles props in many instances, while still allowing users further customize the layout components for other circumstances when they need more control
- Reduced the number of values in the width/height sprinkles since they create a lot of css classes. I don't think we use these values enough to warrant creating utility classes for them. I wanted to also remove the sizes vars from the width and height properties, but decided to leave that to when we have our new token system


packages/libs/react-ui/src/styles/sprinkles.css.ts
```diff
@@ -103,6 +103,10 @@ const responsiveProperties = defineProperties({
       'min-content': 'min-content',
       'max-content': 'max-content',
     },
+    maxHeight: {
+      '100%': '100%',
+      maxContent: 'auto',
```
timoheddes, line 108: What do we want this to be?

eileenmguo: I think this would be a good question to ask in the channel. We likely don't need to give all the "size" variables here

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -51,7 +51,7 @@ export const cardColorVariants = styleVariants(colorVariants, (color) => {
 
 export const expandVariants = styleVariants({
   true: [sprinkles({ width: '100%', maxWidth: '100%' })],
-  false: [sprinkles({ width: 'max-content', maxWidth: 'maxContent' })],
+  false: [sprinkles({ width: 'max-content', maxWidth: '$maxContent' })],
```
sstraatemans, line 54: why is width not `$maxContent`?

eileenmguo: This PR doesn't change any existing styling. The notification's component was implemented to be the width of it's max-content, however I believe this will change in the future when the notification's component is refactored with the new UX designs in figma

packages/libs/react-ui/src/components/Box/Box.tsx
```diff
@@ -9,66 +10,84 @@ export interface IBoxProps
     Pick<
       Sprinkles,
       | 'display'
+      | 'height'
```
sanderlooijenga, line 13: Not actually used in props below?

packages/libs/react-ui/src/components/Box/Box.tsx
```diff
@@ -9,66 +10,84 @@ export interface IBoxProps
...
-      | 'paddingX'
-      | 'paddingY'
-      | 'paddingTop'
       | 'paddingBottom'
       | 'paddingLeft'
       | 'paddingRight'
+      | 'paddingTop'
+      | 'paddingX'
+      | 'paddingY'
+      | 'width'
```
sanderlooijenga, line 32: Not actually used in props below?

.changeset/strange-bees-talk.md
```diff
@@ -0,0 +1,6 @@
+---
+'@kadena/react-ui': minor
+---
+
+Add new width and height properties to sprinkles and update Box, Stack and Grid
```
sanderlooijenga, line 5: Also `className` prop? 😁

packages/libs/react-ui/src/components/Notification/Notification.css.ts
eileenmguo: Also `$maxContent` is a custom token that is 42.5rem / 680px while `max-content` is the native css property value. Maybe I can update the token to be `$maxContentWidth`?

packages/libs/react-ui/src/components/Box/Box.tsx
eileenmguo: ahh thank you for looking closely at these, I forgot to 😅

packages/libs/react-ui/src/components/Grid/GridItem.tsx
```diff
@@ -23,17 +24,20 @@ const assembleColumnSpanVariants = (
...
   columnSpan,
   rowSpan = 1,
 }) => {
-  const className = classNames(
-    gridItemClass,
-    rowSpanVariants[rowSpan],
-    columnSpan && assembleColumnSpanVariants(columnSpan),
-  );
   return (
-    <div className={className} data-testid="kda-grid-item">
```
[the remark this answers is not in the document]
eileenmguo: The reason is because we don't want to clutter the html with unnecessary attributes and this encourages us to test based on aria roles which aligns more with actual user interaction and helps us put more focus on a11y
There are some of the threads where this is discussed on slack:
https://kadena-io.slack.com/archives/C05DZR7GPFH/p1691567264887629
https://kadena-io.slack.com/archives/C05DZR7GPFH/p1696588154162339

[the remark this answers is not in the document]
sanderlooijenga: Thanks for the clarification 🙌

sanderlooijenga:
Should the newly added props also be available to configure in Storybook?

realdreamer (approved):

packages/libs/react-ui/src/components/Notification/Notification.css.ts
realdreamer: yes, `$maxContentWidth` avoid confusion. else it resonates same as `max-content`

ferreroltd (approved):

sstraatemans (approved):

PR #1117, opened 2023-10-24
[@kadena/graph] Endpoint to execute arbitrary PACT queries (Feature)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1205794055767759


packages/apps/graph/generated-schema.graphql
```diff
@@ -149,6 +149,7 @@ type Query {
   blocksFromHeight(chainIds: [Int!], startHeight: Int!): [Block!]!
   chainAccount(accountName: String!, chainId: String!, moduleName: String!): ChainModuleAccount!
   completedBlockHeights(chainIds: [String!], completedHeights: Boolean, heightCount: Int): [Block!]!
+  executePact(chainId: String!, pactQuery: [String!]!): [String!]!
```
alber70g, line 152: Can we change this to each query having it's `code` and `chainId`?

More like
```gql
executePact(queries: [PactQuery]): [String]

type PactQuery {
  code: String!
  chainId: String!
}
```

alber70g (changes requested):

packages/apps/graph/generated-schema.graphql
nil-amrutlal-dept: Fixed! The only thing I would want to know is if the singular and plural query namings are correct. At the moment I did it like this:
```graphql
executePact(query: PactQuery): String
executePacts(queries: [PactQuery]): [String]
```

alber70g: Good one!
I'm thinking that maybe `executePact` could be ambiguous and lead to confusion as `defpact` is a multi-step transaction, and we only execute queries in Pact this way, hence I choose `pactQuery`. What do you think?

alber70g (approved):

PR #1120, opened 2023-10-25
Feat/form focus styles

ferreroltd:
This pr adds focus styles for all input form and fixes some style issues


packages/libs/react-ui/src/styles/vars.css.ts
```diff
@@ -53,7 +53,8 @@ export const vars = createGlobalTheme(':root', {
   },
   shadows: {
     // TODO: Update to match design system
-    $1: `0px 1px 2px 0 $colors$gray40`,
+    $1: `0px 1px 0 0 ${colorPalette.$gray30}`,
+    $2: `0px 1px 0 0 ${colorPalette.$gray60}`,
```
eileenmguo, line 57: What is the reason for adding these to the vars.css file?

ferreroltd: the reason is that we probably sould predefine our shadows here and reuse them, so i added one for each theme, i can also not do it and specify it manually in the Form css

eileenmguo: If it's used somewhere else as well, I think it make sense to add it here, but then it would probably have a more descriptive token name. In this case I think it is specific to the input components so i would define it there

eileenmguo (commented):
bottom border looks too bright in dark mode 
image

packages/libs/react-ui/src/styles/vars.css.ts
ferreroltd: yeah i was thinking on giving it shadow input related name

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -1,20 +1,73 @@
...
+
+    outlineOffset: '2px',
     selectors: {
       [`${darkThemeClass} &`]: {
-        outline: `2px solid ${vars.colors.$gray60}`,
+        outline: `2px solid ${fallbackVar(
+          statusOutlineColor,
+          vars.colors.$blue60,
+        )}`,
       },
```
eileenmguo, line 42: Why do you need this dark theme version?

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -1,20 +1,73 @@
...
+      [`${darkThemeClass} &`]: {
+        outline: `2px solid ${fallbackVar(statusColor, vars.colors.$gray60)}`,
+      },
+    },
+  },
+]);
+
+export const statusClass = style([
+  sprinkles({
+    borderRadius: '$sm',
```
eileenmguo, line 27: I don't think you need this style

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -1,20 +1,73 @@
...
+  }),
+  {
+    outline: `2px solid ${fallbackVar(
+      statusOutlineColor,
+      vars.colors.$blue60,
+    )}`,
+
+    outlineOffset: '2px',
```
eileenmguo, line 35: I think you can move this outlineOffset style to the baseContainerClass and then it should be applied regardless of the state

packages/libs/react-ui/src/components/Form/Input/Input.stories.tsx
```diff
@@ -87,17 +86,6 @@ const meta: Meta<IInputProps> = {
...
-      description:
-        'Width of the leading text. Defaults to the size of the text itself.',
-      control: {
-        type: 'select',
-      },
-      options: [
-        '- Omit this property to auto-size the leading text',
-        ...Object.keys(vars.sizes).map((key) => key as keyof typeof vars.sizes),
-      ],
-    },
```
eileenmguo, line 100: Why did you remove this? This should still be possible to apply via the Input

packages/libs/react-ui/src/components/Form/Input/Input.tsx
```diff
@@ -32,24 +30,17 @@ export interface IInputProps
 
 export const Input: FC<IInputProps> = forwardRef<HTMLInputElement, IInputProps>(
   function Input(
-    {
-      outlined,
-      leadingText,
-      leadingTextWidth,
```
eileenmguo, line 38: LeadingTextWidth should still be able to be applied via the InputProps as well. Additionally, it should override the leadingTextWidth provided by the context if both are configured

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -1,20 +1,73 @@
...
+    selectors: {
+      [`${darkThemeClass} &`]: {
+        outline: `2px solid ${fallbackVar(statusColor, vars.colors.$gray60)}`,
+      },
+    },
+  },
+]);
+
+export const statusClass = style([
```
eileenmguo, line 25: I don't think you're using this class at all. Can it be removed?

packages/libs/react-ui/src/components/Form/InputWrapper/InputWrapper.context.ts
```diff
@@ -0,0 +1,16 @@
+import type { vars } from '@theme/vars.css';
+
+import { createContext } from 'react';
+import type { Status } from '../Form.css';
+
+export type OpenSections = string[];
+
+interface IInputWrapperContext {
+  status?: Status;
```
eileenmguo, line 9: I saw earlier there was a `FormFieldStatus` type. Should that be used here as well or are these different?

packages/libs/react-ui/src/components/Form/Input/Input.stories.tsx
ferreroltd: I thought this always came throught the wrapper, not as a prop as well, i can bring it back

packages/libs/react-ui/src/components/Form/Select/Select.tsx
```diff
@@ -34,13 +35,16 @@ export const Select: FC<ISelectProps> = forwardRef<
...
+
   const Icon = icon && SystemIcon[icon];
   const ChevronDown = SystemIcon.ChevronDown;
 
   return (
     <div
       className={classNames(containerClass, {
         [containerClassDisabled]: disabled,
+        [statusClass]: status,
         [baseOutlinedClass]: outlined,
```
eileenmguo, line 48: Previously in the Input component I saw you were able to apply the status colors using the baseOutlinedClass. Can you do that here too?

ferreroltd: yes

ferreroltd: yes

packages/libs/react-ui/src/components/Form/InputWrapper/InputWrapper.context.ts
ferreroltd: we do export it as FormFieldStatus, inside the context of form is called status, i can rename for consistency

packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -11,12 +11,14 @@ export interface ITextFieldProps
 export const TextField: FC<ITextFieldProps> = ({
   disabled = false,
   inputProps,
+  status,
   ...rest
 }) => {
   const { id } = inputProps;
+  console.log(status);
```
eileenmguo, line 18: cleanup

packages/libs/react-ui/src/components/Form/InputWrapper/InputWrapper.context.ts
ferreroltd: yes will do, i think it makes more sense

packages/libs/react-ui/src/styles/vars.css.ts
eileenmguo: Shall remove these variables completely since they aren't in the sprinkles file anymore?

eileenmguo:
When there is a status selected, it doesn't seem like there is a difference between the focused state and non focused state

packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -1,20 +1,48 @@
...
-import { style } from '@vanilla-extract/css';
+import { createVar, fallbackVar, style } from '@vanilla-extract/css';
+
+export type FormFieldStatus = 'disabled' | 'positive' | 'warning' | 'negative';
+export const statusColor = createVar();
+export const statusOutlineColor = createVar();
 
 export const baseOutlinedClass = style([
+  {
+    outline: `2px solid ${fallbackVar(statusColor, vars.colors.$gray30)}`,
```
eileenmguo, line 12: I think you mean to use the statusOutlineColor

packages/libs/react-ui/src/components/Form/Input/Input.tsx
```diff
@@ -35,21 +35,26 @@ export const Input: FC<IInputProps> = forwardRef<HTMLInputElement, IInputProps>(
...
       icon,
       rightIcon,
+      leadingTextWidth: propLeadingTextWidth,
       disabled = false,
       ...rest
     },
     ref,
   ) {
+    const { status, leadingTextWidth: contextLeadingTextWidth } =
```
eileenmguo, line 46: nit: I would call this wrapperLeadingTextWidth

eileenmguo (commented):
Looking good so far! I like the reduction of code 😍

eileenmguo (approved):
Only one comment left to remove the shadows from sprinkles

salamaashoush (approved):
there are some stuff that could be better but I guess it is more of a general issue in the ui library and should be fixed with tokens,  applying dark theme styles per component

KristinaSpasevska (approved):

sanderlooijenga (approved):

PR #1126, opened 2023-10-26
chore(@kadena/react-ui): Refactor modal component to use react-aria

eileenmguo:
Used react-aria hooks to create Modal and Dialog components and checked that the appropriate accessibility attributes are used. All current implementations of the Modal will need to be replaced with the Dialog component.

The Dialog component is composed of Dialog, DialogHeading, DialogFooter, and DialogContent components. The Content component scrolls when it is larger than the max height of the Dialog.

The Dialog component uses the Modal component to create a styled modal that can be used to display information or prompt the user for input. It accepts a function that provides state to pass to it's children or a ReactNode as children.

The Modal component includes the underlay and portal, but it will not be exposed unless we deem it necessary in the future to compose other overlays. This component accepts a state object that is provided by `react-stately`. If we decide to expose the Modal in the future, we will likely provide a hook that wraps `react-stately` to provide props for the modal.

The following functionality has changed in the Dialog component (previously Modal): 
- The modal closes when hitting the escape key
- The modal content scrolls
- I removed the text from the close button as it seemed unnecessary 
- The modal dimension have been adjusted to max width and height on mobile

The structure has a standalone Modal component that you can pass content to through children and can be controlled by `isOpen` and `onOpenChange` props.

NOTE: Consuming apps have not been updated yet


packages/libs/react-ui/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,44 +1,63 @@
...
-  args: {
-    title: 'Default Title',
-  },
-  render: ({ title }) => {
+  render: () => {
+    const state = useOverlayTriggerState({});
+    const { triggerProps, overlayProps } = useOverlayTrigger(
+      { type: 'dialog' },
+      state,
+    );
```
eileenmguo, line 33: For now, I think it's best to not expose the Modal component. Right now it requires some knowledge of react-aria so I think it would be best to expose a hook to provide wrap react-aria, but seems like over engineering for now

packages/libs/react-ui/src/components/Modal/Modal.tsx
```diff
@@ -1,68 +1,63 @@
...
-  const { clearModal } = useModal();
+export const Modal: FC<IModalProps> = ({
+  className,
+  children,
+  state,
+  isDismissable = true,
+  isKeyboardDismissDisabled,
+}) => {
+  const nodeRef = useRef<HTMLDivElement | null>(null);
+  const underlayRef = useRef<HTMLDivElement | null>(null);
```
eileenmguo, line 31: Is this underlayRef being used for anything? I only see it being applied to the underlay div

packages/libs/react-ui/src/components/Modal/Modal.stories.tsx
salamaashoush: yeah, this is a storybook feel free to remove it and also remove the export from the index file

packages/libs/react-ui/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+  className?: string;
+  title?: ReactNode;
+  children?: ((state: OverlayTriggerState) => ReactNode) | ReactNode;
+}
+
+const BaseDialog = React.forwardRef<HTMLDivElement, IBaseDialogProps>(
+  (props, ref) => {
+    const {
+      title,
+      className,
```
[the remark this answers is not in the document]
salamaashoush: for me  exposing className is a good idea/habit and I would say we will need to have it for all components to allow extensibility but also feel free to remove it if you see it is not necessary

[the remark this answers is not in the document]
eileenmguo: Discussed privately: We will remove them for now to stay consistent with our previous decision to not expose classNames as a prop, but discuss the addition of classNames in the next retro as a team

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -51,7 +51,7 @@ export const cardColorVariants = styleVariants(colorVariants, (color) => {
 
 export const expandVariants = styleVariants({
   true: [sprinkles({ width: '100%', maxWidth: '100%' })],
-  false: [sprinkles({ width: 'max-content', maxWidth: 'maxContent' })],
+  false: [sprinkles({ width: 'max-content', maxWidth: '$maxContent' })],
```
sstraatemans, line 54: is the width not `$maxContent`? like maxWidth?

eileenmguo: No, because at least this version of the notifications component isn't supposed to be always be a fixed size. I believe this will likely change in the future though

packages/libs/react-ui/src/components/Dialog/Dialog.css.ts
```diff
@@ -0,0 +1,53 @@
+import { sprinkles } from '@theme/sprinkles.css';
```
sstraatemans, line 1: is it correct that we can have 2 modals popping up on the same screen?
image

sstraatemans: i dont think this should be possible imho

packages/libs/react-ui/src/components/Modal/Modal.tsx
```diff
@@ -1,68 +1,55 @@
...
-            onClick={handleCloseModal}
-          />
-          <div className={wrapper} data-cy="modal" data-testid="kda-modal">
-            <section className={modal}>
-              <Card fullWidth>
-                <div className={titleWrapper}>
-                  <Heading as="h3">{title}</Heading>
-                </div>
-
-                <button
```
sstraatemans, line 51: the dialog now scrolls the complete content.
I think this is fine, but in designs discussed with isa (for the docs search), the scrolling could be controlled by the content it self.
as in example below the header + the search and tabs always stay visible.
Regardless I think the close button should always be visible and not scroll.
Can we discuss with Isa?

image

packages/libs/react-ui/src/components/Dialog/Dialog.css.ts
eileenmguo: Hmm So I assume you were only able to achieve this by putting a Dialog inside of another Dialog since focus is shifted to the dialog when one is open. Is that correct?

packages/libs/react-ui/src/components/Modal/Modal.tsx
eileenmguo: hmm yeah, I think it's good for us to have a chat about this

eileenmguo: I've updated it to have Header, Footer, and Content cubcomponents. The Content section will scroll if there is overflow

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -30,7 +30,7 @@ export interface IButtonProps
   onClick?:
     | React.MouseEventHandler<HTMLButtonElement>
     | React.FormEventHandler<HTMLButtonElement>;
-  target?: '_blank' | '_self';
+  target?: string;
```
sanderlooijenga, line 33: Why not reuse the type from `HTMLAttributeAnchorTarget`?

packages/libs/react-ui/src/index.ts
```diff
@@ -83,7 +88,6 @@ export {
   Label,
   Link,
   MaskedValue,
-  ModalProvider,
```
sanderlooijenga, line 86: Should `Modal` also be exported here?

packages/libs/react-ui/src/components/Form/index.ts
```diff
@@ -1,4 +1,4 @@
-export type { FormFieldStatus } from './Form.css';
```
sanderlooijenga, line 1: Is it no longer a type?

sanderlooijenga (approved):

packages/libs/react-ui/src/components/Dialog/Dialog.css.ts
eileenmguo: We will assume users can ensure this doesn't happen since this can only be achieved either by the same trigger opening to modals or nested modals.

salamaashoush (approved):

sstraatemans (approved):

sstraatemans (approved):

PR #1135, opened 2023-10-26
Upate Integration Test Workflow for Kadena Client and Client Utils

Ghislain89:
This PR updates the Integration test workflow to a matrix build as well as enable turbo caching. This has the following benefits:

- Integration tests for kadena/client as well as kadena/client-utils can run in parallel (as seperate builds) instead of sequentually
- Both builds will use an isolated instance of devnet, instead of the same instance.
- improved readability for Github actions by providing both runs with a clear name:
image

- If the required artifact (e.g. for kadena/client) is already available in the cache use it, otherwise rebuild using turbo.
- Enable Caching for the integration test task. e.g. if there have been no changes to the src dir for any package, no need to run the tests.
- Add typescript dependency to fix the client-utils build.


webpro (approved):
Excellent stuff!

alber70g (approved):

PR #1142, opened 2023-10-30
HoverTag component

sanderlooijenga:
### HoverTag Component
Adds a new component which works like a "_HoverCard_" component in other libraries (see references below). For now, it's added to the Tools app, in the future it might be moved to the `@kadena/react-ui` library.

[Figma;](https://www.figma.com/file/jZh2B45PIsZJGss6OLnk63/%3Ckda-tag-%2F%3E?type=design&node-id=103-2&mode=design&t=SxUHxahvOmNqSMFV-0)
Screenshot 2023-10-27 at 09 18 04

References;
- https://www.radix-ui.com/primitives/docs/components/hover-card
- https://en.wiktionary.org/wiki/hovercard

---
  - https://app.asana.com/0/0/1205820167883084


packages.json
```diff
@@ -7,73 +7,73 @@
   },
   {
     "name": "@kadena/chainweb-node-client",
-    "version": "0.4.4",
+    "version": "0.5.0",
```
sanderlooijenga, line 10: No idea why this changes 🤷  these changes happened after running `pnpm format`..

KristinaSpasevska (approved):

packages/apps/tools/src/pages/_app.tsx
```diff
@@ -10,6 +11,8 @@ import type { AppProps } from 'next/app';
 import type { FC } from 'react';
 import React from 'react';
 
+KodeMono();
```
eileenmguo, line 14: I think you can remove this now

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
```diff
@@ -0,0 +1,49 @@
...
+        data-testid="kda-hover-tag"
+        className={tagClass}
+        onMouseEnter={(e: React.MouseEvent<HTMLElement>) =>
+          Tooltip.handler(e, ref)
+        }
+        onMouseLeave={(e: React.MouseEvent<HTMLButtonElement>) =>
+          Tooltip.handler(e, ref)
+        }
+      >
+        <span className={tagStyles}>{maskValue(value)}</span>
```
eileenmguo, line 30: Do you think it might be better to have a `MaskedText` component rather than utility function since we don't expose any other utility functions? 

I'm thinking a `MaskedText` component that accepts a string with an optional disabled prop

eileenmguo (commented):
Great tests 😍

I have a suggestion on the structure of the HoverTag, but it will require more changes to the existing tag. Right now the styles are getting duplicated which means it's easier for them to get out of sync. What do you think of the following:
- Updating the `Tag` component to accept ReactNode as children rather than just a string
- Use the `Tag` component in `HoverTag` to wrap the MaskedText and button
- Wrap the tag in a span with the hover properties 

`
<span {...hoverStuff}>
  <Tag>
    <MaskedText />
    <Button />
  </Tag>
</span>
<Tooltip />
`

packages/apps/tools/src/pages/_app.tsx
sanderlooijenga: This in the Tools app, not react-ui, so we still need this 😁

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
sanderlooijenga: I can do that, but I do think I would need to use this utility function in other places. And also; there wouldn't be any harm in exposing this utils function, right?

Could we keep it as is for now and if I do encounter many use cases where I am just using this `maskValue` function to show as a text I will then refactor this and make a `MaskedText` component. Agreed?

sanderlooijenga:
> I have a suggestion on the structure of the HoverTag, but it will require more changes to the existing tag. Right now the styles are getting duplicated which means it's easier for them to get out of sync. What do you think of the following:
> 
> * Updating the `Tag` component to accept ReactNode as children rather than just a string
> * Use the `Tag` component in `HoverTag` to wrap the MaskedText and button
> * Wrap the tag in a span with the hover properties
> 
> ```
> <span {...hoverStuff}>
>   <Tag>
>     <MaskedText />
>     <Button />
>   </Tag>
> </span>
> <Tooltip />
> ```

@eileenmguo I updated the code to your suggestion, please check 😄

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
eileenmguo: yeah, that works 👍

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
```diff
@@ -0,0 +1,45 @@
...
+        data-testid="kda-hover-tag"
+        onMouseEnter={(e: React.MouseEvent<HTMLElement>) =>
+          Tooltip.handler(e, ref)
+        }
+        onMouseLeave={(e: React.MouseEvent<HTMLButtonElement>) =>
+          Tooltip.handler(e, ref)
+        }
+      >
+        <Tag>
+          <span className={monoFontTextStyles}>{maskValue(value)}</span>
```
eileenmguo, line 25: you can use the `Text` component with font set to "mono" for this as well

eileenmguo (approved):

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
sanderlooijenga: Ah super, did that, thank you!

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
sanderlooijenga: Will do, thanks Eileen 👌

packages.json
alber70g: Yes they should be changed when [ci] Release PR is merged. @Ghislain89 is looking into it

PR #1157, opened 2023-11-01
[test] run simulation for a set period, to set up test data

Ghislain89:


packages/apps/graph/src/scripts/devnet/helper.ts
```diff
@@ -31,7 +31,7 @@ export const logger = createLogger('info');
 const getClient = (): IClient =>
   createClient(
     ({ chainId, networkId }) =>
-      `http://localhost:${devnetConfig.PORT}/chainweb/0.0/${networkId}/chain/${chainId}/pact`,
+      `http://127.0.0.1:${devnetConfig.PORT}/chainweb/0.0/${networkId}/chain/${chainId}/pact`,
```
Ghislain89, line 34: I seem to be having issues using localhost sometimes, i will likely revert this once I debug it further.

packages/apps/graph/src/scripts/devnet/simulate.ts
```diff
@@ -161,6 +165,6 @@ export async function simulate({
     }
     counter++;
     // Timeout
-    await new Promise((resolve) => setTimeout(resolve, transferInterval));
+    await new Promise((resolve) => setTimeout(resolve, duration));
```
nil-amrutlal-dept, line 168: This will ensure that when it loops through the length of accounts, it will timeout for the given total duration of the simulation. I am not sure this is the intended behaviour 😅 
```suggestion
    await new Promise((resolve) => setTimeout(resolve, transferInterval));
```

nil-amrutlal-dept (changes requested):

packages/apps/graph/src/scripts/devnet/simulate.ts
```diff
@@ -172,6 +181,21 @@ export async function simulate({
...
+      if (totalTransfers >= maxTransfers) {
+        logger.info('Maximum number of transfers reached');
+        break;
+      }
+
+      // Check if duration has been reached
+      if (Date.now() - startTime >= duration) {
+        logger.info('Duration reached');
+        break;
+      }
```
nil-amrutlal-dept, line 197: The reason behind separating it into 2 `if` statements it just so we can have conclusive logging

PR #1160, opened 2023-11-01
[@kadena/graph] Normalized error handling

MRVDH:
This PR contains several things:

- We have a function that checks what type of error is being thrown, and formats them to a normalised error message, since we have errors coming from multiple systems, not just one.
- Since chainweb-node and kadena/client don't throw clear messages, we try to infer the type and the meaning of the error based on the error object we get.
- We will keep adding different types of errors and their infered meanings the more we find.


nil-amrutlal-dept:
Looks quite good to me 🙌🏽🙌🏽 
Taking into account the difficulties in type safety and the different types of responses we can have, I don't see a better way of uniformizing error handling

packages/apps/graph/src/utils/errors.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Checks if the error is a row not found error. An example case is when we loop over all chainIds to get the account details for a module account. If the module account does not exist on a chain, we get a row not found error. We want to ignore this error and continue with the next chainId.
+ */
+export function isRowNotFoundError(error: any): boolean {
+  return (
+    error instanceof PactCommandError &&
+    error.pactError?.message?.includes('with-read: row not found')
+  );
+}
```
nil-amrutlal-dept, line 73: The only downside is that we'd have to create 'identifying' functions for different types of errors where we'd want to give a more specific description. But in any case, I do not see a better way of doing this

nil-amrutlal-dept:
There's some conflicts to be resolved, but apart from that everything looks good 🚀 🙌🏽

nil-amrutlal-dept (approved):

nil-amrutlal-dept:
General remark: Should we maybe add a timeout to the subscriptions?
In the `graph-client` it just stays in loading state if a certain non-existing request-key is the input. Maybe a timeout, with an error message and a 'Try again' button would be nice? 
Also some validation on the input of the request key would be nice. At the moment I can search for `123` as request key (when we know it has a defined length). 

I think all these are out of scope but just things I noticed while testing these changes.

packages/apps/graph/src/services/node-service.ts
```diff
@@ -52,3 +69,21 @@ export async function getAccountDetails(
...
+      Pact.builder
+        .execution(code)
+        .setMeta({
+          chainId: chainId as ChainId,
+        })
+        .setNetworkId(dotenv.NETWORK_ID)
+        .createTransaction(),
+    );
+
+    return JSON.stringify(commandResult.result);
```
nil-amrutlal-dept, line 85: Before, for the `pactQuery` we were returning the `commandResult.result.data` stringified and not just the result.
```suggestion
    if (commandResult.result.status !== 'success') {
      throw new PactCommandError(
        'Pact Command failed with error',
        commandResult,
      );
    }
    return JSON.stringify(commandResult.result.data);
```

MRVDH: Correct. Though I can imagine that if you want to run raw queries, you would also want to have the raw output (including the info on how the request was built). @alber70g What do you think

nil-amrutlal-dept: That's a good point actually

nil-amrutlal-dept: Nonetheless, I think we should still throw an error if the status is not success in order to give a more descriptive error message

MRVDH: Yep, applied 👍

packages/apps/graph/src/utils/errors.ts
MRVDH: Yep indeed. I think we have to look at what kadena client evolves into with their error handling and base our changes on that in the future

packages/apps/graph/src/index.ts
```diff
@@ -1,23 +1,17 @@
+require('module-alias/register');
```
alber70g, line 1: Can this be done with `import`?
```suggestion
import('module-alias/register');
```

packages/apps/graph/tsconfig.json
```diff
@@ -2,6 +2,9 @@
   "extends": "./node_modules/@kadena-dev/shared-config/tsconfig-base.json",
   "compilerOptions": {
     "types": [".kadena/pactjs-generated", "node"],
-    "target": "ES2020"
+    "target": "ES2020",
+    "paths": {
+      "@src/*": ["./src/*"]
+    }
```
alber70g, line 8: Would it be nicer to have a path for each directory in `./src/`?

alber70g (approved):

packages/apps/graph/src/index.ts
MRVDH: It can, but the auto sort for imports will move it down. For now it's okay, but it will break if an import is sorted above it that uses module aliases.

packages/apps/graph/tsconfig.json
MRVDH: Yep, done!

packages/apps/graph/src/index.ts
alber70g: Ah good one. Yes we can ignore the auto-fix I think?
This is one of the reasons we used to have a custom order. Sometimes you need a order when importing like that.

What happens when you use `import 'module-alias/register'`?

PR #1162, opened 2023-11-01
[@kadena/graph] , [@kadena/graph-client] Implement endpoint for graph and chainweb-data (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1205842872749423


packages/apps/graph/src/graph/objects/GraphAndChainwebData.ts
```diff
@@ -0,0 +1,12 @@
+import { builder } from '../builder';
+
+export default builder.objectType('GraphAndChainwebData', {
```
MRVDH, line 3: I'm not too sure about this name. In theory, all of the queries return "graph and chainweb data". We should call this GraphConfiguration maybe? Since that is what we are getting, right? Configuration data?

MRVDH (changes requested):

packages/apps/graph/src/graph/objects/GraphAndChainwebData.ts
nil-amrutlal-dept: Yep indeed. I changed the name a couple of times, didnt have a name for it. GraphConfiguration sounds good, but we return data such as `minimumBlockHeight` which is not really a configuration. But in any case, it seems like a better name than the current one

MRVDH (approved):

alber70g (approved):

PR #1169, opened 2023-11-03
Use shared vitest config and enable thresholdAutoUpdate for some packages.

Ghislain89:
Currently vitest is configured to gather code coverage in CI as well as not run in watch mode. After various tests both of these do not seem to work correctly. This PR aims to fix this behaviour.

With this PR, all relevant packages now correctly depend on the shared vitest configuration and only override settings that are package specific. Code coverage is now collected again.

There are some packages that do not meet the globally defined criteria defined in the @kadena-dev/shared-config package. The current coverage levels have been set as baseline for these packages and [ThresholdAutoUpdated](https://vitest.dev/config/#coverage-thresholdautoupdate) has been enabled. As developers work to improve coverage, the baseline will be increased as well.

The following packages are impacted by this:

- chainweb-node-client
- tools dApp
- chainweb-stream-client
- client-utils
- kadena/client
- react-ui <-- Ran in to issues for this one. Disabled Threshold updater for this package

The result wil be that, gradually code coverage will be increased for packages that are touched often until they meet the globally defined defaults.

---
  - https://app.asana.com/0/0/1205881579952684


packages/apps/tools/vitest.config.ts
```diff
@@ -12,10 +12,10 @@ export default mergeConfig(
       threads: false,
       coverage: {
         provider: 'v8',
-        lines: 70,
-        functions: 40,
+        lines: 73.76,
```
sanderlooijenga, line 15: What are these values?

sanderlooijenga (approved):

javadkh2 (approved):
Great!

packages/apps/tools/vitest.config.ts
Ghislain89: These are the required coverage thresholds, meaning if your unit tests have less then the required thresholds: You won't be able to merge. Thresholds have been defined globally at 90% for lines, functions branches and statements. 


The tools dApp currently does not meet these, which is why we're overriding these values for this specific package with lower values. The parameter `thresholdAutoUpdate` will automatically increase these values as you write more unit tests. Once the tools dApp meets the required thresholds of 90%, these overrides can be removed.

ash-vd (approved):

packages/apps/tools/vitest.config.ts
sanderlooijenga: Ah super, thanks for the explanation 👌

alber70g (approved):

sstraatemans (approved):

Takadenoshi (approved):

eileenmguo (approved):

KristinaSpasevska (approved):

PR #1170, opened 2023-11-03
chore(@kadena/react-ui): Exported the Textarea and TextareaField components

eileenmguo:
Previously the Textarea and TextAreaField components had inconsistent naming conventions so I've updated them to Textarea and TextareaField since the underlying element is `textarea`. I've also exported both components from the root barrel file


packages/libs/react-ui/src/components/Form/Textarea/index.ts
```diff
@@ -0,0 +1 @@
+export { Textarea, type ITextareaProps } from './Textarea';
```
sstraatemans, line 1: these should be capital A as well?

sstraatemans (approved):

PR #1179, opened 2023-11-06
[@kadena/graph] , [@kadena/graph-client] Estimate transaction gas limit (Feat) 

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1205842872749418


packages/apps/graph/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "paths": {
       "@db/*": ["./src/db/*"],
       "@services/*": ["./src/services/*"],
-      "@utils/*": ["./src/utils/*"]
+      "@utils/*": ["./src/utils/*"],
+      "@devnet/*": ["./src/devnet/*"]
```
nil-amrutlal-dept, line 10: Added devnet as alias since i changed folders

alber70g (approved):
We might want to use `@kadena/client-utils` for the gas estimation and get-balance. We can implement that somewhere in the future though.

PR #1189, opened 2023-11-07
[@tools-website] Home and faucet page UI fixes

KristinaSpasevska:


packages/apps/tools/src/components/Common/Layout/partials/Sidebar/MenuLinkButton.tsx
```diff
@@ -1,41 +1,33 @@
...
   href?: string;
   active?: boolean;
+  target?: string;
 }
 
 export const MenuLinkButton: FC<IMenuLinkButtonProps> = ({
   active,
   title,
   href,
+  target = '_self',
```
sanderlooijenga, line 19: `target` isn't used anywhere?

packages/apps/tools/src/utils/persist.ts
```diff
@@ -9,39 +9,10 @@ export const encode = (value: unknown) =>
...
 export const getItem = (key: string) => {
-  try {
-    const cookie = Cookies.get(getName(key));
-    return cookie !== undefined ? parse(cookie) : undefined;
-  } catch (e) {
-    return undefined;
-  }
-};
-
-export const deleteItem = (key: string) => {
```
sanderlooijenga, line 25: Out of curiosity; Why are these deleted?

packages/apps/tools/src/utils/persist.ts
```diff
@@ -9,39 +9,10 @@ export const encode = (value: unknown) =>
...
-    {} as { [key: string]: string },
-  );
-};
-
-export const purge = () => {
-  const cookies = getItems();
-  Object.keys(cookies).forEach((key) => {
-    deleteItem(key);
-  });
+  const cookie = Cookies.get(getName(key));
```
sanderlooijenga, line 16: Why no longer is `try … catch`?

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
```diff
@@ -43,44 +31,13 @@ export const AccountNameField: FC<IAccountNameFieldProps> = ({
...
-  } = useWalletConnectClient();
-
-  useDidUpdateEffect(() => {
-    if (session && accounts?.length) setMode('select');
-    else setMode('input');
-  }, [session, accounts]);
+  const { selectedAccount, setSelectedAccount } = useWalletConnectClient();
 
   const lookup = {
-    select: (
```
sanderlooijenga, line 64: Should we really delete this? 🤔

packages/apps/tools/src/components/Common/Layout/partials/Sidebar/styles.css.ts
```diff
@@ -68,12 +51,16 @@ export const gridMiniMenuLinkButtonStyle = style([
     padding: '$5',
     textAlign: 'left',
     fontWeight: '$bold',
+    fontSize: '$sm',
   }),
   {
-    transition: 'all 0.1s ease',
     width: '100%',
-    borderBottom: `1px solid ${vars.colors.$borderDefault}`,
+    borderBottom: `1px solid #E3E1E5`,
```
sanderlooijenga, line 58: I guess these fixed colour values are missing from the design tokens?

packages/apps/tools/src/utils/persist.ts
KristinaSpasevska: They were not used anymore

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
KristinaSpasevska: Didn't want to leave it commented out for later, but if you think that's better I can bring this back (for now we don't have the accounts from the wallet)

packages/apps/tools/src/utils/persist.ts
KristinaSpasevska: trying out something, but will revert this

packages/apps/tools/src/components/Common/Layout/partials/Sidebar/styles.css.ts
KristinaSpasevska: yes, I got them from Isa for the time being until they add all the new tokens in the ui library

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
sanderlooijenga: I mean I guess we can always get them back by referencing this PR, so it's fine like this 👍

packages/apps/tools/src/utils/persist.ts
sanderlooijenga: Were they ever used? I think in this case these could stay in, as they could potentially be used later. What do you think?

KristinaSpasevska: Not a problem to retrieve them

sanderlooijenga: If you can run the formatter then that would be great, I'm approving in the meantime 😄

sanderlooijenga (approved):

PR #1195, opened 2023-11-08
Refactor/rename input wrapper

ferreroltd:
This PR is renaming the InputWrapper to FieldWrapper ans applying some spacing in the case that it contains multiple elements. This spacing is also needed due to the styles for focus and states that the inputs may have.

OLD:
image

NEW:
Screenshot 2023-11-08 at 13 28 12
  With focus style:
Screenshot 2023-11-08 at 13 28 06


packages/libs/react-ui/src/components/Form/FormFieldWrapper/FormFieldWrapper.css.ts
```diff
@@ -1,12 +1,12 @@
 import { sprinkles } from '@theme/sprinkles.css';
 import { vars } from '@theme/vars.css';
-import { styleVariants } from '@vanilla-extract/css';
+import { style, styleVariants } from '@vanilla-extract/css';
```
ferreroltd, line 3: removing this unused

packages/libs/react-ui/src/components/Form/FormFieldWrapper/FormFieldWrapper.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type { IFormFieldHeaderProps } from './FormFieldHeader/FormFieldHeader';
+import { FormFieldHeader } from './FormFieldHeader/FormFieldHeader';
+import { FormFieldHelper } from './FormFieldHelper/FormFieldHelper';
+import { FormFieldWrapperContext } from './FormFieldWrapper.context';
+import { statusVariant } from './FormFieldWrapper.css';
+
+export interface IFormFieldWrapperProps
+  extends Omit<IFormFieldHeaderProps, 'label'> {
+  children:
+    | FunctionComponentElement<IInputProps>
```
ferreroltd, line 17: @eileenmguo here we use the IInputProps but it should be | ITextareaProps | ... right?

eileenmguo: This is ensuring it accepts a lone input or input array, but yes, we should add the additional form elements here

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
```diff
@@ -29,11 +29,7 @@ export const containerClassDisabled = style([
     },
   }),
   {
-    opacity: 0.4,
     selectors: {
-      '.inputGroup &': {
-        opacity: 1,
-      },
```
eileenmguo, line 36: I see this change is specifically for the disabled state. Did you test this? I think this is because the .inputGroup also applies the opacity change so this is making sure that lower opacity is not stacked

Took a look and I think it looks fine 👍

ferreroltd: image
yeah i did check and it was fine

sanderlooijenga (approved):

eileenmguo (approved):
Looks good!

MRVDH (approved):

PR #1205, opened 2023-11-09
Remark typescript

realdreamer:
- Created a new docs-tools TS project
- Migrated existing remark plugins to TS support
- Removed mjs remark plugins from docs and using new docs-tools plugins

---
  - https://app.asana.com/0/0/1205845307247932
  - https://app.asana.com/0/0/1205845307247930
  - https://app.asana.com/0/0/1205956519668642


packages/tools/docs-tools/src/data/authors.json
```diff
@@ -0,0 +1,176 @@
+[
```
sstraatemans, line 1: hm I get why this file is in here.
but if we do that we want the assets in the tools as well.

But more optimal this needs to be in the docs repo.
This is data.
no idea how yet :)

sstraatemans: or we should copy it from docs to docs-tools on build time.
maybe that is best

realdreamer: then I think it's better to do dynamic import instead of copying it.

sstraatemans (approved):

PR #1214, opened 2023-11-13
chore(@kadena/react-ui): Update Divider styles to use sprinkles

eileenmguo:
Updated the Divider styles to use sprinkles


ferreroltd (approved):

packages/libs/react-ui/src/components/Divider/Divider.css.ts
```diff
@@ -1,9 +1,15 @@
...
-  borderWidth: '1px',
-  margin: `${vars.sizes.$10} 0`,
-  borderBottomWidth: '0',
-});
+export const dividerClass = style([
+  sprinkles({
+    backgroundColor: '$borderDefault',
+    width: '100%',
+    marginY: '$10',
+    border: 'none',
```
sstraatemans, line 10: now this border is not visible anymore

eileenmguo: it is using backgroundColor and height to show the divider now which was an arbitrary change while moving styles to sprinkles

sstraatemans (approved):

PR #1215, opened 2023-11-13
[@kadena/graph] , [@kadena/graph-client] Non-nullable edges & transfers and pagination related UI adjustments 

nil-amrutlal-dept:
- Pagination Component was implemented 
- Implemented items per page in Transfers 
- Removed unused logger on `graph`. Might be nice to implement a uniform solution within the `graph` project 
- The transfers on each chain show the counterpart in the correct order. That means transfers going to chain 0 -> chain 1, show as chain 0 -> chain 1 even if we are looking at chain 1 transfers (see image below). For these changes, some additional functions were developed and the overall logic was refined

image

---
  - https://app.asana.com/0/0/1205916745201980


packages/apps/graph-client/src/components/compact-transfers-table/compact-transfers-table.tsx
```diff
@@ -56,72 +81,76 @@ export const CompactTransfersTable = (
...
-                    href={`${routes.TRANSACTIONS}/${edge?.node.requestKey}`}
-                  >
-                    <span title={edge?.node.requestKey}>
-                      {truncate(edge?.node.requestKey)}
+                  <Link href={`${routes.TRANSACTIONS}/${transfer.requestKey}`}>
+                    <span title={transfer.requestKey}>
+                      {truncate(transfer.requestKey)}
                     </span>
                   </Link>
                   /
```
MRVDH, line 147: only add the / if there is a crossChainCounterPart

packages/apps/graph-client/src/components/extended-transfers-table/extended-transfers-table.tsx
```diff
@@ -0,0 +1,250 @@
...
+                    </Link>
+                  ) : (
+                    <span style={{ color: 'lightgray' }}>N/A</span>
+                  )}
+                </Table.Td>
+                <Table.Td>
+                  <Link href={`${routes.TRANSACTIONS}/${edge.node.requestKey}`}>
+                    {edge.node.requestKey}
+                  </Link>
+                  /
```
MRVDH, line 234: only add the / if there is a edge.node.crossChainTransfer

MRVDH (approved):

packages/apps/graph-client/src/components/compact-transfers-table/compact-transfers-table.tsx
nil-amrutlal-dept: Fixed

packages/apps/graph-client/src/components/extended-transfers-table/extended-transfers-table.tsx
nil-amrutlal-dept: Fixed

PR #1221, opened 2023-11-14
chore: install jsdom which fixes vitest skips tests

javadkh2:


alber70g (approved):

javadkh2:
its added via another PR

PR #1222, opened 2023-11-14
Feat/action buttons

ferreroltd:
This PR adds the first action button and exposes the 2 new composed components that are 
TextareaCopy:
image
image

and InputCopy

image
image


packages/libs/react-ui/src/components/Form/Input/Input.tsx
```diff
@@ -78,6 +79,7 @@ export const Input: FC<IInputProps> = forwardRef<HTMLInputElement, IInputProps>(
             disabled={disabled}
             {...rest}
           />
+          <div>{children}</div>
```
ferreroltd, line 82: the default position looked okay

packages/libs/react-ui/src/components/Form/Input/InputCopy/InputCopy.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import type { IInputProps } from '@components/Form';
+import { InputCopy } from '@components/Form';
+
+import type { SystemIcon } from '@components/Icon';
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+
+const meta: Meta<IInputProps> = {
+  title: 'Form/InputCopy',
```
eileenmguo, line 9: Could we add another folder to group all input stories together? I think you can just update this to 'Form/Input/InputCopy'

eileenmguo (commented):
Some feedback discussed during our call 
image

Also the button looks different from the design. I would recommend creating a custom button

packages/libs/react-ui/src/components/Form/Textarea/Textarea.tsx
```diff
@@ -46,6 +47,7 @@ export const Textarea: FC<ITextareaProps> = forwardRef<
           disabled={disabled}
           {...rest}
         />
+        <div className={buttonContainerClass}>{children}</div>
       </div>
```
eileenmguo, line 51: Can we combine the two divs?

realdreamer:
image
image

The icons seems bit off with the size. It's smaller than the current implementation

sstraatemans (approved):

salamaashoush (approved):

eileenmguo (approved):

PR #1223, opened 2023-11-15
[@tools-website] Remove wallet connect from header

KristinaSpasevska:


packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
```diff
@@ -107,7 +106,7 @@ const Header: FC<IHeaderProps> = () => {
           ))}
         </NavHeader.Select>
         <div className={walletConnectWrapperStyle}>
-          <WalletConnectButton />
+          {/*<WalletConnectButton />*/}
```
sanderlooijenga, line 109: Should we also disable/comment out the `Provider`? 🤔

KristinaSpasevska: We still use the networks handling from there, so we need to keep it

sanderlooijenga: Hmm.. you're right 🤔 . Does sound like that Provider does too many things. We might want to refactor that in the (near) future. Okay to keep it as is for now though 👍

sanderlooijenga (approved):

PR #1224, opened 2023-11-15
[chore] add documentation on how changeset works in the monorepo and …

Ghislain89:
This PR documents how changeset works in the mono repo as well as document known caveats.

### DIFF
--- a/docs/releasing-with-changesets.md
+++ b/docs/releasing-with-changesets.md
@@ -1,0 +1,62 @@
+# Releasing wth Changesets
+
+## Process overview
+
+### Adding changesets
+
+Changesets makes releasing our products easier, although it might not always be
+clear how the process works. This document aims to explain the process and
+document known caveats.
+
+When creating a PR, a workflow `ci-changelog`[1] will run. This will determine
+whether a changeset is present in the `.changeset`[2] folder. If it finds none,
+this workflow will fail requesting you to add a changeset. From this point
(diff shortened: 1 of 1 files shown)


docs/releasing-with-changesets.md
```diff
@@ -0,0 +1,62 @@
...
+PR for `@kadena/client` and adds a valid changeset and merges their PR to main.
+This leads to the `[ci] Release` pull request in which the
+
+- `.changeset`[2] folder is emptied
+- Changelogs are updated for the relevant packages
+- Versions are updated in the package(s).json
+
+Before _Person A_ has been able to merge the `[ci] Release` pull request to
+trigger a publish, _Person B_ has merged a new feature for `@Kadena/tools` to
+main which includes a changeset. The `ci-changelog`[1] will run again on main
```
eileenmguo, line 49: Just to clarify, does this mean that every time a PR is merged, the `[ci] Release` PR is replaced with a new version of the PR with the updated set of changesets, meaning that all approvals are lost?

Ghislain89: if a `[CI] Release` branch has just been merged (so no longer exists, it will be recreated. if one already exists (e.g. when two people just merge to to main), then it's updated.

In the case of recreation: yes, all approvals are lost. In the case of an update, I don't believe they're lost.

Randynamic (approved):

PR #1226, opened 2023-11-15
feat(pactjs-generator): accept literal as valid input + use null as default value for proof in builder api

javadkh2:
* accept literal as valid input  that let users pass expressions like pact function calls to the Pact.module["module-name"]["function-name"]
```TS
composePactCommand(
   Pact.module.coin.transfer(
       literal(`(a-pact-function-that-return-account "input")`), 
       ()=>"(second-function)", 
       10
   )
)
```

* use null as default value of proof in continuation when using builder api
* refactor types - replace Partial<IPactCommand> with PartialPactCommand


packages/libs/client/src/interfaces/type-utilities.ts
```diff
@@ -39,3 +39,15 @@ export type ExtractCapabilityType<TCommand> = TCommand extends {
       : IGeneralCapability
     : IGeneralCapability
   : IGeneralCapability;
+
+export type Prettify<T extends Record<string, unknown>> = {
```
sanderlooijenga, line 43: Do/should we have tests for these utility types?

packages/libs/client/src/composePactCommand/composePactCommand.ts
```diff
@@ -7,29 +7,33 @@ type NoPayload<TCommand> = TCommand extends { payload: unknown }
 
 type DataOrFunction<TData> = TData | ((a: TData) => TData);
 
-type InitialInput = Partial<IPactCommand> | (() => Partial<IPactCommand>);
+type InitialInput =
+  | Partial<IPartialPactCommand>
```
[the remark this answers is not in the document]
javadkh2: good catch

packages/libs/client/src/interfaces/type-utilities.ts
javadkh2: good idea we can write tests for them

EnoF (approved):

alber70g (approved):

PR #1227, opened 2023-11-15
fix(@kadena/react-ui): Update the Notification Component

eileenmguo:
Updated the Notification component with the following: 
- Updated the `icon` prop to accept ReactNode instead of an icon name. The reason is to decouple the Notification component from the Icons so that when using the Notification component, not all of the Icons will be part of the bundle. When an Icon is not provided, the `Information` icon will be shown.
- Moved the `title` prop to a subcomponent `Heading`. This matches the subcomponent structure better 
- Removed the expanded prop since all use cases used the 100% width option
- Removed the `outlined` prop and added a `styledVariant` prop which has the `bordered` and `borderless` options
- Added a required `role` prop which can be set to `alert`, `status`, or `none`. In most cases that I saw, the `status` role is set, however I opted to not give a default value as to ensure the user thinks about its purpose. We can give this a default value if preferred though
- Removed the inline option all together. I discussed this with Isa as a way to simplify this component

Note: I've decided to keep the icon as a prop instead of a subcomponent because the icon will always be part of the card and because I think it makes the composition more clean. Right now most icons are passed in as a prop, but if we want to make it a subcomponent as well, we should discuss this and update this across the board


nil-amrutlal-dept (approved):

nil-amrutlal-dept (approved):
Looks good on `graph-client`
image

packages/apps/tools/src/components/Global/FormStatusNotification/index.tsx
```diff
@@ -79,11 +79,12 @@ export const FormStatusNotification: FC<IFormStatusNotificationProps> = (
         color={statusToColorMap[status!]}
         hasCloseButton
         icon={statusToIconMap[status!]}
-        expanded
-        title={title ?? titles[status!]}
         onClose={onNotificationCloseClick}
-        variant="outlined"
+        role="status"
       >
+        {title ? (
```
sanderlooijenga, line 85: This is not the correct replacement for `title ?? titles[status!]`

packages/apps/tools/src/pages/faucet/existing/index.tsx
```diff
@@ -24,7 +24,7 @@ import Head from 'next/head';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
 import type { FC } from 'react';
-import React, { useCallback, useState } from 'react';
```
sanderlooijenga, line 27: Why is `React` removed as an import?

packages/apps/tools/src/components/Global/FormStatusNotification/index.tsx
eileenmguo: I think my brain registered && haha

packages/apps/tools/src/pages/faucet/existing/index.tsx
eileenmguo: unintentional. I have re-added it 👍

sanderlooijenga (approved):

sstraatemans (approved):

PR #1243, opened 2023-11-21
feat(@kadena/react-ui): Tooltip Component

eileenmguo:
Updates the tooltip component to use react-aria and already updates the tools app. 

NOTE: Design changes have been made with Isa


packages/apps/immutable-records/next-env.d.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference types="next" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
```
ferreroltd, line 4: what is this file for?

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,21 +1,47 @@
...
+    config,
+    state,
+    ref,
+  );
+
+  const { tooltipProps } = useTooltip(baseTooltipProps, state);
+
+  return (
+    <Box position="relative">
+      {cloneElement(children as React.ReactElement, { ...triggerProps, ref })}
```
ferreroltd, line 38: didnt we want to avoid using this? or is this the only way to pass the props?

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,21 +1,47 @@
...
+
+      {state.isOpen && (
+        <span className={tooltipPositionVariants[position]} {...tooltipProps}>
```
ferreroltd, line 41: niceeee variants!

ferreroltd (approved):

packages/apps/immutable-records/next-env.d.ts
eileenmguo: I need to remove this! It must have been a local file

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
eileenmguo: We want to use this sparingly and in a lot of previous cases, using context was better, but in this case it is the simplest option since we don't want the user to have to access a context

packages/libs/react-ui/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -1,78 +1,120 @@
...
+import { Button } from '@components/Button';
+import { Box, Stack } from '@components/Layout';
 import type { Meta, StoryObj } from '@storybook/react';
 import { withCenteredStory } from '@utils/withCenteredStory';
-import React, { useRef } from 'react';
-import type { ITooltipProps } from './';
-import { Tooltip } from './';
-import { container } from './stories.css';
+import React from 'react';
+import { SystemIcon } from '..';
```
salamaashoush, line 6: this could lead to circular dependencies, generally we should avoid importing from the barrel files in the files exported from the barrel file :D

salamaashoush: aha, it is a story file, but still good idea to just import from the icon folder directly

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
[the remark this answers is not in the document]
eileenmguo: ahh nice 👍

packages/libs/react-ui/src/components/Tooltip/Tooltip.stories.tsx
eileenmguo: Yeah, it was leftover from an automatic import, but I updated it

packages/libs/react-ui/src/styles/vars.css.ts
```diff
@@ -53,7 +53,7 @@ export const vars = createGlobalTheme(':root', {
   },
   shadows: {
     // TODO: Update to match design system
```
sanderlooijenga, line 55: Is this todo now complete?

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,21 +1,52 @@
...
+import { Box } from '@components/Layout';
+import { mergeRefs } from '@react-aria/utils';
+import type { FC, ReactElement, ReactNode } from 'react';
+import React, { cloneElement, useRef } from 'react';
+import { useTooltip, useTooltipTrigger } from 'react-aria';
+import type { TooltipTriggerProps } from 'react-stately';
+import { useTooltipTriggerState } from 'react-stately';
+import { tooltipPositionVariants } from './Tooltip.css';
+export interface ITooltipProps extends Omit<TooltipTriggerProps, 'trigger'> {
+  children: ReactElement;
```
sanderlooijenga, line 10: Just double checking; we need to use `ReactElement` here as opposed to `ReactNode`?

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,21 +1,52 @@
...
-    );
-  },
-);
+export const Tooltip: FC<ITooltipProps> = ({
+  children,
+  content,
+  position = 'right',
+  ...props
+}) => {
+  const config = {
```
sanderlooijenga, line 21: Should we give this a type?

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -2,82 +2,86 @@ import { sprinkles } from '@theme/sprinkles.css';
...
     pointerEvents: 'none',
+    width: 'max-content',
+    maxWidth: '$maxContentWidth',
   }),
   {
-    rotate: '45deg',
+    ':before': {
+      content: '',
+      position: 'absolute',
+      borderTop: '6px solid transparent',
```
sanderlooijenga, line 22: I guess 6px is not a defined size in our tokens?

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
```diff
@@ -21,21 +21,12 @@ export const HoverTag = ({
...
-      <span
-        data-testid="kda-hover-tag"
-        onMouseEnter={(e: React.MouseEvent<HTMLElement>) =>
-          Tooltip.handler(e, ref)
-        }
-        onMouseLeave={(e: React.MouseEvent<HTMLButtonElement>) =>
-          Tooltip.handler(e, ref)
-        }
-      >
+    <Tooltip position="top" content={<code>tooltipContent</code>}>
```
sanderlooijenga, line 28: This needs to be `{tooltipContent}`

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
eileenmguo: No it is not, but also I don't think this one should be a token as it is very specific and we don't want it to change based on the browser fontsize

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
eileenmguo: Yes, because we are applying a ref and props to the child element which means it should only wrap one element. This guards against the user doing something like this:
```
<Tooltip content="example">
  <Button />
  <Button />
  <Button />
</Tooltip>
```

We are expecting that the tooltip is applied to only one item

packages/apps/tools/src/components/Global/HoverTag/HoverTag.tsx
eileenmguo: oops! will update, thanks 👍

packages/libs/react-ui/src/components/Tooltip/Tooltip.tsx
eileenmguo: It is implicitly typed since the props are already typed. I see some of them are missing from storybook though so I will add them 👍

packages/libs/react-ui/src/styles/vars.css.ts
eileenmguo: No, we're not actually using this in our sprinkles. I'm not sure if we will add shadows with the new token update, but I decided to leave this here for now

sanderlooijenga: Alrighty

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
sanderlooijenga: Alright, thanks for clarifying 👌

sanderlooijenga:
Somewhat of a more general question; do we expect the tooltip to close when _clicking_ the handler, looking at storybook examples? 🤔

eileenmguo:
> Somewhat of a more general question; do we expect the tooltip to close when _clicking_ the handler, looking at storybook examples? 🤔

That is only in the controlled state. The tooltip will appear and disappear based on hover or focus by default

salamaashoush (approved):

PR #1252, opened 2023-11-22
[@kadena/graph] Improved naming convention of Reconcile object

MRVDH:

### DIFF
--- a/packages/apps/graph/prisma/schema.prisma
+++ b/packages/apps/graph/prisma/schema.prisma
@@ -151,21 +151,21 @@ model schema_migrations {
 
 /// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by Prisma Client.
 view reconcile {
-  token             String?
-  amount            Int?
-  sender            String?
-  sender_current    Int?
-  sender_previous   Int?
-  receiver          String?
-  receiver_current  Int?
-  receiver_previous Int?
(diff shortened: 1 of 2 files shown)


nil-amrutlal-dept (approved):

packages/apps/graph/prisma/schema.prisma
```diff
@@ -151,21 +151,21 @@ model schema_migrations {
...
+  receiverAccount        String? @map("receiver")
+  receiverCurrentAmount  Int?    @map("receiver_current")
+  receiverPreviousAmount Int?    @map("receiver_previous")
+  chainId                BigInt? @map("chainid")
+  height                 BigInt?
+  requestKey             String  @map("requestkey") @db.VarChar
+  orderIndex             BigInt  @map("idx")
+  blockHash              String  @map("block") @db.VarChar
+  eventId                Int?    @map("id")
+  version                String?
```
alber70g, line 168: Should we maybe add `_amount` to the view itself? Then mapping only becomes the snake_case to camelCase conversion

alber70g (approved):

packages/apps/graph/prisma/schema.prisma
MRVDH: It doesn't automatically do the conversion to camelCase unfortunately, so it wouldn't really matter

PR #1253, opened 2023-11-23
feat(kadena-cli): account commands

jessevanmuijden:
Depends on support for any fungible in @kadena/client-utils. https://github.com/kadena-community/kadena.js/pull/1245


nillo (commented):
Can you separate the changes to libs/client-utils from this PR

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -25,6 +25,48 @@ import { ensureDevnetsConfiguration } from './helpers.js';
...
+      return publicKeys.split(',').map((value) => value.trim());
+    },
+  }),
+  amount: createOption({
+    key: 'amount' as const,
+    prompt: account.amountPrompt,
+    validation: z
+      .string({
+        /* eslint-disable-next-line @typescript-eslint/naming-convention */
+        invalid_type_error: 'Error: -c, --chain-id must be a number',
```
barthuijgen, line 53: this says chain-id but it is in the amount option

packages/tools/kadena-cli/src/prompts/account.ts
```diff
@@ -1,36 +1,36 @@
...
-//   if (isOptional) {
-//     choices.push({
-//       value: 'skip',
-//       name: 'Account is optional. Continue to next step',
-//     });
-//   }
+export const amountPrompt: IPrompt = async () =>
+  await input({
+    validate(value) {
+      return !isNaN(parseFloat(value));
```
barthuijgen, line 17: I usually prefer to add a `.replace(',', '.')` when using parseFloat() on user input so that it accepts both.

packages/tools/kadena-cli/src/account/commands/accountCreate.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { Command } from 'commander';
+import debug from 'debug';
+import { FAUCET_CONSTANTS } from '../../constants/faucet.js';
+import { createCommand } from '../../utils/createCommand.js';
+import { globalOptions } from '../../utils/globalOptions.js';
+
+export const createAccountCommand: (program: Command, version: string) => void =
+  createCommand(
+    'create',
+    'Create account',
```
barthuijgen, line 15: This can be more descriptive I think?

packages/tools/kadena-cli/src/account/commands/accountCreate.ts
```diff
@@ -0,0 +1,93 @@
...
+      globalOptions.predicate(),
+    ],
+    async (config) => {
+      debug('account-create:action')({ config });
+
+      try {
+        const principal = await createPrincipal(
+          {
+            keyset: {
+              pred: config.predicate as 'keys-all' | 'keys-2' | 'keys-any',
```
barthuijgen, line 30: The Type `'keys-all' | 'keys-2' | 'keys-any'` is used 4 times in this PR. That should probably be defined somewhere.

That said does `@kadena/client` not provide this type anywhere? I also noticed differences between for example `keys-2` and `keys-two` across branches. What does it really need to be?

packages/tools/kadena-cli/package.json
```diff
@@ -47,7 +47,7 @@
     "@kadena/pactjs-generator": "workspace:*",
     "@scure/bip39": "^1.2.1",
     "cardano-crypto.js": "https://github.com/kadena-io/cardano-crypto.js.git#jam@chainweaver-keygen",
-    "chalk": "^4.1.2",
+    "chalk": "^5.2.0",
```
nillo, line 50: lets make sure these changes to the packages don't break anything ( yes we should be up to date, but donno if there are breaking changes )

nillo (approved):
LGTM

barthuijgen (approved):

PR #1258, opened 2023-11-23
[@kadena/graph] Tracing and Analysing

MRVDH:
image

### DIFF
--- a/packages/apps/graph/.env.example
+++ b/packages/apps/graph/.env.example
@@ -9,3 +9,6 @@ MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH=11
 
 # Database configuration. The default value is postgresql://devnet@localhost:5432/devnet.
 DATABASE_URL="postgresql://devnet@localhost:5432/devnet"
+
+# Tracing
+ENABLE_TRACING=false
--- a/packages/apps/graph/.env.example
+++ b/packages/apps/graph/.env.example
@@ -3,8 +3,8 @@ CHAIN_COUNT=20
 # Database configuration. The default value is postgresql://devnet@localhost:5432/devnet.
 DATABASE_URL="postgresql://devnet@localhost:5432/devnet"
 
-# This variable determines the maximum calculated block confirmation depth. The default is 51.
-MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH=51
+# This variable determines the maximum calculated block confirmation depth.
+MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH=11
 
 # The following variables are used to configure the network endpoints for the GraphQL API. The default values are localhost:8080 and fast-development.
 NETWORK_HOST="localhost:8080"
--- a/packages/apps/graph/.env.example
+++ b/packages/apps/graph/.env.example
@@ -1,14 +1,15 @@
 CHAIN_COUNT=20
 
+# Database configuration. The default value is postgresql://devnet@localhost:5432/devnet.
+DATABASE_URL="postgresql://devnet@localhost:5432/devnet"
+
+# This variable determines the maximum calculated block confirmation depth. The default is 51.
+MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH=51
+
 # The following variables are used to configure the network endpoints for the GraphQL API. The default values are localhost:8080 and fast-development.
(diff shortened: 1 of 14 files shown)


packages/apps/graph/generated-schema.graphql
```diff
@@ -172,7 +172,7 @@ type Query {
   block(hash: String!): Block
   blocksFromHeight(chainIds: [Int!], startHeight: Int!): [Block!]!
   chainAccount(accountName: String!, chainId: String!, moduleName: String!): ChainModuleAccount
-  completedBlockHeights(chainIds: [String!], completedHeights: Boolean, heightCount: Int): [Block!]!
+  completedBlockHeights(chainIds: [Int!], completedHeights: Boolean, heightCount: Int): [Block!]!
```
alber70g, line 175: Why is this changed? In @kadena/client we reference by string as it could be different in the future

packages/apps/graph/src/graph/builder.ts
```diff
@@ -75,6 +72,24 @@ export const builder = new SchemaBuilder<
     // use where clause from prismaRelatedConnection for totalCount
     filterConnectionTotalCount: true,
   },
+
+  relayOptions: {
+    clientMutationId: 'optional',
+    cursorType: 'String',
+  },
+
+  ...(process.env.ENABLE_TRACING === 'true' && {
```
alber70g, line 81: This can be moved to `utils/dotenv`?

packages/apps/graph/src/services/tracing/trace-analyser.ts
```diff
@@ -0,0 +1,131 @@
...
+  )
+  .addOption(
+    new Option(
+      '-s, --sort <string>',
+      'Sort the queries by the given property',
+    ).default('median'),
+  )
+  .action(async (args) => {
+    try {
+      readFileSync('traces.log');
```
alber70g, line 21: How about passing the filepath as an option. Makes it easier to have multiple tracelogs and ability to compare them

alber70g (approved):

packages/apps/graph/src/services/tracing/trace-analyser.ts
MRVDH: Good idea, done!

packages/apps/graph/src/graph/builder.ts
MRVDH: Indeed, moved!

nil-amrutlal-dept (approved):

packages/apps/graph/.env.example
```diff
@@ -1,11 +1,15 @@
 CHAIN_COUNT=20
 
+# Database configuration. The default value is postgresql://devnet@localhost:5432/devnet.
+DATABASE_URL="postgresql://devnet@localhost:5432/devnet"
+
+# This variable determines the maximum calculated block confirmation depth. The default is 51.
+MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH=51
```
nil-amrutlal-dept, line 7: Can we leave this at 11? This will make the queries faster and I feel that a number much greater than 6 is not that useful anyway

packages/apps/graph/src/utils/dotenv.ts
```diff
@@ -4,22 +4,26 @@ _dotenv.config();
...
-  MAX_BLOCK_DEPTH: parseInt(
-    or(process.env.MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH, '51'),
-    10,
-  ),
   DATABASE_URL: or(
     process.env.DATABASE_URL,
     'postgresql://devnet@localhost:5432/devnet',
   ),
+  MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH: parseInt(
+    or(process.env.MAX_CALCULATED_BLOCK_CONFIRMATION_DEPTH, '51'),
```
nil-amrutlal-dept, line 20: Same here, maybe default to 11

packages/apps/graph/.env.example
MRVDH: Updated!

packages/apps/graph/src/utils/dotenv.ts
MRVDH: Updated!

PR #1259, opened 2023-11-24
First setup of Integration tests between Kadena/Graph and devnet.

Ghislain89:
Draft PR for an integration test package. Initially for Graph.


packages/tools/integration-tests/src/projects/graph/testdata/constants/accounts.ts
```diff
@@ -0,0 +1,29 @@
+import type { ChainId } from '@kadena/types';
+
+export interface IAccount {
+  account: string;
+  publicKey: string;
```
nil-amrutlal-dept, line 5: This could be an array of strings, similar to how it's done in the simulation. The reasoning is that an account can have multiple keys (w: accounts). Also, it would be easier to test transfers from a multi-sig account

nil-amrutlal-dept (approved):

packages/tools/integration-tests/src/projects/graph/testdata/constants/accounts.ts
Ghislain89: I had not taken this scenario in to account (No pun intended). I think this will be a needed update as soon as I actually move towards w:accounts.  Are these commonly used? For now I'd like to focus on expected common use cases.

nil-amrutlal-dept: I am not sure how common they are in the real world, but I have included multi-sig accounts in the traffic simulation for the devnet. The simulation currently has 3 types of transfer: transfer (same chain), cross-chain, and safe-transfers, and all of these can be from/to multi-sig accounts (since it is random, it can happen that a given simulation does not have w: accounts but it is unlikely).

nil-amrutlal-dept: But it does sound good :) Start with single key accounts, and in the next iteration you could adapt it to include multi-sig accounts

alber70g (approved):
I reviewed the files assigned to me by CODEOWNERS

PR #1260, opened 2023-11-24
Let react-hook-form take care of validation

sanderlooijenga:


KristinaSpasevska (approved):

packages/apps/tools/locales/en/common.json
```diff
@@ -63,6 +63,8 @@
   "Testnet": "Testnet",
   "Cross Chain Transfers": "Cross Chain Transfers",
   "Invalid length of request key": "Invalid length of request key",
+  "invalid-pub-key-length": "Invalid length of public key",
```
jessevanmuijden, line 66: it looks like the other translation keys are not using kebab-case. Maybe good to keep it consistent.

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -87,58 +92,45 @@ const NewAccountFaucetPage: FC = () => {
...
-      if (pubKeys.length === 0) return '';
-      return await createPrincipal(pubKeys, chainID, pred);
-    },
-    initialData: '',
-  });
-
-  const schema = z.object({
-    name: z.string(),
-    pubKey: z.string().optional(),
+    queryFn: () => createPrincipal(pubKeys, chainID, pred),
```
jessevanmuijden, line 98: I added a third createPrincipal implementation in this PR: https://github.com/kadena-community/kadena.js/pull/1253/files#diff-64a159ddfe2566f68001bcacb9dbbb1e24e4bc1bb85c1e4794d6836b0fdbfafa. I do not think it is necessary to check for the signature in the tools implementations, since you are doing a dirtyRead for the transaction, but I may be wrong.

PR #1263, opened 2023-11-27
fix(pactjs-cli): send hash with txs

javadkh2:
Updated pactjs-cli to send hash with txs;  this will fix the issue after chainweb-node is updated


nillo:
i think @alber70g added this project to kadena CLI, so this update so probably added to the CLI code also, 
@jessevanmuijden if this is the case, when refactoring to the new createCommand, please take along this change.

If im mixing 2 things, please disregard this comment :)

PR #1272, opened 2023-11-28
[@kadena/graph] Deploy Marmalade in Devnet (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206142839462225


packages/apps/graph/src/devnet/transfer.ts
```diff
@@ -63,8 +63,11 @@ export async function transfer({
       ttl: 8 * 60 * 60, //8 hours
     })
     .setNetworkId(devnetConfig.NETWORK_ID)
+
     .createTransaction();
 
+  console.log(transaction);
```
MRVDH, line 69: can be removed?

MRVDH (commented):
My review is a bit shallow. I think I lack the full knowledge around the whole marmalade deployment process to be able to properly review everything. Maybe a good idea to add to the readme how it works, and then I can go more in depth to see if the code is doing the same and what certain assumptions and trade-offs have been made.

nil-amrutlal-dept:
> My review is a bit shallow. I think I lack the full knowledge around the whole marmalade deployment process to be able to properly review everything. Maybe a good idea to add to the readme how it works, and then I can go more in depth to see if the code is doing the same and what certain assumptions and trade-offs have been made.

@MRVDH readme.md added

PR #1279, opened 2023-11-30
Use new typography tokens

salamaashoush:
1. Fix typography tokens and regenerate the styles. 
2. Update `Text` and `Heading` components to use the new typography tokens


packages/libs/react-ui/src/components/Typography/Heading/Heading.tsx
```diff
@@ -1,78 +1,72 @@
...
+  fontH2Regular,
+  fontH3Bold,
+  fontH3Regular,
+  fontH4Bold,
+  fontH4Regular,
+  fontH5Bold,
+  fontH5Regular,
+  fontH6Bold,
+  fontH6Regular,
+} from '../../../styles';
```
eileenmguo, line 17: you can also write this like the following
```suggestion
} from '@styles/tokens/styles';
```

packages/libs/react-ui/src/styles/vars.css.ts
```diff
@@ -295,7 +324,22 @@ export const darkThemeClass = createTheme(vars.colors, {
...
+  old: vars.colors,
+};
+export const darkThemeClass = createTheme(
+  darkContract,
+  {
+    new: darkThemeValues.kda.foundation.color,
+    old: oldDarkThemeColors,
+  },
+  'dark',
+);
```
eileenmguo, line 342: So it looks like you're referencing the vars created from the original "old" and "updated" contracts in a new contract that only updates the colors via one classname. Is that correct?

packages/libs/react-ui/src/components/Typography/Heading/Heading.tsx
salamaashoush: I know, but this one is clearer and I also don't like the false negative linting error in my editor also it easier to use the relative path to make it clear where is the import is coming from and also make go to the declaration editor functionality work as expected

packages/libs/react-ui/src/components/Typography/Text/Text.tsx
```diff
@@ -2,51 +2,76 @@ import cn from 'classnames';
...
+} from '../../../styles';
+import { colorVariants, transformVariants } from '../typography.css';
+
+// eslint-disable-next-line @kadena-dev/typedef-var
+export const TEXT_ELEMENTS = ['p', 'span', 'code'] as const;
+export type TextElementType = (typeof TEXT_ELEMENTS)[number];
+
+type TextVariant = 'small' | 'smallest' | 'base';
+function getFontClass(
+  variant: TextVariant,
```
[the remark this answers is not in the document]
salamaashoush: because the size infers it only changes the font size but it is not true since it  also changes other font properties, so it is a typography variant not only a size

packages/libs/react-ui/src/styles/index.ts
```diff
@@ -2,8 +2,6 @@ export { sprinkles } from './sprinkles.css';
 export type { Sprinkles } from './sprinkles.css';
 export { breakpoints, mapToProperty, responsiveStyle } from './themeUtils';
 export { tokens } from './tokens/contract.css';
-export { darkTheme } from './tokens/dark.css';
-export { lightTheme } from './tokens/light.css';
 export * from './tokens/styles.css';
```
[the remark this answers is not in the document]
salamaashoush: this won't affect tree shaking in any way since it is a re-export and it is the same as re-exporting one by one expert this one is less verbose

eileenmguo (approved):
Code looks good to me! The colors still need to be updated to use the new color tokens, but that can be done after the new utility classes are merged

sstraatemans (approved):

PR #1288, opened 2023-12-01
[@kadena/graph] Query, Object and Field descriptions

MRVDH:
Please check if all of the definitions of the queries, objects and the fields are correct.

### DIFF
--- a/packages/apps/graph/generated-schema.graphql
+++ b/packages/apps/graph/generated-schema.graphql
@@ -18,7 +18,7 @@ type Block implements Node {
   hash: ID!
   height: BigInt!
   id: ID!
-  minerAccount: ChainModuleAccount!
+  minerAccount: ChainFungibleAccount!
   parent: Block
   parentHash: String
   payloadHash: String!
@@ -41,35 +41,35 @@ type BlockTransactionsConnectionEdge {
 }
 
 """A chain- and fungible-specific account."""
-type ChainModuleAccount implements Node {
+type ChainFungibleAccount implements Node {
   accountName: String!
   balance: Float!
   chainId: ID!
+  fungibleName: String!
   guard: Guard!
   id: ID!
-  moduleName: String!
-  transactions(after: String, before: String, first: Int, last: Int): ChainModuleAccountTransactionsConnection!
-  transfers(after: String, before: String, first: Int, last: Int): ChainModuleAccountTransfersConnection!
+  transactions(after: String, before: String, first: Int, last: Int): ChainFungibleAccountTransactionsConnection!
+  transfers(after: String, before: String, first: Int, last: Int): ChainFungibleAccountTransfersConnection!
 }
 
-type ChainModuleAccountTransactionsConnection {
-  edges: [ChainModuleAccountTransactionsConnectionEdge!]!
+type ChainFungibleAccountTransactionsConnection {
+  edges: [ChainFungibleAccountTransactionsConnectionEdge!]!
   pageInfo: PageInfo!
   totalCount: Int!
 }
 
-type ChainModuleAccountTransactionsConnectionEdge {
+type ChainFungibleAccountTransactionsConnectionEdge {
   cursor: String!
   node: Transaction!
 }
 
-type ChainModuleAccountTransfersConnection {
-  edges: [ChainModuleAccountTransfersConnectionEdge!]!
+type ChainFungibleAccountTransfersConnection {
+  edges: [ChainFungibleAccountTransfersConnectionEdge!]!
   pageInfo: PageInfo!
   totalCount: Int!
 }
 
-type ChainModuleAccountTransfersConnectionEdge {
+type ChainFungibleAccountTransfersConnectionEdge {
   cursor: String!
   node: Transfer!
 }
@@ -107,6 +107,39 @@ type Event implements Node {
   transaction: Transaction
 }
 
+"""A fungible-specific account."""
+type FungibleAccount implements Node {
+  accountName: String!
+  chainAccounts: [ChainFungibleAccount!]!
+  fungibleName: String!
+  id: ID!
+  totalBalance: Decimal!
+  transactions(after: String, before: String, first: Int, last: Int): FungibleAccountTransactionsConnection!
+  transfers(after: String, before: String, first: Int, last: Int): FungibleAccountTransfersConnection!
+}
+
+type FungibleAccountTransactionsConnection {
+  edges: [FungibleAccountTransactionsConnectionEdge!]!
+  pageInfo: PageInfo!
+  totalCount: Int!
+}
+
+type FungibleAccountTransactionsConnectionEdge {
+  cursor: String!
+  node: Transaction!
+}
+
+type FungibleAccountTransfersConnection {
+  edges: [FungibleAccountTransfersConnectionEdge!]!
+  pageInfo: PageInfo!
+  totalCount: Int!
+}
+
+type FungibleAccountTransfersConnectionEdge {
+  cursor: String!
+  node: Transfer!
+}
+
 """General information about the graph and chainweb-data."""
 type GraphConfiguration {
   """The maximum number of confirmations calculated on this endpoint."""
@@ -128,39 +161,6 @@ type MinerKey implements Node {
   blockHash: String!
   id: ID!
   key: String!
-}
-
-"""A fungible-specific account."""
-type ModuleAccount implements Node {
-  accountName: String!
-  chainAccounts: [ChainModuleAccount!]!
-  id: ID!
-  moduleName: String!
-  totalBalance: Decimal!
-  transactions(after: String, before: String, first: Int, last: Int): ModuleAccountTransactionsConnection!
-  transfers(after: String, before: String, first: Int, last: Int): ModuleAccountTransfersConnection!
-}
-
-type ModuleAccountTransactionsConnection {
-  edges: [ModuleAccountTransactionsConnectionEdge!]!
-  pageInfo: PageInfo!
-  totalCount: Int!
-}
-
-type ModuleAccountTransactionsConnectionEdge {
-  cursor: String!
-  node: Transaction!
-}
-
-type ModuleAccountTransfersConnection {
-  edges: [ModuleAccountTransfersConnectionEdge!]!
-  pageInfo: PageInfo!
-  totalCount: Int!
-}
-
-type ModuleAccountTransfersConnectionEdge {
-  cursor: String!
-  node: Transfer!
 }
 
 interface Node {
@@ -196,7 +196,7 @@ scalar PositiveFloat
 
 type Query {
   """Retrieve an account by its name and fungible, such as coin."""
-  account(accountName: String!, moduleName: String!): ModuleAccount
+  account(accountName: String!, fungibleName: String!): FungibleAccount
 
   """Retrieve a block by hash."""
   block(hash: String!): Block
@@ -207,7 +207,7 @@ type Query {
   """
   Retrieve an account by its name and fungible, such as coin, on a specific chain.
   """
-  chainAccount(accountName: String!, chainId: String!, moduleName: String!): ChainModuleAccount
+  chainAccount(accountName: String!, chainId: String!, fungibleName: String!): ChainFungibleAccount
 
   """Retrieve all completed blocks from a given height."""
   completedBlockHeights(chainIds: [String!] = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19"], completedHeights: Boolean = false, heightCount: Int = 3): [Block!]!
@@ -237,13 +237,13 @@ type Query {
   pactQuery(pactQuery: PactQuery!): String!
 
   """Retrieve transactions."""
-  transactions(accountName: String, after: String, before: String, blockHash: String, chainId: String, first: Int, last: Int, moduleName: String, requestKey: String): QueryTransactionsConnection!
+  transactions(accountName: String, after: String, before: String, blockHash: String, chainId: String, first: Int, fungibleName: String, last: Int, requestKey: String): QueryTransactionsConnection!
 
   """Retrieve all transactions by a given public key."""
   transactionsByPublicKey(after: String, before: String, first: Int, last: Int, publicKey: String!): QueryTransactionsByPublicKeyConnection!
 
   """Retrieve transfers."""
-  transfers(accountName: String, after: String, before: String, chainId: String, first: Int, last: Int, moduleName: String): QueryTransfersConnection!
+  transfers(accountName: String, after: String, before: String, chainId: String, first: Int, fungibleName: String, last: Int): QueryTransfersConnection!
 }
 
 type QueryTransactionsByPublicKeyConnection {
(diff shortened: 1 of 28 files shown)


packages/apps/graph/generated-schema.graphql
```diff
@@ -3,8 +3,11 @@ The `BigInt` scalar type represents non-fractional signed whole numeric values.
 """
 scalar BigInt
 
+"""A unit of information that stores a set of verified transactions."""
```
nil-amrutlal-dept, line 6: Would this also be true for an orphan block

packages/apps/graph/src/graph/objects/GraphConfiguration.ts
```diff
@@ -1,12 +1,16 @@
 import { builder } from '../builder';
 
 export default builder.objectType('GraphConfiguration', {
-  description: 'General information about the graph and chainweb-data',
+  description: 'General information about the graph and chainweb-data.',
   fields: (t) => ({
-    maximumConfirmationDepth: t.exposeInt('maximumConfirmationDepth'),
+    maximumConfirmationDepth: t.exposeInt('maximumConfirmationDepth', {
+      description:
+        'The maximum confirmation depth configured in the environment variables.',
```
nil-amrutlal-dept, line 8: same as above

packages/apps/graph/src/graph/objects/Block.ts
```diff
@@ -99,6 +103,7 @@ export default builder.prismaNode('Block', {
           throw normalizeError(error);
         }
       },
+      description: 'The number of blocks that confirm this block.',
```
nil-amrutlal-dept, line 106: same as above

packages/apps/graph/generated-schema.graphql
```diff
@@ -31,6 +36,7 @@ type BlockTransactionsConnectionEdge {
   node: Transaction!
 }
 
+"""An account on a certain chain on a certain module, such as coin."""
```
alber70g, line 39: _A chain specific fungible account_?

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
 """Floats that will have a value of 0 or more."""
 scalar Decimal
 
+"""A record of an execution of a function on a smart contract."""
```
alber70g, line 81: _An event emitted by the execution of a smart-contract function_

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
...
 type Event implements Node {
   block: Block!
   chainId: BigInt!
+
+  """The block height of this event."""
```
alber70g, line 86: _The height of the block where the event was emitted_

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
...
   moduleName: String!
   name: String!
+
+  """
+  The order index of this event, in the case that there are multiple events.
+  """
   orderIndex: BigInt!
   parameterText: String!
+
+  """The module name and the event name combined."""
```
alber70g, line 98:
```suggestion
  """The full eventname, containing module and eventname, e.g. `coin.TRANSFER`"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
...
 
-"""General information about the graph and chainweb-data"""
+"""General information about the graph and chainweb-data."""
 type GraphConfiguration {
+  """
+  The maximum confirmation depth configured in the environment variables.
+  """
   maximumConfirmationDepth: Int!
+
+  """The minimum block height that is available in the graph."""
```
alber70g, line 111:
```suggestion
  """The lowest block-height that is indexed in this endpoint"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
...
   minimumBlockHeight: BigInt
 }
 
+"""Access rule for a certain smart contract or account."""
```
alber70g, line 115:
```suggestion
"""Guard for an account."""
```
I would ignore the smart-contract part for now. We use it only on accounts here.

packages/apps/graph/generated-schema.graphql
```diff
@@ -72,38 +78,55 @@ scalar DateTime
...
 
+"""The public key of the miner that solved a block."""
 type MinerKey implements Node {
   block: Block!
   blockHash: String!
   id: ID!
   key: String!
 }
 
+"""An account on a certain module, such as coin."""
```
alber70g, line 129:
```suggestion
"""A fungible-specific account"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
 scalar PositiveFloat
 
 type Query {
-  account(accountName: String!, moduleName: String!): ModuleAccount!
+  """Find an account by its name and its module, such as coin."""
```
alber70g, line 194:
```suggestion
  """Retrieve an account by its name and fungible, such as coin."""
```

Here we also might want to change `moduleName` to `fungible`. Same for `ModuleAccount` -> `FungibleAccount`

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
...
+  account(accountName: String!, moduleName: String!): ModuleAccount
+
+  """Find a block by its hash."""
```
alber70g, line 197:
```suggestion
  """Retrieve a block by hash."""
```
I dont know wheter all `find`s can be `retrieve`s. Find sounds more "voluntary". Bikeshedding though :)

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
...
   block(hash: String!): Block
+
+  """Find all blocks from a given height."""
   blocksFromHeight(chainIds: [String!], startHeight: Int!): [Block!]!
+
+  """
+  Find an account by its name and its module, such as coin, on a specific chain.
+  """
```
alber70g, line 205:
```suggestion
  """
  Retrieve an account by its name, for a fungible, on a chain
  """
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
...
+  """Estimate the gas limit for a transaction."""
   gasLimitEstimate(transaction: PactTransaction!): Int!
+
+  """Estimate the gas limit for a list of transactions."""
   gasLimitEstimates(transactions: [PactTransaction!]!): [Int!]!
+
+  """Get the configuration of the graph."""
   graphConfiguration: GraphConfiguration!
+
+  """Get the height of the latest block."""
```
alber70g, line 220:
```suggestion
  """Get the height of the block with the highest height."""
```
It's a subtle difference, but we don't retreive the `last` block height, but the highest block height

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
...
   lastBlockHeight: BigInt
   node(id: ID!): Node
   nodes(ids: [ID!]!): [Node]!
+
+  """Send a list of raw pact queries."""
```
alber70g, line 225: Execute arbitrary Pact code via a `local` call without gas-estimation or signature-verification (e.g. `(+ 1 2)` or `(coin.get-details <account>`

packages/apps/graph/generated-schema.graphql
```diff
@@ -168,21 +191,50 @@ type PageInfo {
...
   pactQueries(pactQuery: [PactQuery!]!): [String!]!
+
+  """Send a raw pact query."""
```
alber70g, line 228: Idem

packages/apps/graph/generated-schema.graphql
```diff
@@ -219,37 +271,61 @@ type QueryTransfersConnectionEdge {
   node: Transfer!
 }
 
+"""Information about a public key that can sign transactions."""
```
alber70g, line 274:
```suggestion
"""A signer for a specific transaction"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -219,37 +271,61 @@ type QueryTransfersConnectionEdge {
...
+  """Subscribe to new blocks."""
   newBlocks(chainIds: [Int!]): [Block!]
+
+  """
+  Subscribe to a request key to wait for the related transaction to appear.
+  """
   transaction(requestKey: String!): Transaction
 }
 
+"""A request to execute a smart contract function."""
```
alber70g, line 301:
```suggestion
"""A confirmed transaction."""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -219,37 +271,61 @@ type QueryTransfersConnectionEdge {
...
 type Transaction implements Node {
+  """The JSON stringified error message if the transaction failed."""
   badResult: String
   block: Block
   chainId: BigInt!
+
+  """The PACT code that is executed."""
```
alber70g, line 308:
```suggestion
  """
  The Pact expressions executed in this transaction when it's a `exec` transaction. 
  For a continuation, this field is `cont`
  """
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -219,37 +271,61 @@ type QueryTransfersConnectionEdge {
...
   code: String!
+
+  """
+  The JSON stringified continuation in the case that it is a continuation.
+  """
   continuation: String
   creationTime: DateTime!
+
+  """The JSON stringified data that is related to the request."""
```
alber70g, line 317:
```suggestion
  """The `data` field of a transaction. Formatted as raw JSON"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -219,37 +271,61 @@ type QueryTransfersConnectionEdge {
...
   data: String
   eventCount: BigInt
   events: [Event!]
   gas: BigInt!
   gasLimit: BigInt!
   gasPrice: Float!
+
+  """The JSON stringified result if the transaction was successful."""
```
alber70g, line 325:
```suggestion
  """The transaction result when it was successful. Formatted as raw JSON"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -261,25 +337,36 @@ type Transaction implements Node {
   rollback: Boolean
   senderAccount: String
   signers: [Signer!]
+
+  """The step number in the case that transactions are chained."""
```
alber70g, line 341:
```suggestion
  """
  The step-number when this is a execution of a `defpact`, aka multi-step transaction.
  """
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -261,25 +337,36 @@ type Transaction implements Node {
...
 }
 
+"""A transfer of funds from a fungible between two accounts."""
 type Transfer implements Node {
   amount: Decimal!
   blockHash: String!
   blocks: [Block!]!
   chainId: BigInt!
+
+  """The transfer that is the counterparty of this transfer."""
```
alber70g, line 355:
```suggestion
  """The counterpart of the crosschain-transfer. `null` when it's not a cross-chain-transfer"""
```

packages/apps/graph/generated-schema.graphql
```diff
@@ -261,25 +337,36 @@ type Transaction implements Node {
...
   crossChainTransfer: Transfer
   height: BigInt!
   id: ID!
   moduleHash: String!
   moduleName: String!
+
+  """
+  The order of the transfer in the case that there are chained Transfers.
+  """
```
alber70g, line 364:
```suggestion
  """
  The order of the transfer when it's a `defpact` execution.
  """
```

alber70g (commented):
All comments are suggestion/other perspectives. Food for thought and discussion. Love the bikeshedding :D

packages/apps/graph/generated-schema.graphql
MRVDH: Something to review when we implement orphans I think

alber70g:
> Two things I'd like to ask of the reviewers:
> 
> 1. Please check if all of the definitions of the queries, objects and the fields are correct.
> 2. Please help me find one-line explanations/descriptions of the following fields:
> 
> Block: -- epoch | What is the different with this and the creationTime field? Some epochs in the db are from 2019, and others from 2023.

block.epoch: The moment the difficulty is adjusted to maintain a block validation time of 30 seconds.

> MinerKey | Current description is "The public key of the miner that solved a block.", but I don't think that is correct, as a block can have multiple miner keys.

I think we should move the `minerkey` off of the `block` and put it to `blocks.miner`.

```
type Miner {
  keys: String[]
  predicate: String
  account: String
}
```

> 
> Signer | What is this specifically? Including the fields below.

In prisma.schema we have no relation to this table. But there is as there's a signer for each transaction. 

> -- address | Always empty, what is it? 

A transaction has a signer and a sender. A sender is an account, and the signer for the gas cap is the signer. A signer itself doesn't need an account. It just needs to be there as a sender. 

Also a signer signs for specific caps. Only the signer that signs for `coin.GAS` is associated with the sender_account_

> -- signature | Do we have an explanatory statement of this other than "the signature that was used to sign"?

The result of the signing operation of the hash of the transaction.

> 
> Transaction: 

> -- data | It seems the data can be pretty much anything. What is it specifically? 

Environment data made available to the transaction

> -- logs | This is always a value such as 'fBKTi_8_5T660fnFYvPTW9Wnbjj3276t95H95T9X55Y', but not sure what it means. 

Not sure if this is true though:

Identifier to retrieve the logs for the execution of the transaction

> -- metadata | Always empty, what is it? 

No clue

> -- pactId | A value such as 'MLfYVl8o6D70PzX1H1p1YRwnNGYNp2VHZPyL6kbKctA', what is it? 



> -- proof | A base64 encoded string with an object, input and subject. How should we describe this? 



> -- rollback | Not sure what this is.

`pactId` `proof` and `rollback` have to do with `defpacts` (multistep transactions)

`proof` is usually done when a crosschain tx is done, as the counterpart needs the proof to be able to execute

alber70g:
> > > -- pactId | A value such as 'MLfYVl8o6D70PzX1H1p1YRwnNGYNp2VHZPyL6kbKctA', what is it?
> > 
> > 
> > The value of `pactId` is only filled when it is a cross-chain transaction. The `pactId` is the `requestKey` of the starting transaction. So in cross-chain transactions we always have a starting transaction (on the origin chain) and a finishing transaction (on the target chain). The `pactId` will be filled on the finishing transaction and it will have the value of the starting transaction's `requestKey`
> 
> Wouldn't it be more logical to call the field `crossChainTransactionRequestKey`? cc: @alber70g Or is there a specific reason it's called "pactId"?

Yes, it's for any execution of a `defpact`. So not just cross-chain transactions, also multi-step transactions

MRVDH:
- Added descriptions everywhere
- Renamed module to fungible
- Removed miner keys and replaced it with miner account
- Added a max take on unlimited prisma calls

nil-amrutlal-dept (approved):

PR #1291, opened 2023-12-01
chore(@kadena/react-ui): Add utility classes for the new tokens called atoms

eileenmguo:
Created `atoms` which are css utility classes intended to replace `sprinkles` and use the new design tokens. 

Since the utility classes are used a lot, we will keep both `atoms` and `sprinkles` and update the components to use `atoms` incrementally. Eventually we want to remove `sprinkles` entirely, however we will share a plan for this in the near future.

NOTE: This means that `sprinkles` are now deprecated


packages/libs/react-ui/src/styles/atoms.css.ts
```diff
@@ -0,0 +1,130 @@
...
+      icon: tokens.kda.foundation.color.icon,
+      text: tokens.kda.foundation.color.text,
+    }),
+    backgroundColor: {
+      transparent: 'transparent',
+      ...flattenTokens(tokens.kda.foundation.color.background),
+    },
+    borderColor: flattenTokens(tokens.kda.foundation.color.border),
+  },
+});
```
eileenmguo, line 75: @isaozler and @salamaashoush should we move these to the `systemProperties` since they should already automatically switch depending on the theme?

packages/libs/react-ui/src/styles/tokens/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+const ignoredTokens = ['@hover', '@focus', '@disabled'];
+
+function isValue(token: Token): token is string {
+  return typeof token === 'string';
+}
+
+type Leaves<T> = T extends object
+  ? {
+      [K in keyof T]: `${Exclude<K, symbol | IgnoredToken>}${T[K] extends object
+        ? `.${Leaves<T[K]>}`
```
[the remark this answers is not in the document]
eileenmguo: since we only use sprinkles like the following, I think this naming convention is more readable 
```
atoms({ color: "text.default" })
```

packages/apps/docs/src/components/Layout/components/Menu/menu.css.ts
```diff
@@ -55,9 +55,11 @@ export const menuInLayoutVariants = styleVariants({
   ],
   false: [
     sprinkles({
-      display: 'block',
+      display: {
```
sstraatemans, line 58: 🔥

packages/libs/react-ui/src/styles/themeUtils.ts
```diff
@@ -1,15 +1,16 @@
...
 
 // eslint-disable-next-line @kadena-dev/typedef-var
 export const breakpoints = {
   xs: '',
-  sm: `screen and (min-width: ${640 / 16}rem)`,
-  md: `screen and (min-width: ${768 / 16}rem)`,
-  lg: `screen and (min-width: ${1024 / 16}rem)`,
-  xl: `screen and (min-width: ${1280 / 16}rem)`,
-  xxl: `screen and (min-width: ${1536 / 16}rem)`,
+  sm: 'screen and (min-width: 40rem)',
```
sstraatemans, line 9: 👍🏻

sstraatemans (approved):

PR #1297, opened 2023-12-04
feat(client): extract capability type from string codes

javadkh2:
Using Pact.module is the recommended way for creating pact code, however using just a string is valid as well and in some scenarios is more convenient.

execution already accepts pure strings but with this update execution function can also extract capability types from the pure string codes - of course you still need to install the contract type.

```TS
const transaction = Pact.builder
    .execution(
       `(coin.transfer ${sender} ${receiver} ${amount})`
    )
    .addSigner(keyFromAccount(sender), (withCapability) => [
      withCapability('coin.GAS'),
      // this type infers from the string code
      withCapability('coin.TRANSFER', sender, receiver, amount),
    ])
    .setMeta({ chainId: '0', senderAccount: sender })
    .setNetworkId(NETWORK_ID)
    .createTransaction();
```   

this can be used with Pact.modules
```TS
const transaction = Pact.builder
    .execution(
       `(coin.transfer ${sender} ${receiver} ${amount})`,
       Pact.modules["free.my-module"]["my-function"](...),
    )
```

### DIFF
--- a/packages/libs/client/src/interfaces/type-utilities.ts
+++ b/packages/libs/client/src/interfaces/type-utilities.ts
@@ -1,4 +1,5 @@
 import type { ICap } from '@kadena/types';
+import type { IPactModules } from '../pact';
 
 /**
  * @internal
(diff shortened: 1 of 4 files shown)


packages/libs/client/src/interfaces/type-utilities.ts
```diff
@@ -39,3 +40,32 @@ export type ExtractCapabilityType<TCommand> = TCommand extends {
...
+  : never;
+
+type RootModule<T> = T extends `(${infer moduleType}.${infer func}${FnRest})`
+  ? func extends `${string} ` | `${string} ${string}`
+    ? never
+    : GetFuncReturnType<moduleType, func>
+  : never;
+
+type ModuleWithNamespace<T> =
+  T extends `(${infer namespaceType}.${infer moduleType}.${infer func}${FnRest})`
```
barthuijgen, line 61: In some cases it will be `(${namespace}.${moduleType}.${func} ...)`

alber70g (approved):

ash-vd (approved):

PR #1298, opened 2023-12-05
feat(client-utils): add createPrincipal

jessevanmuijden:
refactored the dirty read call to Pact built-in function `(create-principal)` to a function in the `@kadena/client-utils` library, because the required code was already duplicated in two places and it is needed in a third place (`kadena cli`).


.changeset/nice-tools-destroy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@kadena/client-utils': patch
```
sanderlooijenga, line 2: Should perhaps be a `minor` change, since we're adding a new feature?

packages/libs/client-utils/src/built-in/create-principal.ts
```diff
@@ -0,0 +1,27 @@
+import { addData, execution } from '@kadena/client/fp';
+import { pipe } from 'ramda';
+import { dirtyReadClient } from '../core/client-helpers';
+import type { IClientConfig } from '../core/utils/helpers';
+
+interface ICreatePrincipalInput {
```
sanderlooijenga, line 6: Can we also `export` this interface?

packages/libs/client-utils/src/built-in/create-principal.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @alpha
+ */
+export const createPrincipal = async (
+  inputs: ICreatePrincipalInput,
+  config: Omit<IClientConfig, 'sign'>,
+): Promise<string> => {
+  const command = pipe(
+    () => '(create-principal (read-keyset "ks"))',
+    execution,
+    addData('ks', inputs.keyset),
```
sanderlooijenga, line 23: Couldn't we also use `addKeyset` here?

packages/libs/client-utils/src/integration-tests/built-in.int.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { createPrincipal } from '../built-in/create-principal';
+
+const config = {
+  host: 'http://127.0.0.1:8080',
+  defaults: {
+    networkId: 'fast-development',
+    meta: { chainId: '0' as ChainId },
+  },
+};
```
sanderlooijenga, line 12: Lil fyi; If you convert this to;

```
const config = {
…
} as const;
```

Then you don't need to cast `chainId` 😁

packages/apps/tools/src/services/faucet/create-principal.ts
```diff
@@ -18,31 +18,19 @@ export const createPrincipal = async (
...
-  if (response.result.status === 'success') {
-    return response.result.data as string;
-  }
-
-  throw new Error((response.result.error as any)?.message || 'Unknown error');
+  return createPrincipalUtil(
+    {
+      keyset: {
+        keys,
+        pred: pred as 'keys-all',
```
sanderlooijenga, line 25: Maybe we can do `pred as ICreatePrincipalInput['keyset']['pred']` (once it's exported)? Or better yet, give that as type to the parameter?

packages/apps/graph/src/devnet/helper.ts
```diff
@@ -111,11 +111,20 @@ export const generateAccount = async (
...
+        },
+      },
+      {
+        host: `http://localhost:${devnetConfig.PORT}`,
+        defaults: {
+          networkId: devnetConfig.NETWORK_ID,
+          meta: { chainId },
+        },
+      },
+    );
```
alber70g, line 127: Can we make it explicit instead of relying on overwriting an existing value?

E.g.: 

```
  let account: string | undefined = undefined;
  
  if (keyPairs.length === 1) {
    account = `k:${keyPairs[0].publicKey}`;
  } else if (keyPairs.length > 1) {
    account = await createPrincipal(
      {
        keyset: {
          keys: keyPairs.map((keyPair) => keyPair.publicKey),
        },
      },
      {
        host: `http://localhost:${devnetConfig.PORT}`,
        defaults: {
          networkId: devnetConfig.NETWORK_ID,
          meta: { chainId },
        },
      },
    );
```

alber70g (approved):
Not my strongest part. Maybe @EnoF can also take a look

packages/libs/client-utils/src/built-in/create-principal.ts
```diff
@@ -0,0 +1,27 @@
...
+  keyset: {
+    keys: string[];
+    pred?: 'keys-all' | 'keys-2' | 'keys-any';
+  };
+}
+
+/**
+ * @alpha
+ */
+export const createPrincipal = async (
```
EnoF, line 16: It would be nice if more than just  principals based on keysets could be created. Mainly `capability-guards` should probably be supported.

```
(create-principal (create-capability-guard (some.module.CAPABILITY_NAME)))
```

jessevanmuijden: as discussed, saving this for another PR.

packages/libs/client-utils/src/built-in/create-principal.ts
```diff
@@ -0,0 +1,30 @@
...
+  inputs: ICreatePrincipalInput,
+  config: Omit<IClientConfig, 'sign'>,
+): Promise<string> => {
+  const command = pipe(
+    () => '(create-principal (read-keyset "ks"))',
+    execution,
+    addKeyset('ks', inputs.keyset.pred || 'keys-all', ...inputs.keyset.keys),
+    dirtyReadClient(config),
+  );
+  return command().execute() as unknown as string;
```
javadkh2, line 29: shouldn't be `Promise<string>`

PR #1303, opened 2023-12-06
feat: adds support for a custom Chainweaver url for signWithChainweaver

### LINKED ISSUE #1232: Allow Chainweaver's IP address to be updated
### Please confirm the following:

- [X] I have read the [Code of Conduct](https://github.com/kadena-community/kadena.js/blob/main/CODE_OF_CONDUCT.md)
- [X] The feature was not suggested before

### Is your feature request related to a problem?

Currently, the IP address configuration for Chainweaver is hardcoded to 127.0.0.1, which is expected since it's designed to run locally, either as a desktop app or as a VM on Windows. Specifically, when discussing Windows, a challenge arises with the use of WSL/WSL2 instances, wherein the code runs within these environments. By default, Chainweaver's instance hosted in Virtualbox cannot be accessed at 127.0.0.1 easily. However, this can be overcome by using the IP address of the Virtualbox ethernet adapter.

### Describe the solution you'd like

Allow the hardcoded 127.0.0.1 IP address to be updated either globally or within the `signWithChainweaver` method call.

### Describe alternatives you've considered

_No response_

### Do you have any additional context?

_No response_

### Participation

- [ ] I am willing to submit a pull request for this feature request.

ash-vd:
Adds support for adding a custom url for `signWithChainweaver`. 

Closes #1232

### DIFF
--- a/packages/libs/client/src/signing/chainweaver/signWithChainweaver.ts
+++ b/packages/libs/client/src/signing/chainweaver/signWithChainweaver.ts
@@ -7,105 +7,89 @@ import type {
   IQuicksignResponse,
   IQuicksignSigner,
 } from '../../signing-api/v1/quicksign';
-import type { ISingleSignFunction } from '../ISignFunction';
+import type { ISignFunction } from '../ISignFunction';
 import { addSignatures } from '../utils/addSignatures';
 import { parseTransactionCommand } from '../utils/parseTransactionCommand';
 
 const debug: Debugger = _debug('pactjs:signWithChainweaver');
 
 /**
- * Interface to use when writing a signing function for Chainweaver that accepts a single transaction
- * @public
- */
-export interface IChainweaverSingleSignFunction extends ISingleSignFunction {
-  (
-    transaction: IUnsignedCommand,
-    chainweaverUrl?: string,
-  ): Promise<ICommand | IUnsignedCommand>;
-}
-
-/**
- * Interface to use when writing a signing function for Chainweaver that accepts multiple transactions
- * @public
- */
-export interface IChainweaverSignFunction
-  extends IChainweaverSingleSignFunction {
-  (
-    transactionList: IUnsignedCommand[],
-  ): Promise<(ICommand | IUnsignedCommand)[]>;
-}
-
-/**
- * Sign with chainweaver according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
+ * Creates the signWithChainweaver function with interface {@link ISignFunction}
+ * Lets you sign with chainweaver according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
  *
  * @public
  */
-export const signWithChainweaver: IChainweaverSignFunction = (async (
-  transactionList: IUnsignedCommand | Array<IUnsignedCommand | ICommand>,
+export function createSignWithChainweaver(
   chainweaverUrl = 'http://127.0.0.1:9467',
-) => {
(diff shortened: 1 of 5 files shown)


packages/libs/client/src/signing/chainweaver/signWithChainweaver.ts
```diff
@@ -7,19 +7,42 @@ import type {
...
+
 /**
  * Sign with chainweaver according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
  *
  * @public
  */
-export const signWithChainweaver: ISignFunction = (async (
+export const signWithChainweaver: IChainweaverSignFunction = (async (
   transactionList: IUnsignedCommand | Array<IUnsignedCommand | ICommand>,
+  chainweaverUrl = 'http://127.0.0.1:9467',
```
alber70g, line 45: This should be done with a wrapper function that creates the `signWithChainweaver` function. The final `signWithX` function should always comply with the `ISignFunction`.

My suggestion is to change this and add the `createSignWithChainweaver` function that'll provide the `chainweaverUrl`.

alber70g (changes requested):

packages/libs/client/src/signing/chainweaver/signWithChainweaver.ts
ash-vd: Fixed! This is however now a breaking change 😞

alber70g: We could still expose `signWithChainweaver` and `createSignWithChainweaver` where they both call a non-exposed function that has `chainweaverUrl` as argument where the `signWithChainweaver` uses the native one.

To prevent this from being an issue in the future, we could have a pattern that ALWAYS has the `createSignWithX`-function. Ie: `createSignWithX(options) -> ISignFunction`.

ash-vd: `signWithChainweaver` was the only function that didn't follow the `createSignWithX` pattern yet. I'm in favour of the option to always follow this pattern for the sign functions. It's also way more flexible as options can be added as needed without API changes.

packages/libs/client/src/signing/chainweaver/signWithChainweaver.ts
```diff
@@ -14,75 +14,101 @@ import { parseTransactionCommand } from '../utils/parseTransactionCommand';
...
+
+/**
+ * Creates the signWithChainweaver function with interface {@link ISignFunction}
+ * Lets you sign with chainweaver according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
+ *
+ * @public
+ */
+export function createSignWithChainweaver(
+  options = { chainweaverUrl: 'http://127.0.0.1:9467' },
+): ISignFunction {
```
alber70g, line 106: Minor comment, maybe use just "host" and add an example to the @params

```suggestion
/**
 * Creates the signWithChainweaver function with interface {@link ISignFunction}
 * Lets you sign with Chainweaver according to {@link https://github.com/kadena-io/KIPs/blob/master/kip-0015.md | sign-v1 API}
 *
 * @param options - object to customize behaviour.
 *   - `host: string` - the host of the chainweaver instance to use. Defaults to `http://127.0.0.1:9467`
 * @returns - {@link ISignFunction}
 */
export function createSignWithChainweaver(
  options = { host: 'http://127.0.0.1:9467' },
): ISignFunction {
```

alber70g (approved):

PR #1304, opened 2023-12-06
Also allow WEBAUTHN- public keys

sanderlooijenga:
---
  - https://app.asana.com/0/0/1206091721942276

### DIFF
--- a/packages/apps/tools/src/services/utils/utils.ts
+++ b/packages/apps/tools/src/services/utils/utils.ts
@@ -17,7 +17,7 @@ export const validateRequestKey = (key: string): string | undefined => {
 };
 
 export const validatePublicKey = (key: string): string | undefined => {
-  if (key.length === 64 || /(^WEBAUTHN-[a-f0-9]+)$/.test(key)) {
+  if (key.length === 64 || /(^WEBAUTHN-[a-fA-F0-9]+)$/i.test(key)) {
     return key;
   }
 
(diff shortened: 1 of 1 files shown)


packages/apps/tools/src/services/utils/utils.ts
```diff
@@ -17,7 +17,7 @@ export const validateRequestKey = (key: string): string | undefined => {
 };
 
 export const validatePublicKey = (key: string): string | undefined => {
-  if (key.length === 64) {
+  if (key.length === 64 || /(^WEBAUTHN-[a-f0-9]+)$/.test(key)) {
```
nillo, line 20: Saw this by coincident (just passing through:P) but shouldn't this be:

```
export const validatePublicKey = (key: string): string | undefined => {
  if (key.length === 64 || /(^WEBAUTHN-[a-fA-F0-9]+)$/i.test(key)) {
    return key;
  }
}
```
For instance, if the public key is encoded in base64url, as is common, then it can contain uppercase characters. However, if the public key is represented in hexadecimal format, it is generally case-insensitive, meaning that the uppercase and lowercase representations are equivalent. I don't know the full implementation details :) but if there's a possibility that the WebAuthn public keys include uppercase characters, you should modify the regex to account for this

sanderlooijenga: Appreciate the feedback 💪  will implement the uppercase as well!

KristinaSpasevska (approved):

PR #1311, opened 2023-12-07
[@tools-webiste] Hotfix: logo on mobile disappears

KristinaSpasevska:


ferreroltd (approved):

packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
```diff
@@ -30,6 +30,15 @@ const Header: FC<IHeaderProps> = () => {
 
   const currentTheme = theme === 'system' ? systemTheme : theme;
 
+  const [smallScreen, setSmallScreen] = useState(false);
+
+  useEffect(() => {
```
salamaashoush, line 35: I think it is better to use `window.matchMedia` API instead

here is how you could do it 

```ts
// put this somewhere in utils folder maybe
/**
 * Checks if the media query matches or not
 * it can accept media query builder like `breakpoint.up` etc
 */
export function useIsMatchingMediaQuery(queryInput: string): boolean {
  const query = queryInput.replace(/^@media( ?)/m, '');

  const [getSnapshot, subscribe] = useMemo(() => {
    const mediaQueryList = window.matchMedia(query);
    return [
      () => mediaQueryList.matches,
      (notify: () => void) => {
        mediaQueryList.addEventListener('change', notify);
        return () => {
          mediaQueryList.removeEventListener('change', notify);
        };
      },
    ];
  }, [query]);

  return useSyncExternalStore(subscribe, getSnapshot);
}
```

then use in the component like this
```ts
const isMobile = useIsMatchingMediaQuery('(width <= 768px)');
```
or use the theme breakpoints
```ts
const isMobile = useIsMatchingMediaQuery(
  `(width <= ${lightThemeValues.kda.foundation.breakpoint.sm})`,
);
```

KristinaSpasevska: Ah nice, thanks 🙂  I will make these changes in the other files where I'm testing whether the screen is smaller.

sanderlooijenga: Yeah, a reusable hook is definitely a good improvement. Good step towards supporting it properly. We could also perhaps check the user agent in `getServersideProps` and feed the `device` to the page 🤔  that way perhaps layout shifting doesn't happen?

salamaashoush: yes the idea is to make sure we are implementing responsive design same way in CSS and JS by using media queries for both, the hook I provided is much more efficient and yes could avoid layout shifts since it is `useSyncExternalStore` to avoid unnecessary state changes compared to `useEffect` + `useState` approach

salamaashoush: @KristinaSpasevska I edited the code to be compatible with nextjs

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -9,12 +9,17 @@ export const containerClass = style([
     display: 'flex',
     flexDirection: 'row',
     flexWrap: 'nowrap',
+    width: 'max-content',
```
eileenmguo, line 12: I believe you can do this 
```suggestion
    width: {
      xs: "max-content",
      sm: "100%",
    }
```

eileenmguo (approved):

salamaashoush (approved):

packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
```diff
@@ -30,6 +32,8 @@ const Header: FC<IHeaderProps> = () => {
 
   const currentTheme = theme === 'system' ? systemTheme : theme;
 
+  const isMediumScreen = useIsMatchingMediaQuery(`${breakpoints.sm}`);
```
salamaashoush, line 35: This should be a media query syntax with the media par, so probably you meant something like this
```suggestion
  const isMediumScreen = useIsMatchingMediaQuery(`(width <= ${breakpoints.sm})`);
```
or ```suggestion
  const isMediumScreen = useIsMatchingMediaQuery(`(min-width: ${breakpoints.sm})`);
```

sanderlooijenga (approved):

PR #1312, opened 2023-12-07
[@kadena/graph] Made graph a releasable package

MRVDH:
- Added a bin reference to dist, making it possible to for example run `npx kadena-graph` to start the graphql server.
- Changed the package.json to a releasable format.
- Fixed module aliases for the built version.

### DIFF
--- a/packages/apps/graph/package.json
+++ b/packages/apps/graph/package.json
@@ -1,14 +1,24 @@
 {
   "name": "@kadena/graph",
   "version": "1.0.6",
-  "private": true,
-  "description": "",
-  "keywords": [],
-  "license": "ISC",
-  "author": "",
-  "main": "index.js",
+  "description": "GraphQL server, available for running your own GraphQL endpoint. This project uses chainweb-data as the datasource.",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/kadena-community/kadena.js.git",
+    "directory": "packages/apps/graph"
+  },
+  "license": "BSD-3",
+  "contributors": [
+    "Albert Groothedde <albert@kadena.io>",
+    "Maarten van den Hoven <maarten.vandenhoven@deptagency.com>",
+    "Nil Amrutlal <nil.amrutlal@deptagency.com>"
+  ],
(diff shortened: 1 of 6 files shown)


packages/apps/graph/package.json
```diff
@@ -1,14 +1,24 @@
...
+    "Albert Groothedde <albert@kadena.io>",
+    "Maarten van den Hoven <maarten.vandenhoven@deptagency.com>",
+    "Nil Amrutlal <nil.amrutlal@deptagency.com>"
+  ],
+  "bin": {
+    "kadena-graph": "./dist/index.js"
+  },
   "scripts": {
-    "build": "pnpm run pactjs:generate:contract:coin && pnpm prisma:generate",
+    "prebuild": "pnpm run pactjs:generate:contract:coin && pnpm prisma:generate",
```
nil-amrutlal-dept, line 20: Confirm this change with @alber70g since I believe it was like this but changed for some reason I cant recall rn

packages/apps/graph/src/devnet/crosschain-transfer.ts
```diff
@@ -130,7 +130,7 @@ export async function crossChainTransfer({
   );
 
   const continuation = {
-    pactId: status.continuation?.pactId,
+    pactId: status.continuation?.pactId || '',
```
nil-amrutlal-dept, line 133: As we talked yesterday maybe we should discuss if we should change our functions to use `client-utils`

nil-amrutlal-dept (approved):

packages/apps/graph/package.json
alber70g: I think turbo doesn't execute `prebuilt`s. Maybe it was a problem with Rush, and isn't an issue anymore. Just check in CI whether it executes

packages/apps/graph/src/devnet/crosschain-transfer.ts
alber70g: What do we check here? Is `0` an option here? That will evaluate to `''`

PR #1313, opened 2023-12-07
WIP | Keys support for kadena CLI (feat/kadena-cli/keyz)

nillo:
**This PR adds key generation support to kadena CLI using Hd-wallet**
- Generate plain keys from random buffer -- legacy also supported
- Generate HD keys  (new derivation path supported / legacy )
- Generate keys from encrypted seed (new / legacy ) ( will automatically detect new or legacy based on seed 
-Generate keys from mnemonic phrase (>seed>keys: new / legacy) --legacy also supported
- Delete plain keys:  new / legacy keys  | delete all keys | seed can only be deleted by manually selected from list (won't be deleted with delete all)
- Updated encrypted seed password
-  Decrypt encrypted message (e.g seed/privatekey)

**To-do:**

- Change password prompt to accept empty password to make the following possible:
Default > encrypted 
No password >  Not encrypted (plain keys) 

- remove .enc prefix from filename

**How to test use:**
- _git clone project / git checkout branche (feat/kadena-cli/keyz)_

- _pnpm i_
- _pnpm build_
- _bin/kadena-cli.js --help_


packages/tools/kadena-cli/src/keys/commands/keysGenFromMnemonic.ts
```diff
@@ -0,0 +1,81 @@
...
+    globalOptions.keyAlias(),
+    globalOptions.keyAmount({ isOptional: true }),
+    globalOptions.legacy({ isOptional: true, disableQuestion: true }),
+  ],
+  async (config) => {
+    clearCLI();
+    try {
+      debug('generate-from-mnemonic:action')({ config });
+
+      const loading = ora('Generating..').start();
```
barthuijgen, line 39: how long does this process usually take? I feel like these operations are usually close to instant, do we need a spinner?

nillo: This part is using the kadena-crypto package, it’s a lot slower the the slip 10 version. It’s actually takes up to 10 sec for generating 30 tokens. ( since the internals of that package are running some low level commands which is kind of a black box) we have no influence on it

packages/tools/kadena-cli/src/keys/utils/keySharedKeyGen.ts
```diff
@@ -0,0 +1,94 @@
...
+import { kadenaGenKeypair } from '@kadena/hd-wallet/chainweaver';
+import { fromHexStr, toHexStr } from './keysHelpers.js';
+import type { IKeyPair } from './storage.js';
+
+const defaultAmount: number = 1;
+
+export interface IKeysConfig {
+  keyGenFromChoice: string;
+  keyAlias: string;
+  keySeed?: string;
```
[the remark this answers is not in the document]
nillo: Correct

packages/tools/kadena-cli/src/keys/utils/storage.ts
```diff
@@ -1,75 +1,76 @@
...
+} from '../../utils/filesystem.js';
 import { sanitizeFilename } from '../../utils/helpers.js';
 
-interface IKeyPair {
+export type TSeedContent = string;
+
+export interface IKeyPair {
   publicKey: string;
-  privateKey: string;
+  privateKey?: string;
```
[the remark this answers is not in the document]
nillo: Correct

packages/tools/kadena-cli/src/keys/commands/keysPlainGenerate.ts
```diff
@@ -1,62 +1,108 @@
...
+  } else {
+    const randomKeyPairs = await kadenaKeyPairsFromRandom(amount);
+    return randomKeyPairs.map((keyPair) => {
+      const encryptedPrivateKey = kadenaEncrypt(
+        config.keyPassword,
+        fromHexStr(keyPair.secretKey),
+      );
+      return {
+        publicKey: keyPair.publicKey,
+        privateKey: encryptedPrivateKey,
```
alber70g, line 41: kda-tool uses `public` and `secret` should we do the same?

packages/tools/kadena-cli/src/keys/commands/keysManage.ts
```diff
@@ -1,25 +1,83 @@
...
+        keySeed,
+      );
+
+      if (isLegacy === true) {
+        const newSeed = await kadenaChangePassword(
+          decryptedCurrentSeed,
+          config.securityCurrentPassword,
+          config.securityNewPassword,
+        );
+        encryptedNewSeed = kadenaEncrypt(config.securityNewPassword, newSeed);
```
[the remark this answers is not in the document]
nillo: albert suggested to move this to the HD-wallet.

having that said, if other projects use this. They don't expect this double encryption. how to deal with this.
@javadkh2 @alber70g @barthuijgen

packages/tools/kadena-cli/src/keys/commands/keysPlainGenerate.ts
nillo: i have taken this from hd-wallet, 
The question to support this like is done with kda-tool ( which is not used that much, i would prefer privateKey
@buckie  @emilypi  @sirlensalot  what do you think?

packages/tools/kadena-cli/src/keys/commands/keysGenFromMnemonic.ts
```diff
@@ -0,0 +1,81 @@
...
+  [
+    globalOptions.keyGenFromChoice(),
+    globalOptions.keyMnemonic(),
+    globalOptions.keyPassword(),
+    globalOptions.keyAlias(),
+    globalOptions.keyAmount({ isOptional: true }),
+    globalOptions.legacy({ isOptional: true, disableQuestion: true }),
+  ],
+  async (config) => {
+    clearCLI();
```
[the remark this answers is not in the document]
nillo: i can live with that

packages/tools/kadena-cli/src/keys/commands/keysPlainGenerate.ts
sirlensalot: generally cryptography libraries use secret, e.g. https://github.com/jedisct1/libsodium

packages/tools/kadena-cli/src/keys/commands/keysManage.ts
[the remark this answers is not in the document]
barthuijgen: I think moving it to HD-Wallet is a good idea, that's where the crypto methods live anyway, it can be in control of that.

packages/tools/kadena-cli/src/keys/commands/keysPlainGenerate.ts
LindaOrtega: I would prefer `privateKey`. But if we go with `secretKey` we should annotate the argument to explain what secret key is referring to.

For context, I was working with some developers that were using our cryptography-utils library and they were confused on the difference between `private key` and `secret key`. They thought secret key might have been some kind of password/seed.

ash-vd: I do not have hard evidence on this, but I feel like I've seen and used `secretKey` more than `privateKey` and so I would prefer `secretKey`. `tweetnacl` also uses `secretKey`: https://www.npmjs.com/package/tweetnacl#naclboxkeypair

I know @lsgunnlsgunn has been thinking about a glossary and how we name things. We should use the same wording everywhere. Lisa, what's your opinion on this?

sirlensalot: @LindaOrtega did the devs you worked with have any experience with cryptography? I wonder what eth web3.js calls it.

LindaOrtega: @sirlensalot yes, they had some experience with cryptography and a lot of experience with Ethereum.

========

> I wonder what eth web3.js calls it.

I'm seeing `privatekey` used with Eth (see https://web3js.readthedocs.io/en/v1.10.0/web3-shh.html?highlight=keypair#addprivatekey)

This Node.js crypto library uses `privateKey` (see https://nodejs.org/api/crypto.html#signsignprivatekey-outputencoding).

But Solana uses `secretKey` (see https://solana-labs.github.io/solana-web3.js/classes/Keypair.html#fromSecretKey).

========
What is the difference between `secret key` and `private key`?
I've seen `secret key` and `private key` used interchangeably in the past, so I always assumed they were the same thing. I found [this](https://security.stackexchange.com/questions/90121/secret-public-and-private-keys) Stackoverflow answer on the difference between the two, but I'm unsure of its validity.

I don't have a preference between the two, but I +1 on having similar naming conventions across the org. 
I'm seeing `privatekey`, `privkey`, and `secretkey` used across kadena.js, pact, and chainweb-node repos (see [here](https://github.com/kadena-io/pact/blob/e5c922664fb4df53e3b8e2c48618d71b3ef612ad/src/Pact/Types/Crypto.hs#L171) for an example). 

Am I missing something? i.e. are we always using `secretkey` for external facing APIs?

sirlensalot: OK, I found some stuff, it looks like there is a lot of literature that uses secret for symmetric, and private for asymmetric, such as this internet glossary: https://datatracker.ietf.org/doc/html/rfc4949

Meanwhile all of the ED25519 djb literature exclusively uses "secret", the original paper literally does not have "private" in it.

I can only guess but I think the DJBs of this world are writing crypto libraries and want to audit secrets, whether asymmetric or no, for things like wiped memory etc. Meanwhile OpenSSL uses private key.

That (and Haskell crypto libs using secret) explains Pact, which was built around ED25519.

But it looks like the dispute is legitimate. Pact does have an "API" which is the yaml format and the output of `pact -g`, but otherwise private/secret keys never show up in an API for obvious reasons. 

I also note that this very codebase references existing Kadena code that uses "secret" (https://github.com/kadena-community/kadena.js/blob/a0c54b4bc9a20ff50c2fc589ec6ea3046cab27a8/packages/libs/hd-wallet/src/SLIP10/kadenaKeyPairsFromRandom.ts#L20) which itself uses node.js `crypto` which uses `privateKey` LOL

So both of the following are true:
1. Kadena more or less standardized on `secret`, from its direct influences (haskell, nacl, DJB); maybe the latter is Solana's influence too?
2. Non-DJB literature on asymmetric, plus most crypto projects (Bitcoin, Ethereum), use "private"

So I can now see either side and withdraw from this discussion :) However if you do end up using private, check in before you break our yaml format because we have numerous internal procedures that use that, which uses `secret`

nillo: thanks for all the input @LindaOrtega @sirlensalot @alber70g @barthuijgen 
The team decided to go for secretKey to keep it consistent.

PR #1314, opened 2023-12-07
added react aria text field hook

ferreroltd:
This pull request can be solved in 2 ways: either applying it at each field level, as illustrated in this PR example, or incorporating it into the FormFieldWrapper.

Opting for the per-field application offers a straightforward API, requiring the simple passage of props. Additionally, it simplifies the process of integrating the useNumberField hook *dince it would be same component different hook), where its inclusion is as straightforward as adding it during the creation of the number field components for export.

On the other hand, selecting the FormFieldWrapper approach needs uding React.cloneElement. While this method offers a centralized approach, it introduces more instances where we decided that we wanted to stop using it.

Let me know!

### DIFF
--- a/packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
+++ b/packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
@@ -2,6 +2,7 @@ import type { IFormFieldWrapperProps, IInputProps } from '@components/Form';
 import { FormFieldWrapper, Input } from '@components/Form';
 import type { FC } from 'react';
 import React, { useRef } from 'react';
+import type { AriaTextFieldProps } from 'react-aria';
 import { useTextField } from 'react-aria';
 
 export interface ITextFieldProps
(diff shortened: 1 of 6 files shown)


packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -12,18 +13,26 @@ export const TextField: FC<ITextFieldProps> = ({
   disabled = false,
   inputProps,
   status,
-  ...rest
+  ...props
 }) => {
   const { id } = inputProps;
+  const ref = useRef(null);
+  const { labelProps, inputProps: ariaInputProps } = useTextField(props, ref);
```
salamaashoush, line 20: there are description and error props as well, so we need to use them to have a completely accessible formfiled

salamaashoush:
> Maybe combine the "texts" as in area and input and pass a prop to identify the type. Think about the number of Field

I don't think it is a good idea for a couple of reasons
1. polymorphic components make maintenance difficult and more fragile to breaking changes.
2. textarea has different visuals and behavior like `cols`, `rows`, and auto-resize, etc.

packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -12,18 +13,35 @@ export const TextField: FC<ITextFieldProps> = ({
...
   const { id } = inputProps;
+  const ref = useRef(null);
+  const {
+    labelProps,
+    inputProps: ariaInputProps,
+    descriptionProps,
+    errorMessageProps,
+  } = useTextField(props, ref);
+
+  const computedHelperTextProps =
```
salamaashoush, line 27: what if we have both?

ferreroltd:
To be refactored

PR #1317, opened 2023-12-08
feat(types): export predicate type

jessevanmuijden:


MRVDH (approved):
Approved for the `graph` project

nil-amrutlal-dept (approved):

ash-vd (approved):

packages/apps/tools/src/services/faucet/create-principal.ts
```diff
@@ -15,7 +16,7 @@ const debug = Debug('kadena-transfer:services:faucet');
 export const createPrincipal = async (
   keys: string[],
   chainId: ChainwebChainId,
-  pred = 'keys-all',
+  pred: TPredicate,
```
sanderlooijenga, line 19: The default value should still be present?

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -45,6 +44,8 @@ import { createPrincipal } from '@/services/faucet/create-principal';
 import { fundCreateNewAccount } from '@/services/faucet/fund-create-new';
 import { validatePublicKey } from '@/services/utils/utils';
 import { zodResolver } from '@hookform/resolvers/zod';
+import type { TPredicate } from '@kadena/types';
+import { Predicate } from '@kadena/types';
```
sanderlooijenga, line 48: I think the type library should only contain Types? So we might need to place this somewhere else. @alber70g what do you think?

packages/apps/tools/src/services/faucet/fund-create-new.ts
```diff
@@ -23,7 +24,7 @@ export const fundCreateNewAccount = async (
   keys: string[],
   chainId: ChainwebChainId,
   amount = 100,
-  pred = 'keys-all',
+  pred: TPredicate,
```
sanderlooijenga, line 27: The default value should still be present?

packages/libs/types/src/Predicate.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @alpha
+ */
+export const Predicate = {
```
javadkh2, line 4: why do we need this object?

packages/libs/types/src/Predicate.ts
```diff
@@ -0,0 +1,16 @@
...
+  keysAll: 'keys-all',
+  keysTwo: 'keys-2',
+  keysAny: 'keys-any',
+} as const;
+
+/**
+ * @alpha
+ */
+export type TPredicate =
```
javadkh2, line 13: why string | undefined ?
I think its better to have simple type here
'keys-all' | 'keys-2' | 'keys-any'

packages/libs/client-utils/src/integration-tests/coin.int.test.ts
```diff
@@ -137,7 +138,7 @@ describe('createAccount', () => {
         account: accountTwo.account,
         keyset: {
           keys: [accountTwo.publicKey],
-          pred: 'keys-all',
+          pred: Predicate.keysAll,
```
javadkh2, line 141: I prefer to see "keys-all". it still have code completion

javadkh2 (changes requested):

packages/libs/types/src/Predicate.ts
javadkh2: this whole type means "string" | undefined.

javadkh2:
Adding `TPredicate` was a good idea. 
However I would just make it like this
```TS
TPredicate = 'keys-all' | 'keys-any' | 'keys-2'
```

packages/apps/tools/src/services/faucet/fund-create-new.ts
jessevanmuijden: I don't think so, because Pact will use `keys-all` by default if no predicate is passed.

packages/libs/types/src/Predicate.ts
jessevanmuijden: true, I think we should drop `string`. It was a suggestion from @alber70g , but maybe I understood it wrong.

packages/libs/client-utils/src/integration-tests/coin.int.test.ts
jessevanmuijden: why would you prefer a hardcoded string over a variable?

packages/libs/types/src/Predicate.ts
jessevanmuijden: we need this readonly object, because we do not like typescript enums, but theoretically the predicate is a string enum. The values of this enum are used all over the codebase, so it is cleaner to use the enum than the hardcoded string values.

javadkh2: I think `TPredicate` should be enough. 
then TS only allows those three valid options and your IDE suggest the types to you

packages/libs/client-utils/src/integration-tests/coin.int.test.ts
javadkh2: its not string it a literal (constant) type with only three valid options. 
it turns to string because of the error in the TPredicate
And if something is never going to change, then why should we put it inside a variable?

packages/apps/tools/src/services/faucet/fund-create-new.ts
javadkh2: I don't know about pact but its better to have the default value to avoid any unexpected behaviour later

packages/libs/types/src/Predicate.ts
jessevanmuijden: That was my initial implementation, but then this further refactor was suggested.

jessevanmuijden:
> Adding `TPredicate` was a good idea. However I would just make it like this
> 
> ```ts
> TPredicate = 'keys-all' | 'keys-any' | 'keys-2'
> ```

As I understood, `undefined` should also be supported.

jessevanmuijden:
After some discussion, we decided to go with the simplest solution that supports all use cases, which is `type TPredicate = string | undefined`, with the most common use-cases (keys-all, keys-2, keys-any) documented in the JSDoc. I have created a new PR for the final change here: https://github.com/kadena-community/kadena.js/pull/1355.

PR #1319, opened 2023-12-11
New aria complete button without new tokens

salamaashoush:
1. Use react aria which come with one major change is `onPress` instead of `onClick` read more about why https://react-spectrum.adobe.com/blog/building-a-button-part-1.html. we added `onClick` to allow easy migration but it is deprecated and we should not use it for new code.
2. Unify the `IconButton` and `Button` components using `<Button icon={<SomeIcon/>} />` instead of `IconButton`.
3. Change some props names to be consistent with react-aria naming `compact` -> `isCompact`.
4. `Button` is not a polymorphic component anymore we will have a separate link component.
5. `iconAlign` is replaced by specific icon renders props `startIcon` and `endIcon`, and `icon` is repurposed for icon-only buttons aka `IconButton`.
6. Use `recipe` instead of individual `styleVariants`. 
7. `color` and `variant` are now one prop `variant` and all `alternative` variants are added as a standalone variant eg `<Button color="primary" variant="alternative" />` is now `<Button variant="primaryInverted" />` we use `inverted` postfix is used instead of  `alternative` to match the intended color inversion behavior.
8. The `isCompact` variant now works for all button variations before it only worked for some color variants.

### DIFF
--- a/packages/libs/react-ui/src/components/Button/NewButton.css.ts
+++ b/packages/libs/react-ui/src/components/Button/NewButton.css.ts
@@ -1,0 +1,240 @@
+import { createVar, keyframes, style } from '@vanilla-extract/css';
+import { recipe } from '@vanilla-extract/recipes';
+import { colorPalette } from '../../styles/colors';
+import { tokens } from '../../styles/tokens/contract.css';
+import { bodyBaseBold } from '../../styles/tokens/styles.css';
+import { vars } from '../../styles/vars.css';
+
+const hoverBackgroundColor = createVar();
+const activeBackgroundColor = createVar();
+const hoverColor = createVar();
+const outlineColor = createVar();
+const backgroundColor = createVar();
+const color = createVar();
+
+// TODO: use the new tokens once they are ready in the design system and figma is updated
+
+// eslint-disable-next-line @kadena-dev/typedef-var
+const colorVariants = {
+  primary: {
+    vars: {
+      [color]: vars.colors.$primaryContrast,
+      [backgroundColor]: vars.colors.$primarySurface,
+      [hoverBackgroundColor]: vars.colors.$primaryHighContrast,
+      [activeBackgroundColor]: vars.colors.$primaryHighContrast,
+      [hoverColor]: vars.colors.$primaryContrast,
+      [outlineColor]: vars.colors.$primaryAccent,
+    },
+  },
+  secondary: {
+    vars: {
+      [color]: vars.colors.$secondaryContrast,
+      [backgroundColor]: vars.colors.$secondarySurface,
+      [hoverBackgroundColor]: vars.colors.$secondaryHighContrast,
+      [activeBackgroundColor]: vars.colors.$secondaryHighContrast,
+      [hoverColor]: vars.colors.$secondaryContrast,
+      [outlineColor]: vars.colors.$secondaryAccent,
+    },
+  },
+  tertiary: {
+    vars: {
+      [color]: vars.colors.$tertiaryContrast,
+      [backgroundColor]: vars.colors.$tertiarySurface,
+      [hoverBackgroundColor]: vars.colors.$tertiaryHighContrast,
+      [activeBackgroundColor]: vars.colors.$tertiaryHighContrast,
+      [hoverColor]: vars.colors.$tertiaryContrast,
+      [outlineColor]: vars.colors.$tertiaryAccent,
+    },
+  },
+  warning: {
+    vars: {
+      [color]: vars.colors.$warningContrast,
+      [backgroundColor]: vars.colors.$warningSurface,
+      [hoverBackgroundColor]: vars.colors.$warningHighContrast,
+      [activeBackgroundColor]: vars.colors.$warningHighContrast,
+      [hoverColor]: vars.colors.$warningContrast,
+      [outlineColor]: vars.colors.$warningAccent,
+    },
+  },
+  negative: {
+    vars: {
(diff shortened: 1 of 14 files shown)


packages/libs/react-ui/src/components/Button/NewButton.css.ts
```diff
@@ -0,0 +1,240 @@
...
+      position: 'relative',
+      display: 'flex',
+      placeItems: 'center',
+      cursor: 'pointer',
+      border: 'none',
+      textDecoration: 'none',
+      borderRadius: tokens.kda.foundation.radius.sm,
+      gap: tokens.kda.foundation.size.n2,
+      paddingInline: tokens.kda.foundation.size.n4,
+      paddingBlock: tokens.kda.foundation.size.n2,
```
eileenmguo, line 156: can we use atoms for these?

packages/libs/react-ui/src/components/Button/NewButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+  MouseEvent,
+  ReactNode,
+} from 'react';
+import React, { forwardRef, useCallback } from 'react';
+import type { AriaButtonProps } from 'react-aria';
+import { useButton } from 'react-aria';
+import { Loading } from '../Icon/System/SystemIcon';
+import { button, spinner } from './NewButton.css';
+
+type Variants = Omit<NonNullable<RecipeVariants<typeof button>>, 'onlyIcon'>;
```
eileenmguo, line 17: Does this make all recipe variants required props? If so, I don't think we should remove `NonNullable` and have the variants default to what is set in the recipe so users are not required to add every prop

packages/libs/react-ui/src/components/Button/NewButton.tsx
```diff
@@ -0,0 +1,83 @@
...
+export interface IButtonProps extends AriaButtonProps, Variants {
+  startIcon?: ReactNode;
+  endIcon?: ReactNode;
+  icon?: ReactNode;
+  title?: string;
```
eileenmguo, line 22: can we remove this? I believe people can use `aria-label` for this

eileenmguo (commented):
great documentation 🤩

packages/libs/react-ui/src/components/Button/NewButton.tsx
salamaashoush: no it doesn't, it is a workaround for vanilla extract types to make it work with `Omit`

packages/libs/react-ui/src/components/Button/NewButton.tsx
salamaashoush: this is a different thing, this is the native `title` prop which is used to add a native tooltip and it is useful for icon buttons

packages/libs/react-ui/src/components/Button/NewButton.css.ts
```diff
@@ -0,0 +1,279 @@
...
+});
+
+export const spinner = style({
+  animationName: rotate,
+  animationDuration: '1.5s',
+  animationIterationCount: 'infinite',
+  animationTimingFunction: 'linear',
+});
+
+export const cloak = style({
```
eileenmguo, line 272: can be done with atoms

packages/libs/react-ui/src/styles/atoms.css.ts
```diff
@@ -89,22 +90,36 @@ const responsiveProperties = defineProperties({
...
+    padding: [
+      'paddingBlockStart',
+      'paddingBlockEnd',
+      'paddingInlineStart',
+      'paddingInlineEnd',
+    ],
+    paddingX: ['paddingInlineStart', 'paddingInlineEnd'],
+    paddingY: ['paddingBlockStart', 'paddingBlockEnd'],
+    paddingInline: ['paddingInlineStart', 'paddingInlineEnd'],
+    paddingBlock: ['paddingBlockStart', 'paddingBlockEnd'],
```
eileenmguo, line 122: 😍

packages/libs/react-ui/src/components/Button/NewButton.css.ts
salamaashoush: This is not needed anymore

packages/libs/react-ui/src/styles/atoms.css.ts
```diff
@@ -3,6 +3,11 @@ import mapValues from 'lodash.mapvalues';
 import { breakpoints, flattenTokens } from './themeUtils';
 import { tokens } from './tokens/contract.css';
 
+export const colorAtoms = flattenTokens({
```
eileenmguo, line 6: should we add "currentColor" here?

packages/libs/react-ui/src/components/ProgressCircle/ProgressCircle.tsx
```diff
@@ -0,0 +1,87 @@
...
+      )}
+      width={spinnerSize}
+      height={spinnerSize}
+      viewBox={`0 0 ${spinnerSize} ${spinnerSize}`}
+      fill="none"
+      strokeWidth={strokeWidth}
+    >
+      <circle role="presentation" cx={center} cy={center} r={r} stroke="gray" />
+      <circle
+        {...testProps(props, 'progress')}
```
eileenmguo, line 62: what is the reason for adding this? I thought we would be testing based on role?

eileenmguo (commented):
Looks great!

PR #1329, opened 2023-12-12
fix(pactjs-cli): dependency resolution

javadkh2:
Fixed the issue of resolving dependencies when the namespace is implicitly the same as the module namespace.


nil-amrutlal-dept (approved):
Looks good 🙌🏽

PR #1339, opened 2023-12-14
[@kadena/graph] Replace custom client code with client utils functions

MRVDH:
Since this PR is quite large, I suggest looking at the individual commits, they are grouped per replacement type.

- [ ] [Changed kadena client functions to utils functions](https://github.com/kadena-community/kadena.js/pull/1339/commits/f46970cfb46caef24a7483cc5bbdfe8930cf5d7a)
- [ ] [typo in file name](https://github.com/kadena-community/kadena.js/pull/1339/commits/cf322723c1e159bcfaa439bcab775837b1dc68f4)
- [ ] [getBalance replaced with client-utils function](https://github.com/kadena-community/kadena.js/pull/1339/commits/f4791910d7d2bf5d19f686b4baf7076a88056863)
- [ ] [Replaced simulate config with dotenv and kadena client utils](https://github.com/kadena-community/kadena.js/pull/1339/commits/769923ee32e7858e12d2f7c8d1dbfdedb3dc50fb)
- [ ] [Pulled gas estimates out of devnet](https://github.com/kadena-community/kadena.js/pull/1339/commits/a0a0b5d2d4c17f00f4731b159e74378bb3a17244)
- [ ] [Replaced custom cross chain transfer with client utils version](https://github.com/kadena-community/kadena.js/pull/1339/commits/5f60fc252dd74644c319ef5c56456aa1235d9c63) & [Use executeTo for cross chain transfers](https://github.com/kadena-community/kadena.js/pull/1339/commits/373630edef22990e8fe65a57fbb3a755069440f9)
- [ ] [Replaced transfer with transfer create](https://github.com/kadena-community/kadena.js/pull/1339/commits/34e39aea07333148111febb14567c7a3273f3b0d)
- [ ] [Replaced safe transfer with partially built in client utils functions](https://github.com/kadena-community/kadena.js/pull/1339/commits/4faf382394065b75b2e9d75dcaa06a68832a1c3a)


packages/apps/graph/src/devnet/crosschain-transfer.ts
```diff
@@ -108,43 +35,36 @@ export async function crossChainTransfer({
...
-  inspect('Transfer Submited')(submittedTx2);
-  const status2 = await listen(submittedTx2);
-  inspect('Transfer Result')(status2);
-  return status2;
+  return await new Promise((resolve, reject) => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    crossChainTransferRequest.execute();
+    crossChainTransferRequest.on('listen', resolve);
+    crossChainTransferRequest.on('error', reject);
+  });
```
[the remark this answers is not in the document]
MRVDH: Didn't know that, thanks! 😄

packages/apps/graph/src/devnet/safe-transfer.ts
```diff
@@ -1,22 +1,15 @@
...
-  listen,
-  logger,
-  sender00,
-  signAndAssertTransaction,
-  stringifyProperty,
-  submit,
-} from './helper';
+import { logger, sender00, stringifyProperty } from './helper';
 
 export async function safeTransfer({
```
javadkh2, line 10: Good for now, I will we add safeTransfer to the client-utils as well.

packages/apps/graph/src/services/node-service.ts
```diff
@@ -30,84 +32,93 @@ export type ChainFungibleAccountDetails = {
...
-  );
+  let result;
+  try {
+    result = (await details(
+      accountName,
+      dotenv.NETWORK_ID,
+      chainId as ChainId,
+      dotenv.NETWORK_HOST,
+      fungibleName,
+    )) as any;
```
javadkh2, line 48: Why did you have to cast it to any?

packages/apps/graph/src/services/node-service.ts
```diff
@@ -30,84 +32,93 @@ export type ChainFungibleAccountDetails = {
...
 
-  if (commandResult.result.status !== 'success') {
-    // If the account does not exist on a chain, we get a row not found error.
-    if (
-      (commandResult.result.error as any).message?.includes(
-        'with-read: row not found',
-      )
-    ) {
+    if (typeof result.balance === 'object') {
```
javadkh2, line 50: Thats a weird behaviour in the pact api. Do you think I should change all numbers to the same format in the client. Perhaps always returning the stringified version
```
12.1 => "12.1"
{ decimal : "12.123123123" } => "12.123123123"
```
or maybe convert `12.1` to `{ decimal: "12.1" }`

packages/apps/graph/src/devnet/safe-transfer.ts
MRVDH: That would be awesome!

packages/apps/graph/src/services/node-service.ts
MRVDH: Because Typescript doesn't like the return type: 
image

packages/apps/graph/src/services/node-service.ts
MRVDH: All values to string seems like the most logical thing to do indeed.

packages/apps/graph/src/services/node-service.ts
javadkh2: I see, 
I have addressed this issue in this PR https://github.com/kadena-community/kadena.js/pull/1321. 
I'll let you. know when I merge it

javadkh2 (approved):

packages/apps/graph/src/devnet/simulation/simulate.ts
```diff
@@ -62,7 +62,7 @@ export async function simulate({
     for (let i = 0; i < numberOfAccounts; i++) {
       // This will determine if the account has 1 or 2 keys (even = 1 key, odd = 2 keys)
       const noOfKeys = i % 2 === 0 ? 1 : 2;
-      let account = await generateAccount(noOfKeys);
+      const account = await generateAccount(noOfKeys);
```
nil-amrutlal-dept, line 65: I think this will bring an issue, because when we're funding accounts, in the case of a cross-chain transfer, we alter the account chain, to chain 1. Afterwards, we save the account on `accounts` array which is used to keep track of the existing accounts (with funds). If this is a constant and we do not register that the funded chain was chain 1 and not chain 0, we will run into an error when we try to transfer funds from that account.
image

packages/apps/graph/src/devnet/simulation/simulate.ts
```diff
@@ -82,16 +82,15 @@ export async function simulate({
...
-          sender: sender,
-          receiver: account,
+          sender: {
+            ...sender00,
+            chainId: '0',
+          },
+          receiver: {
+            ...account,
+            chainId: '1',
+          },
```
nil-amrutlal-dept, line 93: We need to persist this alteration

packages/apps/graph/src/devnet/simulation/simulate.ts
```diff
@@ -174,7 +178,7 @@ export async function simulate({
               ...nextAccount,
               chainId: `${getRandomNumber(
                 seededRandomNo,
-                devnetConfig.NUMBER_OF_CHAINS,
+                dotenv.CHAIN_COUNT,
```
nil-amrutlal-dept, line 181: Maybe not in scope of this PR, but if `CHAIN_COUNT` is set to 0, then this will be an issue, because it'll try to make a cross-chain transfer to same chain accounts

packages/apps/graph/src/devnet/simulation/simulate.ts
MRVDH: fixed

packages/apps/graph/src/devnet/simulation/simulate.ts
MRVDH: fixed

packages/apps/graph/src/devnet/simulation/simulate.ts
MRVDH: I'll add the extra check just in case

packages/apps/graph/src/devnet/simulation/simulate.ts
```diff
@@ -172,7 +173,7 @@ export async function simulate({
         let result;
 
         // This is to simulate cross chain transfers
-        if (transferType === 'cross-chain-transfer') {
+        if (transferType === 'cross-chain-transfer' && dotenv.CHAIN_COUNT > 1) {
```
nil-amrutlal-dept, line 176: 🙌🏽

nil-amrutlal-dept (approved):

PR #1343, opened 2023-12-15
Update design-sync cli and re-generate tokens

salamaashoush:


packages/libs/react-ui/design-sync.config.ts
```diff
@@ -14,7 +14,7 @@ const breakpointKeys = Object.keys(breakpoints);
 
 export default defineConfig({
   // you can switch to different branch by changing the #main to #branch-name
-  uri: 'github:kadena-community/design-system/tokens#main',
+  uri: 'github:kadena-community/design-system/tokens#font-weight-tag',
```
eileenmguo, line 17: should we change this back to main?

salamaashoush: yes, I am doing it

eileenmguo (approved):

PR #1346, opened 2023-12-15
fix: navigation is not scrollable on mobile

KristinaSpasevska:

### DIFF
--- a/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
+++ b/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
@@ -24,7 +24,7 @@ export const itemsContainerClass = style([
     flexWrap: 'nowrap',
     width: '100%',
     height: '100%',
-    justifyContent: 'flex-start',
+    justifyContent: 'space-between',
     alignItems: 'center',
     overflowX: 'auto',
   }),
@@ -127,7 +127,6 @@ export const activeLinkClass = style([
 export const childrenClass = style([
   atoms({
     display: 'flex',
-    marginLeft: 'auto',
   }),
   {
     marginRight: vars.sizes.$3,
--- a/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
+++ b/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
@@ -12,10 +12,7 @@ export const containerClass = style([
     height: '$16',
     justifyContent: 'flex-start',
     position: 'relative',
-    width: {
-      xs: 'max-content',
-      sm: '100%',
-    },
+    minWidth: 'max-content',
   }),
   {
     alignItems: 'center',
--- a/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
+++ b/packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
@@ -3,7 +3,7 @@ import { sprinkles, vars } from '../../styles';
 
 export const containerClass = style([
   sprinkles({
-    alignItems: 'stretch',
+    alignItems: 'center',
     backgroundColor: '$gray90',
     color: '$gray40',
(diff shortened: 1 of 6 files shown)


sanderlooijenga (approved):

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -12,10 +12,7 @@ export const containerClass = style([
     height: '$16',
     justifyContent: 'flex-start',
     position: 'relative',
-    width: {
-      xs: 'max-content',
-      sm: '100%',
-    },
+    minWidth: 'max-content',
   }),
```
eileenmguo, line 16: I don't think this results in the behavior you would like. Only the links should be scrollable if I understand correctly? @isaozler

packages/libs/react-ui/src/components/NavHeader/NavHeaderNavigation.tsx
```diff
@@ -21,13 +21,27 @@ export const NavHeaderNavigation: FC<INavHeaderNavigationProps> = ({
...
+        className={glowClass}
+        style={{
+          opacity: !glowX ? 1 : 0,
+          left: '-6rem',
+          transform: `translateX(${glowX}px)`,
+          transitionDuration: `${animationDuration}ms`,
+        }}
+      >
+        <NavGlow />
+      </div>
```
eileenmguo, line 44: Instead of adding an additional glow, can we make the one we already have start at this position?

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
isaozler: image

with minimum space between sections (module links, social icons, and network select) should be scrollable

the logo icon should be fixed

packages/libs/react-ui/src/components/NavHeader/NavHeaderNavigation.tsx
```diff
@@ -34,8 +36,8 @@ export const NavHeaderNavigation: FC<INavHeaderNavigationProps> = ({
           className={glowClass}
           ref={glowRef}
           style={{
-            opacity: glowX ? 1 : 0,
-            transform: `translateX(${glowX}px)`,
+            opacity: 1,
+            transform: `translateX(${glowX + 60}px)`,
```
eileenmguo, line 40: can you add variable names so it's clear what the `60` is for?

packages/libs/react-ui/src/components/NavHeader/useGlow.ts
```diff
@@ -13,7 +13,7 @@ const useGlow = (): IUseGlowReturn => {
   const glowRef = useRef<HTMLDivElement>(null);
   const navRef = useRef<HTMLDivElement>(null);
 
-  const [glowX, setGlowX] = useState(0);
+  const [glowX, setGlowX] = useState(-160);
```
eileenmguo, line 16: also this one

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
```diff
@@ -56,6 +62,7 @@ export const navListClass = style([
...
   {
     listStyle: 'none',
+    paddingInlineStart: '1rem',
```
eileenmguo, line 65: this can be updated to use atoms

eileenmguo:
The dropdown that was previously on the right has moved 
image

packages/libs/react-ui/src/components/NavHeader/NavHeader.css.ts
KristinaSpasevska: this will be included in another PR where we'll adapt to the use of atoms in the pages

eileenmguo (approved):

PR #1347, opened 2023-12-15
Feat/safe transfer

javadkh2:
# safeTransfer 
add `safeTransfer` to `@kadena/client-utils/coin`

the receiver also signs the tx to confirm he/she has access to the account

We used this method:
* If the receiver signs for gas, then we execute just a normal transfer. 
* Otherwise, we add one more code to the transaction to transfer a small amount to the sender from the receiver. We already included this small amount in the initial code, so the sender sends the actual amount plus the small amount

In both scenarios, the receiver also has to sign the transaction to ensure a secure transfer.


packages/libs/client-utils/src/integration-tests/coin.int.test.ts
```diff
@@ -232,3 +233,101 @@ describe('cross chain transfer', () => {
...
+        expect(tx.sigs).toHaveLength(2);
+        expect(tx.sigs[0].sig).toBeTruthy();
+        expect(tx.sigs[1].sig).toBeTruthy();
+      })
+      .execute();
+
+    expect(result).toBe('Write succeeded');
+  });
+
+  it('should fails if receiver doesnt sign', async () => {
```
jessevanmuijden, line 273:
```suggestion
  it('should fail if receiver does not sign', async () => {
```

packages/libs/client-utils/src/integration-tests/coin.int.test.ts
```diff
@@ -232,3 +233,101 @@ describe('cross chain transfer', () => {
     expect(balance).toBe(10);
   });
 });
+
+describe('safeTransfer', () => {
+  it('should transfer kda from sender00 account to receiverAccount and both receiver and sender should sign', async () => {
```
jessevanmuijden, line 238:
```suggestion
  it('should transfer kda from sender00 account to receiverAccount if both receiver and sender sign', async () => {
```

packages/libs/client-utils/src/integration-tests/coin.int.test.ts
```diff
@@ -232,3 +233,101 @@ describe('cross chain transfer', () => {
...
+        sign: createSignWithKeypair([sender00Account]),
+      },
+    );
+
+    await expect(() => task.execute()).rejects.toThrowError(
+      new Error('Signing failed'),
+    );
+  });
+
+  it('should fails if receiver doesnt satisfy the account guard sign', async () => {
```
jessevanmuijden, line 304:
```suggestion
  it('should fail if receiver does not satisfy the account guard sign', async () => {
```

.changeset/wild-pets-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kadena/client-utils': minor
+---
+
+Added safeTransfer which needs both sender and receiver signs
```
jessevanmuijden, line 5:
```suggestion
add safeTransfer which requires both sender and receiver to sign
```

packages/libs/client-utils/src/coin/safe-transfer.ts
```diff
@@ -0,0 +1,76 @@
...
+      signFor(
+        `${contract as 'coin'}.TRANSFER`,
+        sender.account,
+        receiver.account,
+        {
+          decimal: amount,
+        },
+      ),
+    ]),
+    addSigner(receiver.keyset.keys),
```
alber70g, line 65: This is an unbound signature. Is this the safest way to do a safe-transfer?
@EnoF what do you think?

alber70g (commented):
Question on whether this is the way we want to do safe-transfers.

javadkh2:
I think the approach is very conservative since force all keys to sign the Tx. even if the pred is keys-any.
 I will update the PR

alber70g (approved):

PR #1349, opened 2023-12-15
feat(@kadena/react-ui): Updated Tag component to use new tokens and react-aria

eileenmguo:
Updated the Tag component to use the new design tokens and integrated the `useTagGroup` hook from react-aria.

In most cases, `Tags` are used as a group so consumers should use the `TagGroup` and `TagItem` components to compose a tag group. This way, when you access the component via your keyboard, you can tab through the group, use arrow keys to navigate through the tabs, and hit `Backspace` to remove tags (when they are configured with the onRemove prop). Configuration around how the tags behave is placed on the `TagGroup`.

In the Docs we have a case where the tags are used as links - specifically next/link. To accommodate this use case, I have added a `tagAsChild` prop to the `TagGroup` component. There were difficulties adding it to the `TagItem` component because it is wrapping a shared component from react-stately that is multipurpose, but since a lot of the configuration is on the `TagGroup` component and we don't have any instances where we need to configure tags independently, I found that this was the simplest option.

I've removed the onClose props from the `Tag` component so that is only used for styling purposes for the tag group. It is still being exposed so that if someone needs the style of the `Tag` for a specific purpose, they can use it without using the `TagGroup` and the associated accessibility features. (This is the case in the Tools site)

Exported Components: 
- TagGroup (new)
- TagItem (new)
- Tag


packages/libs/react-ui/src/components/Tag/Tag.tsx
```diff
@@ -1,22 +1,15 @@
...
+export const Tag: FC<ITagProps> = ({ children, ...restProps }) => {
   return (
-    <span data-testid="kda-tag" className={tagClass}>
-      <span className={tagLabelClass}>{children}</span>
-      {onClose ? (
-        <button className={closeButtonClass} onClick={onClose}>
-          <SystemIcon.Close size="sm" />
-        </button>
-      ) : null}
+    <span className={tagClass} {...restProps}>
```
sanderlooijenga, line 11: Just to double check if this was done intentionally; this means one can override the `className` 😁

packages/libs/react-ui/src/components/Tag/Tag.stories.tsx
```diff
@@ -1,62 +1,129 @@
-import { Button } from '@components/Button';
-import type { ITagProps } from '@components/Tag';
-import { Tag } from '@components/Tag';
```
sanderlooijenga, line 3: Don't we still want to add this component as well to our Storybook?

packages/libs/react-ui/src/components/Tag/Tag.tsx
eileenmguo: oops! not intentional. I'll update this

packages/libs/react-ui/src/components/Tag/Tag.stories.tsx
eileenmguo: This isn't a common use case, so I didn't, but I can add it if you think it's beneficial?

sanderlooijenga: I wanted to see how this component looks since it is different from the ones being added to the storybook right now. It might be beneficial, but I'll let you decide 😁

packages/apps/docs/src/components/Layout/Blog/Blog.tsx
```diff
@@ -90,16 +89,17 @@ export const Blog: FC<IPageProps> = ({
...
-                    <Link
-                      className={tagLinkClass}
-                      key={tag}
-                      href={`/tags/${tag}`}
-                    >
-                      <Tag>{tag}</Tag>
-                    </Link>
-                  ))}
+                {tags && (
+                  <TagGroup tagAsChild aria-label={`${title} Tags`}>
```
sstraatemans, line 93: I think the taggroup needs an display:inline-block option for this issue (or something).

image

packages/libs/react-ui/src/components/Tag/Tag.stories.tsx
eileenmguo: I'll add it 👍

packages/apps/docs/src/components/Layout/Blog/Blog.tsx
eileenmguo: fixed 👍

sanderlooijenga (approved):

sstraatemans (approved):

PR #1352, opened 2023-12-18
Refactor/decomposed field wrapper

ferreroltd:


packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -1,29 +1,58 @@
...
 export const TextField: FC<ITextFieldProps> = ({
   disabled = false,
-  inputProps,
   status,
+  leadingText,
+  startIcon,
+  leadingTextWidth,
+  type,
+  ref,
+  id,
```
eileenmguo, line 22: props that are intended for accessibility can be removed from props since react-aria should provide them

packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -1,29 +1,58 @@
...
-      htmlFor={id}
-      disabled={disabled}
-      status={status}
-      {...rest}
-    >
-      <Input disabled={disabled} {...inputProps} />
-    </FormFieldWrapper>
+    <div className={statusVal ? statusVariant[statusVal] : undefined}>
+      {label !== undefined && (
+        <FormFieldHeader htmlFor={id} label={label} tag={tag} info={info} />
```
eileenmguo, line 35: This only passes `htmlFor`. You'll want to make sure to pass all react-aria props to the header (I think. Not sure if it should be only the title).

packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -1,29 +1,58 @@
...
+          startIcon={startIcon}
+          type={type}
+          ref={ref}
+          id={id}
+          outlined={outlined}
+          {...rest}
+        />
+      </Stack>
+      {Boolean(helperText) && status !== 'negative' && (
+        <FormFieldHelper>{helperText}</FormFieldHelper>
```
eileenmguo, line 51: you'll also want to pass react-aria props to the description and error texts

packages/apps/tools/src/components/Global/PublicKeyField/index.tsx
```diff
@@ -9,6 +9,7 @@ import { publicKeyFieldStyles } from './styles.css';
 
 interface IPublicKeyFieldProps
   extends Partial<Omit<IFormFieldWrapperProps, 'children'>> {
```
eileenmguo, line 11: I believe you can just extend `ITextFieldProps`

packages/libs/react-ui/src/components/Form/Input/Input.tsx
```diff
@@ -27,26 +29,26 @@ export interface IInputProps
...
+ * @deprecated Use `TextField` instead.
+ */
 export const Input: FC<IInputProps> = forwardRef<HTMLInputElement, IInputProps>(
   function Input(
     {
       outlined,
       leadingText,
       startIcon,
-      leadingTextWidth: propLeadingTextWidth,
+      leadingTextWidth,
```
eileenmguo, line 44: you can remove leading textWidth entirely. It will not work with atoms anyways

packages/libs/react-ui/src/components/Form/TextField/TextField.tsx
```diff
@@ -1,29 +1,57 @@
...
+import { statusVariant } from '../FormFieldWrapper/FormFieldWrapper.css';
 
 export interface ITextFieldProps
-  extends Omit<IFormFieldWrapperProps, 'children' | 'htmlFor'> {
-  inputProps: Omit<IInputProps, 'disabled' | 'children' | 'leadingTextWidth'>;
-}
-
+  extends Omit<IFormFieldWrapperProps, 'children' | 'htmlFor'>,
+    Omit<IInputProps, 'disabled' | 'children'> {}
+// remove exports for textArea, select and input
```
eileenmguo, line 12: cleanup?

sanderlooijenga:
Filling in a public key (like `567223a8fac06e78a8e5b5bcbd02e9b4d839402e2e509e0f657aaf79cc30c40b`) doesn't work anymore, while that same key does work on the [current Tools dApp](https://tools.kadena.io/faucet/new)

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
```diff
@@ -91,37 +91,31 @@ export const AddNetworkModal: FC<IAddNetworkModalProps> = (props) => {
...
                       label={t('Network label')}
-                      inputProps={{
-                        id: 'label',
-                        ...register('label'),
-                        onChange: (e) => setLabel(e.target.value),
-                        value: label,
-                        placeholder: 'devnet',
-                      }}
+                      id="label"
+                      {...register('label')}
```
salamaashoush, line 95: I know it was there before but I wonder why use `register` in this case since you also use the component in controlled by passing `value` and `onChange`, I would recommend removing them and use `register` only, same for the other fields

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -121,7 +121,11 @@ const NewAccountFaucetPage: FC = () => {
 
   const onFormSubmit = useCallback(
     async (data: FormData) => {
+      console.log(data, pubKeys);
```
salamaashoush, line 124: cleanup

packages/apps/tools/src/components/Global/AddNetworkModal/AddNetworkModal.tsx
ferreroltd: @sanderlooijenga this change is actually how it would work without using the forwardRef, is this an okay change for tools?

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -293,13 +297,11 @@ const NewAccountFaucetPage: FC = () => {
...
                   helperText={errors?.pubKey?.message}
-                  inputProps={{
-                    ...register('pubKey', {
-                      onChange: () => {
-                        clearErrors('pubKey');
-                      },
-                    }),
-                  }}
+                  {...register('pubKey', {
+                    onChange: () => {
```
salamaashoush, line 301: you can remove this and use `mod: 'onChange'` for the `useForm` hook instead

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -309,6 +311,7 @@ const NewAccountFaucetPage: FC = () => {
                   onClick={() => {
                     const value = getValues('pubKey');
                     const valid = validatePublicKey(value || '');
+                    console.log('add pub key', value, valid);
```
salamaashoush, line 314: cleanup

salamaashoush (approved):
it looks solid 
1.  some old stuff should be refactored (about mixing `react-hook-form` with controlled component props), it is old code and could be done sepratly
2. Clean up console logs

packages/apps/tools/src/pages/faucet/new/index.tsx
sanderlooijenga: We can't, the `clearErrors` needs to be called inside `onChange` since we manually call `setError` elsewhere. So it's correct as it is 😁

salamaashoush: in this case, I would recommend using a custom validation function instead of manually setting the errors, you can use zod refinements for that purpose https://zod.dev/?id=refine

salamaashoush: but I said in the other comments it is not a concern for this PR, and we can improve or keep as it is later if it is needs that

sanderlooijenga (approved):

packages/libs/react-ui/src/components/Form/SelectField/SelectField.tsx
```diff
@@ -1,26 +1,52 @@
...
-      <Select disabled={disabled} {...selectProps}>
-        {children}
-      </Select>
-    </FormFieldWrapper>
+    <>
+      <div className={statusVal ? statusVariant[statusVal] : undefined}>
+        {label !== undefined && (
+          <FormFieldHeader htmlFor={id} label={label} tag={tag} info={info} />
+        )}
+        <Stack gap="$2" direction="column">
```
eileenmguo, line 38: minor cleanup: you can remove the Stack

packages/libs/react-ui/src/components/Form/TextareaField/TextareaField.tsx
```diff
@@ -1,23 +1,38 @@
...
 
   return (
-    <FormFieldWrapper htmlFor={id} disabled={disabled} {...rest}>
-      <Textarea disabled={disabled} {...textAreaProps} />
-    </FormFieldWrapper>
+    <div className={statusVal ? statusVariant[statusVal] : undefined}>
+      {label !== undefined && (
+        <FormFieldHeader htmlFor={id} label={label} tag={tag} info={info} />
+      )}
+      <Stack gap="$2" direction="column">
```
eileenmguo, line 27: small cleanup here as well

eileenmguo (approved):

PR #1353, opened 2023-12-18
Test/tools e2e

Ghislain89:
This PR has a few changes:

**Overal architecture changes**
In order to keep CI pipelines generic, i've created a baseConfig on which specific dApp configs depend. This allows for dApp specific settings, without polluting the base framework config.
```
playwright.base.config.ts <-- contains basic, shareable configuration for the framework
playwright.tools.config.ts <--- contains dApp specific configuration such as start command for webserver and environment variables. 
```

**CI**
Instead of using mocks, CI now starts up devnet, deploys the faucet contracts and then connects the Tools dApp to this deployed version of devnet.  This allows for actuall integration of most (if not all) scenario's

**i18n translations**
As a best practice, we try to use accessibility locators to identify elements, as this ensures that our test interacts with the system under test in a similar manner as how an end user would. In order to prevent something simple as changing a label to a new value from resulting in test failures. I have introduced i18n to allow for keys to be instead instead of the translated value. The i18n fixture imports the translation files and handles translating the key to the desired value.  In below Example Chain ID is the key.
Example:
```TS
 await this._card.setValueForCombobox(
      this._i18n.t(`Chain ID`),
      account.chainId,
    );
```

I'd like to merge this, and continue on the testcases for the transaction page.


packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -325,6 +325,7 @@ const NewAccountFaucetPage: FC = () => {
                     });
                   }
                 }}
+                aria-label="Add Public Key"
```
Ghislain89, line 328: @sanderlooijenga @KristinaSpasevska I added a label here so I can identify the button.

alber70g (approved):

PR #1355, opened 2023-12-18
feat(types): export predicate type

jessevanmuijden:

### DIFF
--- a/packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
+++ b/packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
@@ -1,20 +1,19 @@
 import type { ISelectProps } from '@kadena/react-ui';
 import { FormFieldWrapper, Select } from '@kadena/react-ui';
+import type { TPredicate } from '@kadena/types';
 
 import type { FC, FormEventHandler } from 'react';
 import React, { useCallback } from 'react';
 
-export type PredKey = 'keys-all' | 'keys-any' | 'keys-2';
-
-export type OnPredSelectChange = (value: PredKey) => void;
+export type OnPredSelectChange = (value: TPredicate) => void;
 
 const ELEMENT_ID = 'select-pred';
 
-const predicates: Array<PredKey> = ['keys-all', 'keys-any', 'keys-2'];
+const predicates: Array<TPredicate> = ['keys-all', 'keys-any', 'keys-2'];
 
 const PredKeysSelect: FC<
   Omit<ISelectProps, 'children' | 'value' | 'onChange' | 'icon' | 'id'> & {
-    value: PredKey;
+    value: TPredicate;
     onChange: OnPredSelectChange;
   }
 > = ({ value, onChange, ...rest }) => {
--- a/packages/libs/types/src/Predicate.ts
+++ b/packages/libs/types/src/Predicate.ts
@@ -1,0 +1,7 @@
+/**
+ * The most common predicate values are 'keys-all', 'keys-2' and 'keys-any'.
+ * However, it is allowed to use any string. If a predicate is undefined,
+ * the default value used by Pact is 'keys-all'.
+ * @alpha
+ */
+export type TPredicate = string | undefined;
--- a/.changeset/tall-baboons-smile.md
+++ b/.changeset/tall-baboons-smile.md
@@ -1,0 +1,8 @@
+---
+'@kadena/client-utils': patch
+'@kadena/client': patch
+'@kadena/graph': patch
+'@kadena/types': patch
+---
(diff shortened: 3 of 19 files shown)


nil-amrutlal-dept (approved):

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
```diff
@@ -1,20 +1,19 @@
...
 
-export type PredKey = 'keys-all' | 'keys-any' | 'keys-2';
-
-export type OnPredSelectChange = (value: PredKey) => void;
+export type OnPredSelectChange = (value: TPredicate) => void;
 
 const ELEMENT_ID = 'select-pred';
 
-const predicates: Array<PredKey> = ['keys-all', 'keys-any', 'keys-2'];
+const predicates: Array<TPredicate> = ['keys-all', 'keys-any', 'keys-2'];
```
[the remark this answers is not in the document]
jessevanmuijden: I believe that the tools website team decided to limit the possible predicates to this string.

packages/libs/types/src/Predicate.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * The most common predicate values are 'keys-all', 'keys-2' and 'keys-any'.
+ * However, it is allowed to use any string. If a predicate is undefined,
+ * the default value used by Pact is 'keys-all'.
+ * @alpha
+ */
+export type TPredicate = string | undefined;
```
javadkh2, line 7: I think we also need to expose a type like `CommonPredicate`.
and then in the client use it instead of `TPredicate ` to avoid breaking type suggestion.
I will add one more overload to the functions in the client to allow normal string.
```suggestion
export type TPredicate = string | undefined;
export type CommonPredicate = "keys-all" | "keys-2" | "keys-any"
```

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
```diff
@@ -1,20 +1,19 @@
 import type { ISelectProps } from '@kadena/react-ui';
 import { FormFieldWrapper, Select } from '@kadena/react-ui';
+import type { TPredicate } from '@kadena/types';
```
javadkh2, line 3: I don't think we have the convention of prefixing types with `T`. 
So why not just use `Predicate`?

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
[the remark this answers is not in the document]
javadkh2: if we use `CommonPredicate` then
```suggestion
const predicates: Array<CommonPredicate> = ['keys-all', 'keys-any', 'keys-2'];
```

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
jessevanmuijden: I prefer without the prefix

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
```diff
@@ -1,20 +1,19 @@
...
-const predicates: Array<PredKey> = ['keys-all', 'keys-any', 'keys-2'];
+const predicates: Array<TPredicate> = ['keys-all', 'keys-any', 'keys-2'];
 
 const PredKeysSelect: FC<
   Omit<ISelectProps, 'children' | 'value' | 'onChange' | 'icon' | 'id'> & {
-    value: PredKey;
+    value: TPredicate;
     onChange: OnPredSelectChange;
   }
 > = ({ value, onChange, ...rest }) => {
```
alber70g, line 19: Why is this `value` predicate limited to the 3 that are built-in?

alber70g (changes requested):
I don't think capturing `predicate` as only one of the 3 values is valuable. We've had the discussion before, where the result was that essentially we only have `string` as type for predicate. Unless we know what predicates are available (which can be any on the blockchain), the type should be just `string` and the options should be shown in the docs/intellisense.

Typescript doesn't support `'keys-all' | ... | string` as this will be "flattened" to `string` anyway, and looses any value to have the union there. So until Typescript supports that, I think it should just be `string`.

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
jessevanmuijden: I don't know. I think it has to do with the functionality that was in scope for the tools website.

Randynamic:
Status?

PR #1356, opened 2023-12-18
chore: added tokens in form

ferreroltd:
This pr adds the new tokes system to the Form folder of components


eileenmguo (approved):
I believe this is a separate issue, but the copy version of the input component is still focusable when it is disabled

.changeset/brown-needles-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kadena/react-ui': patch
+---
+
+Used new token system
```
eileenmguo, line 5: It would be good if these comments are more descriptive. What components were updated for example

PR #1365, opened 2023-12-20
Extract tracker label from tracker card

EnoF:
- this should allow the tracker pattern to be used without forcing the use of TrackerCard component


eileenmguo (commented):
The TrackerCard will need to be refactored/removed completely as it is not very flexible and specific to the tools, however we do want to figure out what elements are useful and potentially pull those into components. In this case I see you've pulled out a versatile piece specifically for accounts which is an example of that. Let's sit together in the new year to discuss which parts of this we should convert into composable components

PR #1375, opened 2023-12-20
[@kadena/graph] Indexes for most used columns

MRVDH:


packages/apps/graph/cwd-extra-migrations/2.3.0.4_add_indexes_to_most_looked_up_columns.sql
nil-amrutlal-dept, line 1: There are also some incremental Id's that were implemented for example, on block and event table. Would it not be valuable to add indexes to those as well? 
These are referred [here](https://app.asana.com/0/1205801361945248/1206001849982337/f) as incremental columns

MRVDH: Yep good catch, should have mentioned that in the description. I considered adding the indexes for the incremental ids, but I found out that we actually don't use those incremental ids at all. They are not even exposed in the graph, so I felt that it would be unnecessary at this point, but maybe in the future.

MRVDH: Nevermind, they are used on blocks and events. Added to the sql script 👍

packages/apps/graph/prisma/schema.prisma
nil-amrutlal-dept, line 1: Is the `@@index` for blocks not missing?

MRVDH: It's on the map of the id field. I noticed for events it was incorrect so I added it there now though

nil-amrutlal-dept (approved):

PR #1376, opened 2023-12-20
feat: refactor fs methods in cli to use service

barthuijgen:


packages/tools/kadena-cli/src/services/fs.service.ts
```diff
@@ -1,13 +1,17 @@
+import type { Dirent } from 'node:fs';
 import fs from 'node:fs/promises';
-import path from 'path';
+import { dirname } from 'path';
```
nillo, line 4: i like this service!

nillo (approved):

PR #1377, opened 2023-12-20
[@kadena/graph] Marmalade Simulation

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206142839462226


MRVDH (approved):

packages/apps/graph/src/devnet/marmalade/token/create-token-id.ts
```diff
@@ -0,0 +1,46 @@
+import type { IAccount } from '@devnet/helper';
+import { Pact, readKeyset } from '@kadena/client';
+import { dirtyReadClient } from '@kadena/client-utils/core';
+import {
+  addKeyset,
+  composePactCommand,
+  execution,
+  setMeta,
+} from '@kadena/client/fp';
+import { dotenv } from '@utils/dotenv';
```
alber70g, line 10: I'm usually wrapping this into another object that does the parsing and validation of the env-vars and then re-exporting with proper types, as `dotenv` props are only `strings`

alber70g (approved):

nil-amrutlal-dept:
Hey @javadkh2. As we talked previously, I have developed the `marmalade-v2` interface functions as similarly to `client-utils` as I could. Can you please give them a look when you have some time? 
The ones I am referring to in particular are `create-token-id`, `create-token`, `mint-token`, and `transfer-token`

packages/apps/graph/src/devnet/marmalade/token/create-token-id.ts
nil-amrutlal-dept: If you check our `dotenv` object, we can see that not all the properties are `string`

PR #1380, opened 2023-12-21
Update new button component to use the new tokens

salamaashoush:


packages/libs/react-ui/src/components/Button/NewButton.css.ts
```diff
@@ -13,128 +11,206 @@ const outlineColor = createVar();
 const backgroundColor = createVar();
 const color = createVar();
```
isaozler, line 12: missing the iconColor

ferreroltd (approved):

eileenmguo (commented):
It looks like the `isCompact` arg isn't doing anything on storybook. 

Before this PR is merged, I think the old button should be removed and replaced with this new one so that the apps are updated as well 🎄

salamaashoush:
> It looks like the `isCompact` arg isn't doing anything on storybook.
> 
> Before this PR is merged, I think the old button should be removed and replaced with this new one so that the apps are updated as well 🎄

Yes there was cascading issues for the padding since atoms have low priority than styles in vanilla extract

to this PR small and easy to review I will create a separate MR to update the apps and remove the old button

packages/libs/react-ui/src/components/Button/NewButton.css.ts
salamaashoush: done

isaozler (approved):

PR #1388, opened 2023-12-22
[@kadena/graph] Restructure devnet folder and files (separation of concerns)

nil-amrutlal-dept:
In this PR I tried to keep things as clean as possible:
- separated deployment and simulation into different folders
- separated marmalade and coin simulations into different folders
- split helper file into two helper files, one on each 'subject' folder
- created general util files for exports used throughout the project (example: `logger.ts`


packages/apps/graph/src/devnet/deployment/helper.ts
```diff
@@ -0,0 +1,72 @@
...
+  createClient(
+    ({ chainId }) =>
+      `${dotenv.NETWORK_HOST}/chainweb/0.0/${dotenv.NETWORK_ID}/chain/${chainId}/pact`,
+  );
+
+export const submit = (tx: ICommand): Promise<ITransactionDescriptor> =>
+  getClient().submit(tx);
+
+export const listen = (tx: ITransactionDescriptor): Promise<ICommandResult> =>
+  getClient().listen(tx);
```
alber70g, line 35: Any particular reason to recreate the client every time you need it? You could use the same client by calling `getClient` once.

packages/apps/graph/src/devnet/deployment/helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    const signedTx = keyPairs.reduce((acc, keyPair) => {
+      const { sig } = sign(acc.cmd, {
+        publicKey: keyPair.publicKey,
+        secretKey: keyPair.secretKey,
+      });
+      if (!sig) throw Error('Failed to sign transaction');
+      return addSignatures(acc, { sig, pubKey: keyPair.publicKey });
+    }, tx);
+    return signedTx;
+  };
```
alber70g, line 49: There is a [`createSignWithKeypair` function in `@kadena/client`](https://github.com/kadena-community/kadena.js/blob/1f6cadbee5cbf01a67ca2950d7e3a82f99f07705/packages/libs/client/src/signing/keypair/createSignWithKeypair.ts#L49-L69) that you can use. This can simplify things when you have multiple keys. It creates a "keyring" that you can then use to sign with.

alber70g (approved):

packages/apps/graph/src/devnet/deployment/helper.ts
nil-amrutlal-dept: Great catch. Changes made :)

packages/apps/graph/src/devnet/deployment/helper.ts
nil-amrutlal-dept: Perfect! Indeed it makes it a lot cleaner 👐🏽  Implemented changes

PR #1391, opened 2023-12-22
feat(client-utils): asyncPipe from

javadkh2:
working on a way to start a pipeline from a point 
for example just run the continuation of the crossChain transfer  
```TS
const transfer = crossChainTransfer(....)
transfer
.from("submit", { requestId, networkId, chianId })
.executeTo("listen-continuation")
```


Randynamic:
Can we create docs for this as well?

packages/libs/client-utils/vitest.config.ts
```diff
@@ -3,7 +3,7 @@ import { defineConfig, mergeConfig } from 'vitest/config';
 
 const localConfig = defineConfig({
   test: {
-    exclude: ['src/**/*.int.test.ts'],
+    exclude: ['src/**/*.int.test.ts', 'src/interfaces/**/*'],
```
alber70g, line 6: Was this on purpose?

alber70g (approved):

PR #1394, opened 2023-12-27
[@kadena/graph-client] Added query information on each page

MRVDH:
chrome_1vc8fPoF9Y


packages/apps/graph-client/src/components/graphql-query-dialog/graphql-query-dialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+        GraphQL
+      </Button>
+      <Dialog
+        isOpen={isOpen}
+        onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+      >
+        {() => (
+          <DialogContent>
+            <p>Query</p>
+            {queries.map((query, index) => (
```
nil-amrutlal-dept, line 33: If there's more than one query with variables, does it still make sense to represent them in the same `<DialogContent>`. My concern is only if the variables happen to have the same name, it can be confusing to determine which one belongs to which.

nil-amrutlal-dept:
Looks good 🙌🏽  just left a small comment. Also, maybe it would be nice to have it on the right side of the page? Just because I think its very close to the breadcrumbs. In any case, it is prob going to be changed again when we have a design so wouldn't bother too much with it 
image

nil-amrutlal-dept (approved):

packages/apps/graph-client/src/components/graphql-query-dialog/graphql-query-dialog.tsx
MRVDH: Yep you're right. I have redone it to make more sense. Let me know what you think!

chrome_1vc8fPoF9Y

MRVDH:
> Looks good 🙌🏽 just left a small comment. Also, maybe it would be nice to have it on the right side of the page? Just because I think its very close to the breadcrumbs. In any case, it is prob going to be changed again when we have a design so wouldn't bother too much with it ~snip~

Makes more sense indeed. Moved to the right!

PR #1402, opened 2023-12-29
[@tools-website] Fix transactions module with custom network

KristinaSpasevska:


packages/apps/tools/src/services/transfer-tracker/get-transfer-status.ts
```diff
@@ -156,26 +159,39 @@ export async function getXChainTransferInfo({
...
   senderChain: ChainwebChainId;
   receiverChain: ChainwebChainId;
   network: Network;
+  networksData: INetworkData[];
   t: Translate;
 }): Promise<IStatusData> {
   debug(getXChainTransferInfo.name);
   try {
-    const networkId = chainNetwork[network].network;
+    // const { selectedNetwork: network, networksData } = useWalletConnectClient();
```
sanderlooijenga, line 175: Can we remove this?

packages/apps/tools/src/services/transfer-tracker/get-transfer-status.ts
```diff
@@ -156,26 +159,39 @@ export async function getXChainTransferInfo({
...
+    const networkData: INetworkData | undefined = networksData.find(
+      (item) => (network as Network) === item.networkId,
+    );
+
+    if (!networkData) return { error: 'No network found' };
 
     const requestObject = {
       requestKey,
-      networkId,
+      networkId: network,
```
sanderlooijenga, line 184: On other places `networkData.networkId` is used, should we do that here as well, to keep it consistent? 

Also, we're doing this network logic now in multiple places (inside several `services`), should we refactor it to a reusable function perhaps?

packages/apps/tools/src/services/transfer-tracker/get-transfer-status.ts
```diff
@@ -187,11 +203,16 @@ export async function getXChainTransferInfo({
         rollback: false,
         step: 1,
       })
-      .setNetworkId(networkId)
+      .setNetworkId(network)
```
sanderlooijenga, line 206: Same as comment above

packages/apps/tools/src/utils/network.ts
```diff
@@ -89,6 +89,9 @@ export const getApiHost = ({
   if (!api.startsWith('http')) {
     scheme = 'https://';
   }
+  if (api.includes('localhost')) {
+    scheme = 'http://';
```
sanderlooijenga, line 93: What has changed that is now needed?

KristinaSpasevska: if the user enters only `localhost:8080` as a value in the field '`network api`' when adding new network in will prefix it with '`https://`' and I wanted to handle this case

packages/apps/tools/src/services/transfer-tracker/get-transfer-status.ts
KristinaSpasevska: We already have a `getApiHost` function and `client` to create different clients (for sender and receiver networks and chains). Other than these two reusable functions, what else do you think can be reusable? Generating the request object? Yet again we would need to add all these values as parameters. 
Let me know what do you think

packages/apps/tools/src/utils/network.ts
sanderlooijenga: Alright 👍

packages/apps/tools/src/services/transfer-tracker/get-transfer-status.ts
sanderlooijenga: Ah right, yeah let's just leave as is then for now 👌

sanderlooijenga (approved):

PR #1426, opened 2024-01-04
New LinkButton component

salamaashoush:


isaozler (approved):

packages/libs/react-ui/src/index.ts
```diff
@@ -122,3 +122,6 @@ export {
   maskValue,
   useDialog,
 } from './components';
+
+// re-exports from 'react-aria'
+export { RouterProvider } from 'react-aria';
```
eileenmguo, line 127: Can you add some documentation or examples of how this component would work with next/link?

salamaashoush: it is already documented here https://react-spectrum.adobe.com/react-aria/routing.html, I will add the link in storybook

eileenmguo:
I see that my previous comment didn't get posted. Are you planning to remove the old Button component in a separate PR?

salamaashoush:
Yes i am on it already it will be separate MR

eileenmguo (approved):
I assume there are instances of this being used in the docs. Can we update them as part of this PR?

salamaashoush:
> I assume there are instances of this being used in the docs. Can we update them as part of this PR?

if means the LinkButton/Button I will a separate PR for them, there is no need to block this MR for the apps since it will require more review from owners and also needs a bit of visual test

PR #1428, opened 2024-01-05
[@kadena/graph-client] Create Non-Fungible Account Pages (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206264041352933


packages/apps/graph-client/src/constants/non-fungible.ts
```diff
@@ -0,0 +1 @@
+export const NONFUNGIBLETRANSACTION: string = 'marmalade-v2.ledger';
```
MRVDH, line 1: Let's underscore constants:

```suggestion
export const NON_FUNGIBLE_TRANSACTION: string = 'marmalade-v2.ledger';
```

MRVDH (approved):
I trust your html/css skills 😄

PR #1429, opened 2024-01-05
chore: fix format + linting errors

realdreamer:


packages/tools/kadena-cli/.eslintrc.cjs
```diff
@@ -4,5 +4,8 @@ require('@rushstack/eslint-config/patch/modern-module-resolution');
 module.exports = {
   extends: ['@kadena-dev/eslint-config/profile/lib'],
   parserOptions: { tsconfigRootDir: __dirname },
-  "ignorePatterns": ["src/dapp/templates/**", "src/dapp/pact/**" ]
+  "ignorePatterns": ["src/dapp/templates/**", "src/dapp/pact/**" ],
+  rules: {
+    '@typescript-eslint/strict-boolean-expressions': 'off',
+  },
```
alber70g, line 10: Let's not make this rule less stringent. It's a signal if there are too many options in the boolean expression.

packages/tools/kadena-cli/package.json
```diff
@@ -84,6 +84,6 @@
     "eslint": "^8.45.0",
     "prettier": "~3.0.3",
     "typescript": "5.2.2",
-    "vitest": "^0.34.6"
+    "vitest": "^1.1.0"
```
alber70g, line 87: Should it be updated in all other packages as well?

packages/tools/kadena-cli/src/keys/utils/storage.ts
```diff
@@ -145,7 +145,7 @@ export async function readKeyFileContent(
   filePath: string,
 ): Promise<TSeedContent | IKeyPair | undefined> {
   const fileContents = await services.filesystem.readFile(filePath);
-  if (fileContents === null) {
+  if (!fileContents) {
```
alber70g, line 148: Is this the reason to have `strict-boolean-expressions` disabled?

This should be compared to `null` if that's what we expect since, `readFile: (path: string) => Promise<string | null>;` If you want to also compare for path length, you should use `if(fileContents === null && fileContents.length === 0)`.

You could consider `fs.exists` or something similar if you're actually interested in whether the file exists


Edit: I see you changed the signature of the function to return `void`. You can check for `undefined`

packages/tools/kadena-cli/src/utils/test.util.ts
```diff
@@ -18,11 +18,14 @@ class OutputCapture {
...
     this._print = print;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     process.stdout.write = (chunk, encoding?: any, callback?: any) => {
       this._stdout.push(chunk.toString());
       if (this._print) this._originalStdoutWrite(chunk, encoding, callback);
       return true;
     };
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     process.stderr.write = (chunk, encoding?: any, callback?: any) => {
```
alber70g, line 29: We can use `unknown` since it covers the intent. We don't know what it is that we get passed here.

alber70g (approved):

packages/tools/kadena-cli/package.json
realdreamer: all other packages are already `1.1.0` only cli package was behind because of feature repo. But, I think this is already updated. I could see some merge conflicts. I'll rebase and check it once

packages/tools/kadena-cli/src/keys/utils/storage.ts
realdreamer: nope, this is not the reason for disabling the rule. `!fileContents` works fine(without TS error) for both `undefined` or `null` case. 

i don't remember exactly why I disabled. I guess there were lots of warnings previously but not anymore.

realdreamer:
closing this PR! some of the changes are not relevant any more. so create a new PR for only fixing format issues

PR #1430, opened 2024-01-05
feat(@kadena/react-ui): ComboBox Component

sanderlooijenga:
The ComboBox component is based on React Aria's [ComboBox](https://react-spectrum.adobe.com/react-aria/useComboBox.html).

It is needed because of the [following UI](https://transfer.chainweb.com/transfer-ledger-create.html) needed to be implemented (click _"Change/Verify Ledger Account"_)

It can be seen/tested [here](https://react-ui-git-feat-react-ui-combobox-kadena-js.vercel.app/?path=/docs/form-combobox--docs).

See also as a reference;
- https://component.gallery/components/combobox/

---
  - https://app.asana.com/0/0/1206220622047570


packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <div style={{ width }}>
+        {/* @ts-expect-error */}
+        <TextField
+          {...inputProps}
+          label={label}
+          id={id}
+          startIcon={startIcon}
+          ref={inputRef}
+        >
+          <Button {...buttonProps} ref={buttonRef}>
```
salamaashoush, line 63: could you use the Button from NewButton instead, it also has a text variant which can fit more in this use case

packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+import React, { useRef } from 'react';
+import type { AriaComboBoxProps } from 'react-aria';
+import { mergeProps, useComboBox, useFilter } from 'react-aria';
+import { useComboBoxState } from 'react-stately';
+
+export interface IComboBoxProps<T>
+  extends Omit<AriaComboBoxProps<T>, 'id' | 'label'> {
+  id: IInputProps['id'];
+  label: ITextFieldProps['label'];
+  width: CSSProperties['width'];
```
salamaashoush, line 16: why do we need width, you can get the size of the popover from the text input ref and make all the combo box just flow with the content for example when it is inside a flex/grid container

packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+      inputRef,
+      buttonRef,
+      listBoxRef,
+      popoverRef,
+    },
+    state,
+  );
+
+  return (
+    <div style={{ display: 'inline-flex', flexDirection: 'column' }}>
```
eileenmguo, line 53: you can use inline sprinkles for this or the Stack component
```suggestion
    <div className={sprinkles({ display: 'inline-flex', flexDirection: 'column' })}>
```

packages/libs/react-ui/src/components/ListBox/ListBox.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <ul
+      {...listBoxProps}
+      ref={listBoxRef as RefObject<HTMLUListElement>}
+      style={{
+        margin: 0,
+        padding: 0,
+        listStyle: 'none',
+        maxHeight: 150,
+        overflow: 'auto',
+      }}
```
eileenmguo, line 63: can we move this to a class?

packages/libs/react-ui/src/components/Form/ComboBox/index.ts
```diff
@@ -0,0 +1 @@
+export { ComboBox, type IComboBoxProps } from './ComboBox';
```
eileenmguo, line 1: The team decided to only exporting `Field` components moving forward that utilize the react-aria hooks since they add aria props to labels, error messages, etc. 

The PR for the `TextField` component has not been merged yet so it is probably better to wait until that one is completed to apply it to this component, but would be good to add a TODO to add the `Field` props in the future.

packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
sanderlooijenga: Fixed by using the Stack

packages/libs/react-ui/src/components/ListBox/ListBox.tsx
sanderlooijenga: Done!

packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,88 @@
...
+  label: ITextFieldProps['label'];
+  width: CSSProperties['width'];
+  startIcon?: ITextFieldProps['startIcon'];
+}
+
+/**
+ * @see https://react-spectrum.adobe.com/react-aria/useComboBox.html
+ */
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const ComboBox = <T extends object>({
```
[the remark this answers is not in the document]
sanderlooijenga: Done, thanks for suggestion

packages/libs/react-ui/src/components/Form/ComboBox/ComboBox.tsx
sanderlooijenga: I did use the Icon now, but not the text variant yet. Do you think that would still indicate enough that it's a clickable element that can/will open the options list?

PR #1434, opened 2024-01-08
used starticon in Breadcrumbs and exposed individual components for composing

ferreroltd:
This PR aligns the Breadcrumbs component to be able to compose it with the inner components, this also exposed the prop startIcon to receive any type of icon element


packages/libs/react-ui/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -25,18 +26,14 @@ const meta: Meta<
...
       description: {
         component:
-          'The Breadcrumb component displays the position of the current page within the site hierarchy, allowing page visitors to navigate the page hierarchy from their current location. It uses a composition of the `Root` and `Item` subcomponents to define the paths and structure of the entire breadcrumb component.<br><br><i>Note: In times when you need to use an external `Link` component (like next/link in Next.js), you can wrap the external component in `Breadcrumb.Item` and set the `asChild` prop to pass on styles and props to the child component.</i>',
+          'The Breadcrumb component displays the position of the current page within the site hierarchy, allowing page visitors to navigate the page hierarchy from their current location. It is composed by BreadcrumbsContainer and BreadcrumbsItem.<br><br><i>Note: In times when you need to use an external `Link` component (like next/link in Next.js), you can wrap the external component in BreadcrumbsItem and set the `asChild` prop to pass on styles and props to the child component.</i>',
       },
     },
   },
   argTypes: {
-    icon: {
+    startIcon: {
```
eileenmguo, line 34: I think it's better to keep the name `icon` since there will only ever be one icon. What do you think?

packages/libs/react-ui/src/components/Breadcrumbs/index.tsx
```diff
@@ -1,17 +1,4 @@
...
-  Item: FC<IBreadcrumbItemProps>;
-}
-
-export type { IBreadcrumbItemProps, IBreadcrumbsProps };
-
-export const Breadcrumbs: IBreadcrumbs = {
-  Root: BreadcrumbsContainer,
-  Item: BreadcrumbsItem,
-};
+export { BreadcrumbsContainer } from './Breadcrumbs';
```
eileenmguo, line 1: In other components, the "container" component is just named what the component is. So in this case I think it would be better if this was just exported as `Breadcrumbs`

so when used it would look like this: 
```
<Breadcrumbs>
  <BreadcrumbItem>Item<BreadcrumbItem/>
</Breadcrumbs>
```

ferreroltd: I like this indeed will add the change

packages/libs/react-ui/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
ferreroltd: I thought of this, but since we did it like that in Select for example I kept the same "idea" so it tells a little bit of the positioning, but i this idea, dont mind refactoring in both places

eileenmguo: Yeah, I think this one could go either way so I'm fine with this 👍

realdreamer: imo usually bread crumbs have icons at the starting position only. So, its better to have it as `icon` when there is no possibility of having icon at the end.

sanderlooijenga (approved):
Approved on behalf of the changes inside tools dApp 👍

MRVDH (approved):
Approved for the Graph project

realdreamer (approved):
Approving for the docs project

eileenmguo (approved):

PR #1436, opened 2024-01-08
Added new favicon

Randynamic:


sanderlooijenga:
On the [deployed preview](https://tools-git-fix-tools-favicon-kadena-js.vercel.app/) I don't see the favicon, should it appear there already?

sanderlooijenga:
> On the [deployed preview](https://tools-git-fix-tools-favicon-kadena-js.vercel.app/) I don't see the favicon, should it appear there already?

I think it should, looking at the deployed preview of an other branch; https://tools-git-feat-tools-websitetrack-and-trace-pa-300bd5-kadena-js.vercel.app/

sstraatemans:
there is a `common/images` package in the mono repo.
there are all the icons. 
for docs we copy these to our package on build time. so we have 1 place where the favicons live.

or maybe they can be added to the react-ui?

Randynamic:
Don't we want to have different favicons per site based on the eventual theme.
@sanderlooijenga it is weird the favicon doesn't changes indeed will look into this tomorrow

alber70g (approved):

PR #1439, opened 2024-01-09
[@kadena/graph] Performance improvements and bug fixes

MRVDH:
* Changed how the confirmation depth is displayed in the graph client.
* Set the default max confirmation depth calculation to 7.
* Upped the connection limit to avoid graph and graph client startup errors.
* Applied some minor fixes to the graph documentation.
* Applied some minor typing fixes in graph and graph client.
* Moved git functions from utils to its own service directory.
* Moved the block update script to its own scripts directory.
* Deleted the blocks service and its utils functions.
* Removed the obsolete isRowNotFoundError error function.
* Removed unused api report file.
* Fixed env var being case sensitive.
* Fixed tracing scripts missing module alias.
* Replaced subscriptions to only return IDs.
* Fixed global ID resolving and parsing.
* Usage of new subscriptions on graph client.
* Connection pool limit fixes.
* Applied skip in graph client so that it does not send unnecessary requests.
* Changed favicon.


packages/apps/graph-client/src/pages/block/overview/[hash].tsx
```diff
@@ -107,12 +107,12 @@ const Block: React.FC = () => {
...
-                        configData.graphConfiguration?.maximumConfirmationDepth
-                        ? `> ${data.block.confirmationDepth - 1}`
-                        : data.block.confirmationDepth
-                      : data.block.confirmationDepth}
+                    {!configData?.graphConfiguration
+                      ?.maximumConfirmationDepth ||
+                    data.block.confirmationDepth <
+                      configData.graphConfiguration.maximumConfirmationDepth
+                      ? data.block.confirmationDepth
+                      : `>${data.block.confirmationDepth}`}
```
Comment (nil-amrutlal-dept) [inline packages/apps/graph-client/src/pages/block/overview/[hash].tsx:115]: Small and probably unimportant but: if the confirmation depth limit is set to 6 and a block's depth is 6, the UI will display >6 which is technically incorrect.

nil-amrutlal-dept:
I think this issue was not taken into account in this PR, but I stumbled across this:
image


This time is was on a cont transaction.

packages/apps/graph-client/src/pages/block/overview/[hash].tsx
Comment (MRVDH) [reply packages/apps/graph-client/src/pages/block/overview/[hash].tsx:115]: This is true, but if the max confirmation depth is set to 6, then it will never display >6 because it doesn't calculate further. There will never be a 7.

nil-amrutlal-dept (approved):

MRVDH:
> I think this issue was not taken into account in this PR, but I stumbled across this: ~snip~
> 
> This time is was on a cont transaction.

Yep I see this still happening for me as well, so will check again :/

MRVDH:
- [ ] Fix JSON errors (again).
- [ ] Show proper error on account or module not being found.

MRVDH:
> Is it normal to still get an error on the loading page? image

Nope... I will take another look 😕

PR #1447, opened 2024-01-10
[@kadena/graph] Reduce complexity of cross-chain counterpart lookup in Transfers (Feat)

nil-amrutlal-dept:
For this, I tested a couple of scenarios:

- One using Prisma's query methods and ensuring type safety; this is the one included in the PR
- The other way was by using a single query, which although it removed complexity, added some other downsides because it made use of `$queryRaw`. These downsides had to do with the manual mapping of fields, which are not automatically mapped to our defined Prisma models, and no type safety. 

I ended up doing some tracing tests as well and saw that although the current complexity is 2 (higher than the single query way), it does take less time.

**TLDR: Reduced complexity from 4 -> 2. It was possible to bring it down to 1 but it had no real improvement**

Note: also fixed a small encountered issue in cross-chain transfers (simulation)

### DIFF
--- a/.changeset/funny-ants-sort.md
+++ b/.changeset/funny-ants-sort.md
@@ -1,0 +1,6 @@
+---
+'@kadena/graph': patch
+---
+
+Increased effienciency and reduced complecity on the transfer cross-chain field
+search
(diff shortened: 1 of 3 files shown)


MRVDH (approved):
Nice!

PR #1452, opened 2024-01-11
fix(docs): streamline changes in md files

sstraatemans:


packages/apps/docs/package.json
```diff
@@ -32,6 +32,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
+    "-": "^0.0.1",
```
realdreamer, line 35: what's the use of this package and `D` both seems archived and not maintained anymore?

realdreamer (approved):

PR #1456, opened 2024-01-11
feat(client): Add signMethod to the signWithKeyPair for flexibility

javadkh2:
accept the sign method for signWithKeyPair
this let us to reuse this function for scenarios that keypair is encrypted

### DIFF
--- a/.changeset/proud-snails-compare.md
+++ b/.changeset/proud-snails-compare.md
@@ -1,0 +1,5 @@
+---
+'@kadena/client': minor
+---
+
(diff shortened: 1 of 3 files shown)


nillo (approved):

barthuijgen (approved):

alber70g:
Closed as this isn't needed anymore. We will sign hashes instead of transactions with `signFromSeed` function from `@kadena/hd-wallet`, and use `addSignatures` from `@kadena/client` to add it to transactions

PR #1461, opened 2024-01-12
Give zIndex to Tooltip component

sanderlooijenga:
---
  - https://app.asana.com/0/0/1206193208017747

### DIFF
--- a/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
+++ b/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
@@ -24,6 +24,7 @@ export const base = style([
       borderBottom: `6px solid ${tokens.kda.foundation.color.background['layer-3'].default}`,
       borderLeft: '6px solid transparent',
     },
+    zIndex: tokens.kda.foundation.zIndex.overlay,
   },
 ]);
 
--- a/.changeset/great-spiders-lay.md
+++ b/.changeset/great-spiders-lay.md
@@ -1,0 +1,5 @@
+---
+'@kadena/react-ui': patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -24,6 +24,7 @@ export const base = style([
       borderBottom: `6px solid ${tokens.kda.foundation.color.background['layer-3'].default}`,
       borderLeft: '6px solid transparent',
     },
+    zIndex: tokens.kda.foundation.zIndex.overlay,
```
eileenmguo, line 27: do you think we should add the zIndex tokens to atoms?

sanderlooijenga: I mean, yeah we could! Right now it has `[-1, 0, 1]` as options. If you think it's helpful, I will implement it in the atoms 👍

eileenmguo: I don't have a strong preference, so it's up to you 👍

eileenmguo (approved):

packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
salamaashoush: I think the values in the tokens are not generic and are specific to some components hence they should not be a utility class

sanderlooijenga: That's true, so leave them out of the atoms (at least for now)?

PR #1469, opened 2024-01-12
[@kadena-clli] Add devnet simulate command (Feat)

nil-amrutlal-dept:
In the PR, we find the simulate feature integrated in `kadena-cli` under the `devnet` command.
The simulate feature is responsible for simulating traffic in the devnet. Alongside the simulation, some auxiliary files implement the different transfer types and utils functions. 
The simulate feature also produces a log file. 

Note: some additional files are included in the PR due to lint and style fixes 

---
  - https://app.asana.com/0/0/1205886760574941


MRVDH (approved):

packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -8,66 +8,62 @@ import { FAUCET_CONSTANTS } from '../../constants/faucet.js';
...
-  [
-    globalOptions.accountName(),
-    globalOptions.amount(),
-    globalOptions.fungible(),
-    globalOptions.network(),
-    globalOptions.chainId(),
-  ],
-  async (config) => {
-    debug('account-fund:action')({ config });
+export const fundCommand: (program: Command, version: string) => void =
```
nillo, line 11: Please align ur ide :) and remove all the files that are not related to your actual changes from this PR

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
```diff
@@ -0,0 +1,55 @@
...
+  };
+  const directory = path.join(`${folder}/`);
+  createDir(directory);
+  const filepath = path.join(directory, filename);
+  const headers = `${Object.keys(exampleData).join(',')}\n`;
+  fs.writeFileSync(filepath, headers);
+  return filepath;
+}
+
+export function appendToFile(
```
nillo, line 49: @barthuijgen throughout this whole file helper you want it to make use of your service so it can use memoryfs?

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -195,6 +196,64 @@ export const globalOptions = {
       'Version of the kadena/devnet Docker image to use (e.g. "latest")',
     ),
   }),
+  //Simulate
```
nillo, line 199: please take a look at this comment, since i have the same comment here for this code: https://github.com/kadena-community/kadena.js/pull/1476/files#r1453533628

nillo (changes requested):
@alber70g same question as with other PR, i have looked at the implementation, but other then that, its usecase / usage is something i have not enough input on please review

packages/tools/kadena-cli/src/account/commands/accountFund.ts
nil-amrutlal-dept: Fixed, removed the non-related files

packages/tools/kadena-cli/src/utils/globalOptions.ts
nil-amrutlal-dept: Fixed

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
barthuijgen: Yes definitely! This does seem like code that would be good to add unit test for. 

Please make use of the `services.filesystem` implementation, it does not have appendFile yet, but you can add it to the service, send me a message if it is unclear how to do that.

alber70g (changes requested):

packages/tools/kadena-cli/README.md
```diff
@@ -79,6 +79,18 @@ brew install kda-cli
...
+To install the executable from this repo:
+
+```sh
+pnpm install
+pnpm build --filter @kadena/kadena-cli
+# if you are using NVM, you should have this environment variable available
+ln -s $(pwd)/bin/kadena-cli.js $NVM_BIN/kda
+# if not, you can replace $NVM_BIN to any path you have added in your $PATH
+```
+
```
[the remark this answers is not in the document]
nil-amrutlal-dept: Yep, so I had to do this to be able to use the `kda` command from any directory on my terminal. In my case (not using brew) I had to create a symlink to the bin file, which in my case (using NVM), is `$NVM_BIN/kda`. 
`$NVM_BIN`: environment variable which specifies my bin directory. I can rewrite it to make it more clear. 
On my case `pnpm link` and `pnpm install -g` did not work

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/crosschain-transfer.ts
```diff
@@ -0,0 +1,71 @@
...
+    );
+    gasPayer = sender00;
+  }
+
+  if (!gasPayer.keys.map((key) => key.secretKey)) {
+    console.log(
+      `Gas payer ${gasPayer.account} does not have a secret key; using sender00 as gas payer`,
+    );
+    gasPayer = sender00;
+  }
```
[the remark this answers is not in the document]
nillo: i think these validations should not be part of the options since hes comparing the result of 2 totally we do this also in other places, and it would complicate the option logic

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -195,6 +196,64 @@ export const globalOptions = {
...
+  }),
+  simulateLogFolder: createOption({
+    key: 'logFolder' as const,
+    prompt: simulatePrompts.simulateLogFolderPrompt,
+    validation: z.string(),
+    option: new Option(
+      '-l, --log-folder <logFolder>',
+      'Specify the directory where the log file will be generated.',
+    ),
+  }),
```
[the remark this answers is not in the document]
nil-amrutlal-dept: Having also this in mind: https://github.com/kadena-community/kadena.js/pull/1476#discussion_r1453533628

What should be the go-to behavior? The first option should be a one letter flag when possible (since the second already states what it is) or should we keep one letter flags for booleans and common options?

packages/tools/kadena-cli/src/account/commands/accountFund.ts
nil-amrutlal-dept: Note: these files were auto formatted using `pnpm format:src` so I am not sure if the problem is in my IDE

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
nil-amrutlal-dept: Fixed

packages/tools/kadena-cli/README.md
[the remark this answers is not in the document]
nil-amrutlal-dept: Ended up installing a different version manager that makes use of a shared bin directory. Managed to install `kadena` globally by doing `pnpm link -g`

packages/tools/kadena-cli/src/devnet/commands/devnetSimulation.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (!containerIsRunning(config.devnetConfig.name)) {
+        console.log('Devnet is not running. Please run the devnet first.');
+        return;
+      }
+
+      await simulateCoin({
+        network: {
+          id: 'fast-development',
+          host: `http://localhost:${config.devnetConfig.port}`,
+        },
```
alber70g, line 38: can we use `globalOptions.network()` here?

packages/tools/kadena-cli/src/devnet/utils/docker.ts
```diff
@@ -141,3 +141,13 @@ export function updateDevnet(version?: string): void {
...
+
+export function containerIsRunning(name: string): boolean {
+  const runningContainers = execSync(
+    'docker ps --filter ancestor=kadena/devnet --format "{{.Names}}"',
+  )
+    .toString()
+    .trim()
+    .split('\n');
+  return runningContainers.includes(name);
+}
```
alber70g, line 153: When using `ancestor=kadena/devnet`, argument `name` becomes obsolete and you can check whether the amount if lines returned by `docker ps .... ` > 0, right?

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/crosschain-transfer.ts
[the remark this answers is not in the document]
alber70g: Can we add `sender00` as a predefined global account?

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/safe-transfer.ts
```diff
@@ -0,0 +1,92 @@
...
+  sender,
+  amount,
+}: {
+  network: { host: string; id: string };
+  receiver: IAccount;
+  chainId: ChainId;
+  sender: IAccount;
+  amount: number;
+}): Promise<ICommandResult> {
+  const extraAmount = new PactNumber('0.001').toPactDecimal();
```
alber70g, line 22: Can we use the smallest amount possible? 1*10^-8

alber70g (approved):

packages/tools/kadena-cli/src/devnet/utils/docker.ts
nil-amrutlal-dept: https://github.com/kadena-community/kadena.js/pull/1469#discussion_r1457389453

packages/tools/kadena-cli/src/devnet/commands/devnetSimulation.ts
nil-amrutlal-dept: We can indeed use it. However, since I already prompt the user to choose a `devnet` I don't know if it's necessary to have both of the options. The reason for the `devnet` prompt instead of `network`, is that we can validate if the selected devnet configuration is running (by verifying the running containers against the devnet name)

packages/tools/kadena-cli/src/devnet/utils/docker.ts
nil-amrutlal-dept: The reason for verifying the name is that we can make sure the chosen devnet config is indeed the one running. This way we not only validate that a devnet is indeed running but also that the devnet configuration chosen is the correct one. This is useful because we can therefore use the port in that config without the need for an additional promt

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -195,6 +196,73 @@ export const globalOptions = {
...
+    prompt: simulationPrompts.defaultChainIdPrompt,
+    validation: z.string(),
+    transform: (chainId: string) => {
+      return chainId as ChainId;
+    },
+    option: new Option(
+      '-c, --default-chain-id <defaultChainId>',
+      'Default chain id for the simulation (eg. 0).',
+    ),
+  }),
```
alber70g, line 264: Maybe we can move the options specifically for simulation to it's own options set.

alber70g (approved):

packages/tools/kadena-cli/src/prompts/simulate.ts
alber70g, line 1: Maybe we can move this file to `./src/devnet/prompts/simulate.ts` similar to the options specific to simulation

nil-amrutlal-dept: They are already there. Just being exported from the index.ts` file

packages/tools/kadena-cli/src/utils/simulationOptions.ts
```diff
@@ -0,0 +1,66 @@
...
+      '-na, --accounts <accounts>',
+      'Amount of accounts to be created in the simulation.',
+    ),
+  }),
+  simulationTransferInterval: createOption({
+    key: 'interval' as const,
+    prompt: simulationPrompts.transferIntervalPrompt,
+    validation: z.number(),
+    option: new Option(
+      '-ti, --interval <interval>',
```
nillo, line 23:
```
simulationNumberOfAccounts: createOption({
    key: 'accounts' as const,
    prompt: simulationPrompts.numberOfAccountsPrompt,
    validation: z.number(),
    option: new Option(
      '-na, --accounts <accounts>',
      'Amount of accounts to be created in the simulation.',
    ),
  }),
```

please uphold this format:

```
simulationNumberOfAccounts: createOption({
    key: 'simulationNumberOfAccounts' as const,
    prompt: simulationPrompts.numberOfAccountsPrompt,
    validation: z.number(),
    option: new Option(
      '-na, --simulation-number-of-accounts <simulationNumberOfAccounts>',
      'Amount of accounts to be created in the simulation.',
    ),
  }),
```

For this and all prompts below

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
```diff
@@ -0,0 +1,57 @@
...
+export type TransferType =
+  | 'fund'
+  | 'transfer'
+  | 'cross-chain-transfer'
+  | 'safe-transfer'
+  | undefined;
+
+export type TokenActionType = 'create' | 'mint' | 'transfer' | undefined;
+
+export function createDir(directory: string): void {
```
nillo, line 27: move to filesystem service, so others can use this also, (if not already there)

nillo (changes requested):
lgtm except the minor change request

packages/tools/kadena-cli/src/devnet/utils/docker.ts
```diff
@@ -141,3 +141,13 @@ export function updateDevnet(version?: string): void {
   execSync(`docker pull ${image}`);
   console.log(chalk.green(`Updated ${image}`));
 }
+
+export function containerIsRunning(): boolean {
```
barthuijgen, line 145: When you use this method, are you making the assumption that the container is exposing a specific port as well? Because this method does not verify it, someone could have started a container using `kadena/devnet` but exposed it on a different port

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/simulate.ts
```diff
@@ -0,0 +1,284 @@
...
+import { safeTransfer } from './safe-transfer.js';
+import { transfer } from './transfer.js';
+
+const simulationTransferOptions: TransferType[] = [
+  'cross-chain-transfer',
+  'transfer',
+  'safe-transfer',
+];
+
+export const MARMALADE_TEMPLATE_FOLDER = 'src/devnet/contracts/marmalade-v2';
```
barthuijgen, line 28: This const isn't being used right? please remove it. (also it wouldn't work if used)

packages/tools/kadena-cli/src/devnet/utils/docker.ts
nil-amrutlal-dept: I am not using this method anymore. It was indeed not validating the port but the name before. Not it only validates if a container is running. If not useful, I can just remove the function because I am not using it

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/simulate.ts
```diff
@@ -0,0 +1,284 @@
...
+        requestKey: result.reqKey,
+        action: 'fund',
+      });
+    }
+
+    // Generate first seeded random number
+    let seededRandomNo = seedRandom(seed);
+    let counter: number = 0;
+
+    while (true) {
```
barthuijgen, line 145: There is a lot of complexity within this while loop, making it hard for me to see how exactly it will break the loop, or if it is possible that it will run forever. This code would really benefit from moving some logic to separate methods

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
barthuijgen: there is a `ensureDirectoryExists` method

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
```diff
@@ -0,0 +1,57 @@
...
+    to: '',
+    amount: 0,
+    requestKey: '',
+    action: undefined,
+  };
+  const directory = path.join(`${folder}/`);
+  createDir(directory);
+  const filepath = path.join(directory, filename);
+  const headers = `${Object.keys(exampleData).join(',')}\n`;
+  fs.writeFileSync(filepath, headers);
```
barthuijgen, line 46: please use the `service.filesystem` instead

packages/tools/kadena-cli/src/devnet/utils/docker.ts
nil-amrutlal-dept: Removed method since its not being used

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/simulate.ts
nil-amrutlal-dept: Removed. Well spotted 👍🏽

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
nil-amrutlal-dept: Fixed

packages/tools/kadena-cli/src/devnet/utils/simulation/file.ts
nil-amrutlal-dept: Fixed

packages/tools/kadena-cli/src/devnet/utils/simulation/coin/simulate.ts
nil-amrutlal-dept: Everything inside the loop is a little bit hard to refactor since I would have to pass a bunch of common variables inside each method and return objects with multiple properties due to the necessary logging in addition to some conditions that validate if the iteration is valid or if it should be skipped.

I can add some more comments to add clarity to what is happening and check if I can refactor it into smaller methods. 
Answering your question: the loop could go for infinity. This loop's goal is the simulate random traffic in the devnet (with the possibility of replicating it via the seed number). The only way to break the loop automatically is with a maximum time flag which is being developed on another PR, otherwise you'd have to shut it yourself.

barthuijgen (approved):

packages/tools/kadena-cli/src/utils/simulationOptions.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { ChainId } from '@kadena/types';
+import { simulate as simulationPrompts } from '../prompts/index.js';
+export const simulationOptions = {
+  //Simulation
+  simulationNumberOfAccounts: createOption({
+    key: 'simulationNumberOfAccounts' as const,
+    prompt: simulationPrompts.numberOfAccountsPrompt,
+    validation: z.number(),
+    option: new Option(
+      '-na, --simulation-number-of-accounts <simulationNumberOfAccounts>',
```
nillo, line 14: @alber70g there was some confusion about this
but this should be
-n or only --simulation-number-of-accounts and removal of the -na 

correct | lets get this confusion out :)

nillo (approved):
Approved but please let Albert confirm my remark.  ;)

packages/tools/kadena-cli/src/devnet/commands/devnetSimulation.ts
alber70g: Simulation should be network independent. So it can be any network that's running anywhere, not just on `localhost`.

packages/tools/kadena-cli/src/utils/simulationOptions.ts
alber70g: So since this is a specifically for devnet simulate, should we move it to the `simulate` directory?

nil-amrutlal-dept: Moved to simulation specific directory; updated the flags to single characters

PR #1476, opened 2024-01-15
[@kadena/cli] Implement max time option for simulate

nil-amrutlal-dept:
This PR contains an additional flag for max time for the simulation command


MRVDH (approved):

packages/tools/kadena-cli/src/devnet/commands/simulate.ts
```diff
@@ -16,10 +16,13 @@ export const simulateCommand: CreateCommandReturnType = createCommand(
     globalOptions.simulateTokenPool(),
     globalOptions.simulateMaxAmount(),
     globalOptions.simulateSeed(),
+    globalOptions.simulateMaxTime({ isOptional: true }),
   ],
   async (config) => {
```
nillo, line 21: With the changes made by bart in the PR we were waiting for , that was merged before you started.

you see that every command has
` async (config) => {`
- some logic for getting/passing config coming from Config
- a action function `result = await someAction(config, whatever)`
```
export const someAction = async (config, whatever): Promise<CommandResult<{}>> => {
   try {
    // something
    return { success: true, data: response };
  } catch (error) {
    return {
      success: false,
      errors: [`Error in someAction: ${error.message}`],
    };
  }
};
```

This so someAction can be tested independent

- a asserter: `assertCommandError(result)`
- some login,

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -253,6 +253,15 @@ export const globalOptions = {
       'Specify the directory where the log file will be generated.',
     ),
   }),
+  simulateMaxTime: createOption({
```
nillo, line 256:
```
simulateMaxTime: createOption({
    key: 'simulateMaxTime' as const,
    prompt: simulatePrompts.simulateMaxTimePrompt,
    validation: z.number().optional(),
    option: new Option(
      '-m, --simulate-max-time <simulateMaxTime>',
      'Specify the maximum time in miliseconds the simulation will run.',
    ),
  }),
```

the name op the options, should match troughout the whole option>
also the first option should be a single character)

if you say -xx whatever within globaloptions this is wrong, and will be solved.
-m instead of -mt since --simulate-max-time already states what it is

nillo (changes requested):
- Please also remove the .gitkeep file from the PR
- I didnt take a dive into the actual functionality of what simulate is doing, @alber70g you know what this is for, maybe you can take a look at this PR also

packages/tools/kadena-cli/src/devnet/commands/simulate.ts
nillo: hmmm on second thought  since there is no actual result this might not actually be needed, there is nothing to assert if its non-existent :P

barthuijgen: it can still have errors right? Like `guardDocker` can throw an error, you want to have a warning if container is not running. That could still use the warnings/errors messages of the command-result structure, even if no data is returned

nil-amrutlal-dept: Yes or even on the simulation itself, we can have errors. I will implement some error handling

alber70g (approved):

nillo (approved):

PR #1481, opened 2024-01-16
[@tools-website] Update utility classes

KristinaSpasevska:
Update all new classes and use atoms in Tools website.
- use atoms and new tokens in stylesheets in pages, global components and general layout components;
- use new props in imported components from react-ui library when applicable


packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -1,66 +1,74 @@
...
+    fontSize: 'sm',
     zIndex: 1,
   }),
   {
+    width: tokens.kda.foundation.size.n12,
+    right: tokens.kda.foundation.size.n12,
+    top: tokens.kda.foundation.size.n16,
     height: '100vh',
-    borderLeft: `solid 1px ${vars.colors.$borderSubtle}`,
+    backgroundColor: tokens.kda.foundation.color.neutral.n1,
```
eileenmguo, line 18: can we use one of the utility classes for this? like "base.default"

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -1,66 +1,74 @@
...
+    borderLeft: `solid 1px ${tokens.kda.foundation.color.border.base}`,
```
eileenmguo, line 19: you can use tokens.kda.foundation.border.width.hairline I think instead of 1px for consistency

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -1,66 +1,74 @@
...
+  atoms({
+    height: '100%',
     flexDirection: 'column',
     position: 'relative',
-  },
+  }),
   {
-    height: '100%',
-    borderLeft: `solid 1px ${vars.colors.$borderSubtle}`,
+    borderLeft: `1px solid #E3E1E5`,
```
eileenmguo, line 36: I don't think we should ever use hardcoded color values. Can this be replaced with one of the tokens?

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -1,66 +1,74 @@
...
-  sprinkles({
+  atoms({
     outline: 'none',
     background: 'none',
     border: 'none',
     cursor: 'pointer',
   }),
   {
-    borderBottom: `solid 1px ${vars.colors.$borderSubtle}`,
+    borderBottom: `1px solid #E3E1E5`,
```
eileenmguo, line 57: same here, the tokens should already apply the colors based on the theme

packages/apps/tools/src/components/Common/Layout/partials/Header/styles.css.ts
```diff
@@ -1,29 +1,29 @@
...
+    color: tokens.kda.foundation.color.neutral.n50,
+    padding: tokens.kda.foundation.size.n3,
     selectors: {
       [`${darkThemeClass} &:hover`]: {
-        color: vars.colors.$white,
+        color: tokens.kda.foundation.color.neutral.n100,
       },
       [`&:hover`]: {
-        color: vars.colors.$white,
+        color: tokens.kda.foundation.color.neutral.n0,
```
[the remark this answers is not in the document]
eileenmguo: yeah, that's how it was before, but now they change depending on the theme. I see this is in the header so what you can do is apply the dark class and it will always stay in dark theme. If this needs to happen more often, we can try to find a different solution

packages/apps/tools/src/components/Global/ResourceLinks/styles.css.ts
```diff
@@ -1,47 +1,52 @@
...
-  sprinkles({
-    marginTop: '$4',
-    paddingTop: '$4',
+  atoms({
+    marginBlockStart: 'md',
+    paddingBlockStart: 'md',
   }),
   {
-    borderTop: `1px solid ${vars.colors.$borderSubtle}`,
+    borderTop: `1px solid #E3E1E5`,
```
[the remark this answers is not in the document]
eileenmguo: I think you should be able to use the border tokens for these

packages/apps/tools/src/components/Global/ResourceLinks/styles.css.ts
```diff
@@ -1,47 +1,52 @@
...
-  sprinkles({
-    paddingLeft: '$4',
-    marginTop: '$2',
-    color: '$blue80',
+  atoms({
+    paddingInlineStart: 'md',
+    marginBlockStart: 'sm',
   }),
   {
+    color: tokens.kda.foundation.color.palette.blue.n80,
```
eileenmguo, line 39: you should use the semantic color. In this case maybe you even want to use the `Link` component from react-ui?

eileenmguo (commented):
looking good so far!

packages/apps/tools/src/components/Common/Layout/partials/Header/styles.css.ts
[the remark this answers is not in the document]
KristinaSpasevska: There already is such selector so that it always stays in dark theme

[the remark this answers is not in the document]
eileenmguo: if that's the case you shouldn't need to use a selector here to set both since the darkThemeClass will always be applied

eileenmguo (approved):

PR #1488, opened 2024-01-17
[client-utils]feat: add template parsing from string to yaml converter

barthuijgen:
Add the `createPactCommandFromStringTemplate` and `getPartsAndHoles` method to `@kadena/client-utils/nodejs`

This allows finding holes in templates to data can be collected, and subsequently parsing the template with data into a transaction directly from strings, removing the need for filesystem usage.

### DIFF
--- a/packages/libs/client-utils/src/nodejs/yaml-converter.ts
+++ b/packages/libs/client-utils/src/nodejs/yaml-converter.ts
@@ -165,7 +165,8 @@ export const convertTemplateTxToPactCommand = (
     },
   };
 
-  const { publicMeta, meta, ...kdaToolTxWithoutMeta } = kdaToolTx;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { publicMeta, meta, code, ...kdaToolTxWithoutMeta } = kdaToolTx;
   const metadata = meta || publicMeta || ({} as IPublicMeta);
 
   return {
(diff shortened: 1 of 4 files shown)


packages/libs/client-utils/src/nodejs/yaml-converter.ts
```diff
@@ -191,6 +195,18 @@ export const createPactCommandFromTemplate = async (
   )(path, cwd);
 };
 
+export const createPactCommandFromStringTemplate = async (
```
javadkh2, line 198: Nice function

packages/libs/client-utils/src/nodejs/yaml-converter.ts
```diff
@@ -85,91 +87,93 @@ export const getPartsAndHolesInCtx = (
...
+      })
+      .join('');
+  };
+
+const loadYaml = (filledYamlString: string) => {
+  return yaml.load(filledYamlString) as ITemplateTransaction;
 };
 
 // Responsible for replacing holes in a context with values
 export const replaceHolesInCtx = (args: Record<string, string | number>) => {
```
javadkh2, line 125: @alber70g @nil-amrutlal-dept 
Do we really need this function? Why not just using `replaceHoles` directly?
also in the `createPactCommandFromTemplate` you called this function but actually only the filledYamlString is used.

javadkh2 (approved):

packages/libs/client-utils/src/nodejs/yaml-converter.ts
nil-amrutlal-dept: I will have a look at this when I have some time. Added an issue on the board.
Need to recheck the flow to see if I need it

barthuijgen: I think it mostly has to do with the functional pipe, the next method passed to is `parseYamlToKdaTx` and it expects the entire ctx object, for the new method I didn't need the entire ctx since it won't try to load any files.

PR #1494, opened 2024-01-17
[@kadena-cli]feat(account): add account manual and from wallet sub commands

realdreamer:
https://app.asana.com/0/1205969628270714/1205991424580865/f

`account add-manual` 
`account add-from-wallet` commands


packages/tools/kadena-cli/src/account/commands/accountAddWallet.ts
```diff
@@ -0,0 +1,74 @@
...
+    const publicKeysList: Array<string | undefined> = [];
+    for (const key of keyWalletConfig.keys) {
+      const content = await services.filesystem.readFile(
+        path.join(WALLET_DIR, keyWalletConfig?.folder, key),
+      );
+      const parsed = content !== null ? (yaml.load(content) as IKeyPair) : null;
+      publicKeysList.push(parsed?.publicKey);
+    }
+
+    publicKeysList.filter((key) => !!key);
```
barthuijgen, line 55: This line doesn't do anything, `.filter()` doesn't mutate

packages/tools/kadena-cli/src/account/commands/accountAddWallet.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const selectPublicKeys = await checkbox({
+      message: 'Select public keys to add to account',
+      choices: publicKeysList.map((key) => ({ value: key })),
+    });
+
+    (config as unknown as IAddAccountManualConfig).publicKeysConfig =
```
barthuijgen, line 62: I would advice not to mutate `config`, just define a new variable instead, if you want the rest of config just do 
```
const whatever = {...config, publicKeysConfig: selectedPublicKeys }
```

packages/tools/kadena-cli/src/account/commands/accountAddWallet.ts
realdreamer: yup! true blunder 👍🏽 .

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,42 @@
...
+    globalOptions.publicKeys(),
+    globalOptions.predicate(),
+  ],
+
+  async function addAccount(config): Promise<void> {
+    debug('account-add-manual:action')({ config });
+
+    const sanitizedAlias = sanitizeFilename(config.accountAlias).toLowerCase();
+    const filePath = path.join(defaultAccountPath, `${sanitizedAlias}.yaml`);
+
```
barthuijgen, line 35: keep prompts in this action

packages/tools/kadena-cli/src/account/utils/addHelpers.ts
```diff
@@ -0,0 +1,251 @@
...
+    process.exit(1);
+  }
+
+  await writeAlias(config, filePath);
+  console.log(
+    chalk.green(
+      `\nThe account configuration "${config.accountAlias}" has been saved.\n`,
+    ),
+  );
+}
```
barthuijgen, line 251: move console log to the commander action

packages/tools/kadena-cli/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./node_modules/@kadena-dev/shared-config/tsconfig-base.json",
   "compilerOptions": {
     "module": "NodeNext",
-    "types": [".kadena/pactjs-generated"]
-  }
+    "types": [".kadena/pactjs-generated", "vitest/globals"]
```
barthuijgen, line 5: please remove this and manually import

packages/tools/kadena-cli/src/account/utils/__tests__/addHelpers.test.ts
```diff
@@ -0,0 +1,358 @@
...
+  });
+
+  it('should return undefined when predicate is not provided', async () => {
+    const accountName = await createAccountName({
+      ...defaultConfig,
+      predicate: undefined as unknown as Predicate,
+      publicKeysConfig: ['publicKey1', 'publicKey2'],
+    });
+    expect(accountName).toBeUndefined();
+  });
```
barthuijgen, line 56: this test should not be needed. trust typescript, and if input comes from the user use zod as early as possible to validate it

packages/tools/kadena-cli/src/account/utils/__tests__/addHelpers.test.ts
```diff
@@ -0,0 +1,358 @@
...
+    };
+    const filePath = path.join(
+      root,
+      defaultAccountPath,
+      `${config.accountAlias}.yaml`,
+    );
+    const fs = services.filesystem;
+    // To start fresh delete the file if it exists
+    // eslint-disable-next-line no-unused-expressions
+    (await fs.fileExists(filePath)) && (await fs.deleteFile(filePath));
```
barthuijgen, line 291: just use an if statement

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,74 @@
...
+    globalOptions.accountAlias(),
+    globalOptions.accountName(),
+    globalOptions.fungible(),
+    globalOptions.network(),
+    globalOptions.chainId(),
+    globalOptions.publicKeys(),
+    globalOptions.predicate(),
+  ],
+
+  async function addAccount(config): Promise<void> {
```
barthuijgen, line 30: I think this method is still doing too much, please add a test that does add-manual and add-wallet and see how easy it is to re-use all the logic.

packages/tools/kadena-cli/src/account/utils/addHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+      ...config,
+      publicKeys: accountDetails.publicKeys.join(','),
+      publicKeysConfig: accountDetails.publicKeys,
+      predicate: accountDetails.predicate,
+    };
+    return updatedConfig;
+  }
+};
+
+export const getFilePath = (fileName: string): string => {
```
barthuijgen, line 33: Please use more descriptive names, `getFilePath` is quite vague, but it is for account only, especially for methods are are exported try to have clear naming that reflects its purpose. `getAccountFilePath` would already be better. Also take a look over the PR yourself to see if you can find other methods that could use clarification of their names

packages/tools/kadena-cli/src/account/utils/addAccount.ts
```diff
@@ -0,0 +1,29 @@
+import type { CommandResult } from '../../utils/command.util.js';
+import type { IAddAccountManualConfig } from '../types.js';
+import { getFilePath, getUpdatedConfig } from './addHelpers.js';
+import { validateAccountDetails } from './validateAccountDetails.js';
+import { writeConfigInFile } from './writeConfigInFile.js';
+
+export async function addAccount(
+  config: IAddAccountManualConfig,
+  callback: () => Promise<boolean>,
```
barthuijgen, line 9: Isn't this missing something? When I define a callback to decide between local and on-chain data, it would help to actually have access to the local and on-chain data to compare.

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,41 @@
...
+  program: Command,
+  version: string,
+) => void = createCommand(
+  'add-manual',
+  'Add an existing account to the CLI',
+  [
+    globalOptions.accountAlias(),
+    globalOptions.accountName(),
+    globalOptions.fungible(),
+    globalOptions.network(),
```
barthuijgen, line 23: network can be missing, there should be a `ensureNetworkExists` method somewhere to check for this

packages/tools/kadena-cli/src/account/utils/validateAccountDetails.ts
```diff
@@ -0,0 +1,61 @@
...
+    return {
+      success: true,
+      data: {
+        config: configWithAccountName,
+        accountDetails,
+        isConfigAreSame,
+      },
+    };
+  } catch (error) {
+    if (error.message.includes('row not found') === true) {
```
barthuijgen, line 48: if you add an account that does not exist on chain, you should still write the config to a file and remember the account, but maybe inform the user that it does not exist on-chain.

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,41 @@
...
+    globalOptions.chainId(),
+    globalOptions.publicKeys(),
```
barthuijgen, line 25: if there are multiple public keys, it has to use `(create-principal (read-keyset "ks"))` to create the account name as described in the story, this seems to be missing.

packages/tools/kadena-cli/src/account/utils/validateAccountDetails.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export interface IValidateAccountDetailsData {
+  isConfigAreSame: boolean;
+  config: IAddAccountManualConfig;
+  accountDetails: IAccountDetailsResult;
+}
+
+export async function validateAccountDetails(
+  config: IAddAccountManualConfig,
+): Promise<CommandResult<IValidateAccountDetailsData>> {
```
barthuijgen, line 18: `CommandResult` was only needed for the main logic method called from the cli command, not other helpers and utils. Making error messages should then also be moved up to the `addAccount` method

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,41 @@
...
+    globalOptions.predicate(),
+  ],
+
+  async (config): Promise<void> => {
+    try {
```
barthuijgen, line 30: `try {} catch(){}` is not needed, you may have copied it from other commands but we need to remove it there too, `createCommand` already handles this

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -50,6 +50,16 @@ export const globalFlags = {
...
+  accountAlias: createOption({
+    key: 'accountAlias' as const,
+    defaultIsOptional: false,
+    prompt: account.accountAliasPrompt,
+    validation: z.string(),
+    option: new Option(
+      '-a, --account-alias <accountAlias>',
+      'Enter an alias to store your account',
+    ),
+  }),
```
alber70g, line 62: I wouldn't use the `-a` flag for `accountAlias` and reserve it for `account`

alber70g (approved):

packages/tools/kadena-cli/src/utils/globalOptions.ts
realdreamer: As part of cleaning up global options we thought of look into this deeper because currently few other options also use `-a` at the moment.

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -0,0 +1,115 @@
...
+    const accountDetailsFromChain = accountName
+      ? await getAccountDetailsForAddAccount({
+          accountName,
+          networkId: networkConfig.networkId,
+          chainId,
+          networkHost: networkConfig.networkHost,
+        })
+      : undefined;
+
+    const isEmptyAccountDetails =
```
barthuijgen, line 46: This name is misleading, doesn't it mean there is an account? "Is empty" makes me think if it's true there is no account

packages/tools/kadena-cli/src/account/types.ts
```diff
@@ -0,0 +1,29 @@
+import type { ChainId } from '@kadena/types';
+import type { INetworkCreateOptions } from '../networks/utils/networkHelpers.js';
+
+export type Predicate = 'keys-all' | 'keys-2' | 'keys-any';
```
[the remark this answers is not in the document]
realdreamer: yes, it can be `string`. But this type currently used for internal type casting because `createPrincipal` method from client utils library currently supports only this `Predicate` type as part of `pred` key. 

I had a chat with @javadkh2 and created a ticket to fix this type https://app.asana.com/0/1205553603119741/1206522684909673/f

barthuijgen (approved):

PR #1495, opened 2024-01-18
[@kadena/graph] Performance improvements for Prisma queries

MRVDH:
With this change, Prisma only queries the fields that were requested from the GraphQL query.

With the help of @hayes: https://github.com/hayes/pothos/issues/1131

### DIFF
--- a/.changeset/strange-birds-buy.md
+++ b/.changeset/strange-birds-buy.md
@@ -1,0 +1,5 @@
+---
+'@kadena/graph': patch
+---
+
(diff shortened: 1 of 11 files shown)


nil-amrutlal-dept:
It is indeed a bit weird that we need to individually select the fields in each object, but I saw the reply on https://github.com/hayes/pothos/issues/1131. 

Everything looks good 🙌🏽

nil-amrutlal-dept (approved):

PR #1497, opened 2024-01-18
[@kadena/graph] Remove old styles and colours

MRVDH:


packages/apps/graph-client/src/context/chain-tree-context.tsx
nil-amrutlal-dept, line 1: Maybe it's important to keep this file. This chain tree context was responsible for updating the confirmation depth on the client side. This is needed because when we only fetch a given block once and when we do so we get the current confirmation depth. Once that black has a new descendant, we need to update the old confirmation depth value.

packages/apps/graph-client/src/context/chain-tree-context.tsx
```diff
@@ -1,82 +0,0 @@
...
-  );
-};
-
-export function useChainTree(): IChainTreeContext {
-  const context = useContext(ChainTreeContext);
-  if (!context) {
-    throw new Error('useChainTree must be used within a ChainTreeProvider');
-  }
-  return context;
-}
```
nil-amrutlal-dept, line 82: The context provider could be removed indeed

packages/apps/graph-client/src/context/chain-tree-context.tsx
```diff
@@ -1,82 +0,0 @@
...
-    } else {
-      break;
-    }
-
-    //set the updated block in the chain
-    chainTree[currentBlock.chainId][currentBlock.hash] = currentBlock;
-  }
-
-  return chainTree;
-}
```
nil-amrutlal-dept, line 56: This is the function that I mentioned. We could create an in-file IBlock definition, so we don't have dependencies

packages/apps/graph-client/src/context/chain-tree-context.tsx
MRVDH: This is the new temporary view, until we have a design for a new type of explorer. So while we could keep the code, I think for now it's not necessary and I'd rather reinstate the code once we have implemented the new designs. Let me know if you agree

chrome_PdJb3N7ac5

nil-amrutlal-dept: I don't think we're updating the confirmation depth on the new view, correct? 
But yes, I agree that if the idea is to leave it like this for now, we could delete it and when we have the new view reinstate it

MRVDH: Correct, I purposely didn't do that because the list grows so quickly. That's also why I added the pause button

nil-amrutlal-dept (approved):

PR #1503, opened 2024-01-19
WIP - feat/kadena-cli/tx/sign

nillo:

### DIFF
--- a/packages/tools/kadena-cli/src/utils/command.util.ts
+++ b/packages/tools/kadena-cli/src/utils/command.util.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import type { Ora } from 'ora';
 
 export class CommandError extends Error {
-  public warnings?: string[];
+  public warnings: string[];
   public errors: string[];
   public exitCode: number = 0;
   public args?: Record<string, unknown>;
@@ -13,14 +13,14 @@ export class CommandError extends Error {
     exitCode,
     args,
   }: {
-    errors: string[];
+    errors?: string[];
     warnings?: string[];
     exitCode?: number;
     args?: Record<string, unknown>;
   }) {
-    super(`${warnings?.join('\n')}${errors.join('\n')}`);
-    this.errors = errors;
-    this.warnings = warnings;
+    super(`Error:\n${warnings?.join('\n')}${errors?.join('\n')}`);
+    this.errors = errors ?? [];
+    this.warnings = warnings ?? [];
     this.exitCode = exitCode ?? 0;
     this.args = args;
   }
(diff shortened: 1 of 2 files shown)


packages/tools/kadena-cli/src/tx/commands/txCreateTransaction.ts
```diff
@@ -66,40 +66,45 @@ export const createTransactionCommandNew = createCommandFlexible(
...
+      console.log(result.data.transaction);
 
-    const relativePath = path.relative(process.cwd(), result.data.filePath);
-    console.log(`\ntransaction saved to: ./${relativePath}`);
+      const relativePath = path.relative(process.cwd(), result.data.filePath);
+      console.log(`\ntransaction saved to: ./${relativePath}`);
 
-    return { outFile: relativePath };
+      return { outFile: relativePath };
+    } catch (error) {
```
barthuijgen, line 105: we should probably just do this try{}catch{} in the createCommand right? if its such a generic handling

KdWarlordz: would be cleaner yeah,

packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -0,0 +1,341 @@
...
+ * @param {boolean} signed - Whether to retrieve signed or unsigned transactions.
+ * @returns {Promise<string[]>} A promise that resolves to an array of transaction file names.
+ * @throws Throws an error if reading the transaction directory fails.
+ */
+export async function getTransactions(
+  signed: boolean,
+  path: string,
+): Promise<string[]> {
+  try {
+    const filePath = path ? `${process.cwd()}/${path}` : TRANSACTION_PATH;
```
barthuijgen, line 67: Prefer using path.join

packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -0,0 +1,341 @@
...
+        : join(TRANSACTION_PATH, transactionFile);
+    const transactionFilePath = join(TRANSACTION_PATH, transactionFile);
+    const fileContent = await services.filesystem.readFile(filePath);
+
+    if (fileContent === null) {
+      throw Error(`Failed to read file at path: ${transactionFilePath}`);
+    }
+    const transaction = JSON.parse(fileContent);
+    if (signed) {
+      tx.ICommandSchema.parse(transaction);
```
barthuijgen, line 248: Parse should also return the data with correct type so you don't need to use "as x"

barthuijgen (approved):

PR #1506, opened 2024-01-19
chore(@kadena/react-ui): Cleanup NavHeader and incorporate into the Docs website

eileenmguo:


packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
```diff
@@ -73,47 +66,44 @@ const Header: FC<IHeaderProps> = () => {
   };
 
   return (
-    <NavHeader.Root brand={isMediumScreen ? 'DevTools' : 'Kadena'}>
-      <NavHeader.Navigation activeHref={pathname}>
+    <NavHeader
+      logo={
+        <Link href={'/'}>
+          <Logo />
```
KristinaSpasevska, line 72: We used to have two logos for smaller/larger screens, so this isn't the case anymore?

eileenmguo:
> Also to double check, will the glow be removed from the Header?

Yes, the new branding does not have any glow

packages/apps/tools/src/components/Common/Layout/partials/Header/Header.tsx
eileenmguo: That was unintentional! I will fix it 👍

realdreamer (approved):
LGTM 👍🏽

sstraatemans (approved):

PR #1509, opened 2024-01-19
Aria and tokes complete form fields

salamaashoush:
1. Implement TextFiled with react-aria and new tokens.
2. Reimplement TextareaField with react-aria and new tokens.
3. Remove Input and Textarea (use TextField, and TextareaField without label).
4. Remove FormWrapper
5. Refactor all form styles.
6. Implement Select with react-aria and new token a (note Select is the full field with the label, but ofc like another field it could be used without a label) 
7. Implement Combobox with react-aria and new token.
8. Add some small utils

NOTES:
All form fields follow react-aria props and API I have added lots of stories to cover possible use cases but if you do not know how to use the fields we have free docs 
- https://react-spectrum.adobe.com/react-aria/useTextField.html -> TextField and TextareaField
- https://react-spectrum.adobe.com/react-aria/useSelect.html-> Select
- https://react-spectrum.adobe.com/react-aria/useComboBox.html -> Combobox


packages/apps/tools/src/components/Global/AccountNameField/index.tsx
```diff
@@ -1,68 +1,43 @@
...
   const { selectedAccount, setSelectedAccount } = useWalletConnectClient();
 
-  const lookup = {
-    input: (
-      <Input
-        {...restInputProps}
+  return (
+    <div className={accountInputWrapperStyle}>
+      <TextField
+        validationBehavior="aria"
```
eileenmguo, line 29: should we make validationBehavior="aria" by default?

packages/libs/react-ui/src/components/Form/Select/Select.stories.tsx
```diff
@@ -14,66 +19,215 @@ const meta: Meta<ISelectProps> = {
...
-        disabled={Boolean(disabled)}
-        outlined={Boolean(outlined)}
-        value={value}
       >
-        <option value={'1'}>option 1</option>
-        <option value={'2'}>option 2</option>
-      </Select>
+        <p>
+          Keep in mind that native validation is only triggered when the a form
+          is submitted. for realtime validation use the them `isInvalid` prop.
```
eileenmguo, line 211:
```suggestion
          is submitted. for realtime validation use the `isInvalid` prop.
```

packages/libs/react-ui/src/components/Form/TextField/TextField.css.ts
```diff
@@ -0,0 +1,70 @@
...
+    outline: 'none',
+    flex: 1,
+    fontSize: 'base',
+    overflow: 'hidden',
+    lineHeight: 'lg',
+  }),
+  {
+    paddingInlineStart: token('spacing.md'),
+    paddingInlineEnd: token('spacing.md'),
+    paddingBlock: token('spacing.sm'),
```
eileenmguo, line 34: I feel like a lot of these styles are duplicates. If so, can we pull them into a shared class so that any changes will automatically be applied to all of the form components?

eileenmguo (commented):
I noticed the ComboBox doesn't allow you to close the dropdown unless you click one of the options where as with the Select you can click outside of the dropdown to close the options. Can we apply the same behavior to the ComboBox?

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
salamaashoush: I it is already the default, just added this here from a habit :D

packages/libs/react-ui/src/components/Form/TextField/TextField.css.ts
salamaashoush: I see the same but I kept them separate to avoid unintentionally breaking other stuff, maybe later we can dry those a bit

salamaashoush:
> I noticed the ComboBox doesn't allow you to close the dropdown unless you click one of the options where as with the Select you can click outside of the dropdown to close the options. Can we apply the same behavior to the ComboBox?

good catch, apparently it was a bigger issue, the popover was not implemented correctly, now it is fixed

PR #1512, opened 2024-01-22
Aria and tokes complete form fields

salamaashoush:
1. Implement TextFiled with react-aria and new tokens.
2. Reimplement TextareaField with react-aria and new tokens.
3. Remove Input and Textarea (use TextField, and TextareaField without label).
4. Remove FormWrapper
5. Refactor all form styles.
6. Implement Select with react-aria and new token a (note Select is the full field with the label, but ofc like another field it could be used without a label) 
7. Implement Combobox with react-aria and new token.
8. Add some small utils

NOTES:
All form fields follow react-aria props and API I have added lots of stories to cover possible use cases but if you do not know how to use the fields we have free docs 
- https://react-spectrum.adobe.com/react-aria/useTextField.html -> TextField and TextareaField
- https://react-spectrum.adobe.com/react-aria/useSelect.html-> Select
- https://react-spectrum.adobe.com/react-aria/useComboBox.html -> Combobox


MRVDH (approved):
Minor fix pushed and approved for the graph-client!

packages/libs/react-ui/src/utils/useAnimation.ts
```diff
@@ -0,0 +1,113 @@
+import { useLayoutEffect } from '@react-aria/utils';
+import type { RefObject } from 'react';
+import { useCallback, useRef, useState } from 'react';
+import { flushSync } from 'react-dom';
+
+export function useEnterAnimation(
```
eileenmguo, line 6: can you add more documentation on the purpose of these utilities?

eileenmguo (approved):

packages/libs/react-ui/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "declarationDir": "types",
     "outDir": "dist/esm",
     "allowJs": true,
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
```
eileenmguo, line 12: What was the reason for this change?

packages/libs/react-ui/src/utils/useAnimation.ts
salamaashoush: those utils are ported from react-aria, they are used to check if the element has an animation and if it is ended/exiting or started/entering, I followed the implementation of react-aria-component, so if we need to add animation we can style using `[data-entring]` and `[data-exsiting]` for example currently the Popover support this pattern, but this is optional and we can get rid of it if not needed and since I wasn't sure if I need to implement those or not I implemented the behavior anyway, I will add js doc for those utils

packages/libs/react-ui/tsconfig.json
salamaashoush: `node` resolution doesn't support modern `exports` fields in `package.json` and react-ui uses those from crackle, so I was getting errors can't resolve `@kendena/react-ui/styles`, so I had to switch to a modern module resolution like `node16` or `bundle` and since bundle is the recommend way with bundlers I used it, in all the apps and react-ui as well

packages/apps/docs/src/scripts/utils.ts
```diff
@@ -1,12 +1,12 @@
-import type { Content, Root } from 'mdast-util-from-markdown/lib';
+function hasChildren(content: any): content is {
+  children: any[];
+} {
+  return 'children' in content;
+}
 
-export const getTypes = <T>(
```
sstraatemans, line 3: what happened with all the any?

salamaashoush: this is an internal import which broke on `moduleResolution` bundler mode, and we shouldn't use them anyway, those types are re-exports from the main package `mdast`

salamaashoush: and btw this is still typesafe I wrote a guard for that reason, the thing is `Root` and `Content` don't overlap so probably they are some other  type we should use here

sstraatemans (approved):

packages/apps/docs/src/scripts/utils.ts
salamaashoush: found a better solution there are no any

KristinaSpasevska (dismissed):

packages/apps/tools/src/pages/transactions/cross-chain-transfer-tracker/index.tsx
```diff
@@ -229,12 +226,11 @@ const CrossChainTransferTracker: FC = () => {
             <Grid>
               <GridItem>
                 <RequestKeyField
-                  helperText={inputError || undefined}
-                  status={validRequestKey}
+                  errorMessage={inputError || errors.requestKey?.message}
+                  isInvalid={validRequestKey}
```
sanderlooijenga, line 230: Can we rename `validRequestKey` to `invalidRequestKey` or make it `isInvalid={!validRequestKey}`. Now it doesn't look logical

realdreamer (approved):

packages/apps/tools/src/pages/faucet/new/index.tsx
```diff
@@ -348,17 +341,17 @@ const NewAccountFaucetPage: FC = () => {
             <div className={inputContainerClass}>
               <div className={accountNameContainerClass}>
                 <AccountNameField
-                  inputProps={register('name')}
-                  error={errors.name}
+                  {...register('name')}
+                  isInvalid={!!errors.name}
```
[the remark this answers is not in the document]
salamaashoush: those are really spesfic refactoring, I just had to make a minimal change to get the stuff working, we can refactor all of these later

packages/apps/tools/src/components/Global/AccountNameField/index.tsx
```diff
@@ -1,68 +1,43 @@
...
-  helperText,
-  error,
-  ...rest
-}) => {
-  const elementId = 'kd-select-account-input';
-  const mode = 'input';
+interface IAccountNameFieldProps
+  extends Omit<
+    ITextFieldProps,
+    'errorMessage' | 'validationBehavior' | 'onValueChange'
```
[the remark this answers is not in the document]
salamaashoush: those are really spesfic refactoring, I just had to make a minimal change to get the stuff working, we can refactor all of these later

packages/apps/tools/src/components/Global/PredKeysSelect/index.tsx
```diff
@@ -11,40 +10,41 @@ export type OnPredSelectChange = (value: PredKey) => void;
...
+  selectedKey?: PredKey;
+}
+
+const PredKeysSelect: FC<IPredKeysSelectProps> = ({
+  selectedKey,
+  onSelectionChange,
+  ...rest
+}) => {
+  const onSelectChange = useCallback(
+    (key: string | number) => {
```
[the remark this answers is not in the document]
salamaashoush: this is not the correct type react-aria provide the key as either number or string

packages/apps/tools/src/pages/transactions/cross-chain-transfer-tracker/index.tsx
salamaashoush: those are really spesfic refactoring, I just had to make a minimal change to get the stuff working, we can refactor all of these later

sanderlooijenga (approved):

PR #1514, opened 2024-01-23
[@kadena/graph] Add documentation xchain lookup (Fix)

nil-amrutlal-dept:
n+1 problem mitigated since we're making a single query to fetch transactions that match certain criteria, and we're including the related transfers in that same query. This way we fetch all the data you need in a single round trip to the database.

---
  - https://app.asana.com/0/0/1206361522212871


packages/apps/graph/src/graph/objects/transfer.ts
```diff
@@ -91,6 +91,21 @@ export default builder.prismaNode('Transfer', {
         } catch (error) {
           throw normalizeError(error);
         }
+
+        /* Note: Multiples options were tested to find the xchain counterpart including using a single raw query.
```
MRVDH, line 95:
```suggestion
        /* Note: Multiple options were tested to find the cross chain counterpart, including using a single raw query.
```

MRVDH (approved):

alber70g (approved):

PR #1539, opened 2024-01-25
fix(@kadena/react-ui): Make `Tooltip` component more robust

eileenmguo:
Fixes the bug where the tooltip does not show up on hover when used with the `Button` component


salamaashoush (approved):

MRVDH:
Tests seem to fail. I've given it a try to fix it but `screen.queryByText(messageText)` returns null, like the tooltip is not rendering for some reason

PR #1544, opened 2024-01-29
Prepare a rough documentation framework for client-utils

lsgunnlsgunn:
@alber70g @javadkh2 
In poking around in https://github.com/kadena-community/kadena.js/tree/main/packages/libs/client-utils, I find I'm a little mystified about what exactly needs to be documented. It looks like there's more to it than these:

- asyncPipe
- submitClient
- preflightClient
- dirtyReadClient
- crossChainClient

I don't see much information to guide me (sorry, I'm not really a programmer so I need a little hand-holding to flesh out descriptions).

In following up on the idea of using AI, would it be possible to run some kind of bot over the `src` directory and extract information from the code in a format similar to what I threw together in this PR (name, description, usage, parameters, return value, examples, errors)?

What I'm hoping to get is a rough draft something along the lines of what you see in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference documentation. But I'm not sure what we could use to generate something like that using AI. Thoughts?

---
  - https://app.asana.com/0/0/1206362614074808

### DIFF
--- a/packages/apps/docs/src/docs/reference/kadena-client/client-utils.md
+++ b/packages/apps/docs/src/docs/reference/kadena-client/client-utils.md
@@ -103,9 +103,14 @@ The Kadena `client-utils` package includes the following `core` helper functions
 
 ### asyncPipe
 
-Use `asyncPipe` to ...
+Use `asyncPipe` to open a connection to the Kadena network and execute multiple functions in order to return a single result.
+This utility enables you to pass the output from the first function as input to a second function and return the combined results.
 
 #### Basic usage
+
+```typescript
+asyncPipe( function(), function())
+```
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


packages/apps/docs/src/docs/reference/kadena-client/client-utils.md
```diff
@@ -0,0 +1,273 @@
...
+For example, you can import specific `coin` module functions into a TypeScript program with the following statement:
+
+```typescript
+import { getBalance, transferCrossChain } from "@kadena/client-utils/coin"
+```
+  
+The library also exports `core` utilities that you can use to develop your own API functions for any type kind of smart contract.
+
+The client utilities are listed alphabetically in this reference to make they easy to find by name.
+However, you typically call the utilities in the following order to represent the normal workflow:
```
javadkh2, line 25: Actually its not correct. there is no order and each function works independently.
`asyncPipe` is a general functional programming helper that let you to pipe (compose) some functions together. 
I already used it to create other helpers (`submitClient`, `preflightClient`, `dirtyReadClient`, `dirtyReadClient`)
the use case of asyncPipe is creating new utilities for example I am working on `queryAllChainsClient` that let you execute a command on all chains in parallel.
Regarding xClient (e.g. submitClient), each function usually manages the whole workflow from singing to submitting and listening for the result of a command. so a very high level API is like this
`config->command->result`

javadkh2:
I like the structure 
However the client-utils is a multi purpose package. I think its good to split the document to different sections
like how we export the modules. 
I suggest to have 5 titles match with exports in top level and then for each section we use the document structure.
```package.json
"exports": {
    "./built-in": "./lib/built-in/index.js",
    "./coin": "./lib/coin/index.js",
    "./core": "./lib/core/index.js",
    "./nodejs": "./lib/nodejs/index.js",
    "./marmalade": "./lib/marmalade/index.js"
  },
  ```

lsgunnlsgunn:
Going to revisit using multiple sections in the new Docs2.0 structure.

PR #1549, opened 2024-01-30
[@client-utils] Marmalade Deployment Scripts

nil-amrutlal-dept:
This PR contains the marmalade deployment scripts. The goal is to provide in `client-utils` a simple function call that will handle all the processes of deploying marmalade. 

Some auxiliary functionality was also deployed including directory/path management and a GitHub service used to manage files located in the remote repository.

I added some comments on the files so its easier to review them.

Finally, unit tests were also developed to keep the coverage above the designated threshold.

---
  - https://app.asana.com/0/0/1206405918941172


packages/libs/client-utils/src/built-in/deploy-contract.ts
nil-amrutlal-dept, line 1: Generic deploy contract

packages/libs/client-utils/src/nodejs/deploy-from-directory.ts
nil-amrutlal-dept, line 1: Generic deploy from directory

packages/libs/client-utils/src/nodejs/deploy-template.ts
nil-amrutlal-dept, line 1: Generic deploy template

packages/libs/client-utils/src/nodejs/marmalade/deployment/config.ts
nil-amrutlal-dept, line 1: File containing interface definition for all the configurations regarding marmalade deployment

packages/libs/client-utils/src/nodejs/marmalade/deployment/defaults.ts
nil-amrutlal-dept, line 1: Default values for marmalade deployment. This makes it a lot easier to deploy marmalade into the devnet

packages/libs/client-utils/src/nodejs/marmalade/deployment/namespace.ts
nil-amrutlal-dept, line 1: Namespaces have a very particular logic in marmalade. Therefore, this file has some marmalade-specific logic which includes deploying several specific files with different specific namespaces

packages/libs/client-utils/src/nodejs/services/download-git-files.ts
nil-amrutlal-dept, line 1: Necessary since all the templates and code files are downloaded from a GitHub repository

packages/libs/client-utils/etc/client-utils-built-in.api.md
```diff
@@ -4,15 +4,45 @@
...
+import { IPactDecimal } from '@kadena/types';
+import { IPactInt } from '@kadena/types';
 import type { ISignFunction } from '@kadena/client';
+import { ITransactionDescriptor } from '@kadena/client';
+import { PactValue } from '@kadena/types';
+
+// Warning: (ae-forgotten-export) The symbol "IDeployContractInput" needs to be exported by the entry point index.d.ts
+//
+// @alpha (undocumented)
+export const createPactCommand: (inputs: IDeployContractInput) => Partial<IPactCommand>;
```
alber70g, line 22: This is a very generic name. I suggest to make it very specific for contract input. I think we even have a `createPactCommand` function somewhere else already

packages/libs/client-utils/src/built-in/deploy-contract.ts
```diff
@@ -0,0 +1,92 @@
...
+  signers: string[];
+  meta: {
+    gasLimit: number;
+    chainId: ChainId;
+    ttl: number;
+    senderAccount: string;
+  };
+  data?: { key: string; value: string };
+  keysets?: { name: string; pred: string; keys: string[] }[];
+}
```
alber70g, line 19: This already exist somewhere, right? Maybe it can extend from an existing type if you're adding properties

packages/libs/client-utils/src/nodejs/deploy-from-directory.ts
```diff
@@ -0,0 +1,194 @@
...
+  signers: string[];
+  meta: {
+    gasLimit: number;
+    chainId: ChainId;
+    ttl: number;
+    senderAccount: string;
+  };
+  data?: { key: string; value: string };
+  keysets?: { name: string; pred: string; keys: string[] }[];
+}
```
alber70g, line 20: Idem

packages/libs/client-utils/src/nodejs/deploy-from-directory.ts
```diff
@@ -0,0 +1,194 @@
...
+
+interface IDeployConfig {
+  /** These multiple / single values provided for chain will overwrite the
+   * arguments if provided via the deployArguments object
+   */
+  chainIds: ChainId[];
+  templateConfig?: ITemplateConfig;
+  codeFileConfig?: ICodeFileConfig;
+  clientConfig: IClientConfig;
+}
```
alber70g, line 49: Are templateConfig and codeFileConfig ever to be used simultaneousely?
Otherwise I suggest creating a discriminated union for it

packages/libs/client-utils/src/nodejs/deploy-from-directory.ts
```diff
@@ -0,0 +1,194 @@
...
+          } catch (error) {
+            console.error(error);
+          }
+        }
+      } else if (
+        codeFiles &&
+        codeFiles.length > 0 &&
+        codeFileConfig &&
+        !templateConfig
+      ) {
```
alber70g, line 152: Might be good to create two separate functions. One for the first block, and one for the second block. Maybe even move the choice for which one to use up the stack (to the caller code).

Given these two statements: 
```
      if (templateFiles && templateFiles.length > 0) {
```
and

```ts
      if (
        codeFiles &&
        codeFiles.length > 0 &&
        codeFileConfig &&
        !templateConfig
      )
```

packages/libs/client-utils/src/nodejs/marmalade/deployment/defaults.ts
```diff
@@ -0,0 +1,91 @@
...
+  ],
+  secretKey: '251a920c403ae8c8f65f59142316af3c82b631fba46ddea92ee8c95035bd2898',
+};
+
+export const defaultChainId: ChainId = '0';
+export const defaultNetworkId = 'fast-development';
+export const defaultNetworkHost = 'http://localhost:8080';
+
+/* Defines the arguments for the marmalade contracts: these values are going
+to be replaced in the marmalade templates and code files  */
```
alber70g, line 28: We can use actual TSdoc here. 

```suggestion
/** 
 * Defines the arguments for the marmalade contracts: these
 * values are going to be replaced in the marmalade 
 * templates and code files  
 */
```

_Idem for the other occurrences_

packages/libs/client-utils/src/nodejs/marmalade/deployment/deploy.ts
```diff
@@ -0,0 +1,199 @@
...
+          fileExtension: remoteConfig.codefileExtension,
+          networkId: deploymentArguments.network,
+          clientConfig,
+        });
+      }),
+    );
+
+    console.log(`Deploying Marmalade Contracts...`);
+
+    /* sort the templates alphabetically so that the contracts are deployed in the correct order
```
alber70g, line 151: Any reason to use `/* */` for multiline comments and not

```
// comment
// next line
```

or 

```
/** 
 * comment
 * next line
 */
```

packages/libs/client-utils/src/nodejs/marmalade/deployment/deploy.ts
```diff
@@ -0,0 +1,199 @@
...
+      'Downloading marmalade templates and namespace definition files...',
+    );
+
+    await Promise.all([
+      getMarmaladeTemplates({
+        repositoryConfig,
+        remoteConfig,
+        localConfig,
+        flatFolder: true,
+      }),
```
alber70g, line 93: What do these functions do? Their results aren't used, or they return nothing resulting in a side-effect. 

I suppose they write to disk. I'm not sure how I feel about "unused" function calls


_Idem for other calls below_

packages/libs/client-utils/src/nodejs/marmalade/deployment/deploy.ts
```diff
@@ -0,0 +1,199 @@
...
+        // Only b is in marmaladeNamespaceOrder, b comes first
+        return 1;
+      } else if (indexB === -1) {
+        // Only a is in marmaladeNamespaceOrder, a comes first
+        return -1;
+      } else {
+        // Both a and b are in marmaladeNamespaceOrder, sort based on their positions
+        return indexA - indexB;
+      }
+    };
```
alber70g, line 174: Can we export this function so it can be unit-tested in isolation

packages/libs/client-utils/src/nodejs/deploy-from-directory.ts
javadkh2: can you please use the type deom deploy-contract.ts

nil-amrutlal-dept:
Going to merge this PR with comments unhandled. This was discussed with @alber70g and the reasoning was that with pact/toolbox, this code might not be necessary anymore and therefore there's no use in putting too much effort and time into it.

PR #1551, opened 2024-01-30
refactor: removed the code we don't need now

javadkh2:


packages/apps/dev-wallet/src/modules/db/db.service.tsx
alber70g, line 1: The extension can be `ts`

alber70g (approved):
Minor comment

PR #1556, opened 2024-01-31
fix/ add listBox component and udpate tests

Ghislain89:
This PR adds page objects for the listBox component and updates the tools test to use it.


Ghislain89 (commented):
Need to investigate flaky test.

sstraatemans (approved):

PR #1557, opened 2024-01-31
[@kadena/graph-client] Change Account view to include non-fungibles

MRVDH:
- Removed fungible from main search
- Removed non-fungible search
- Added non-fungible tab to account
- Added fungible search to fungible tab account
- Removed obsolete non-fungible pages and routes

https://app.asana.com/0/1205801361945248/1206309584555287/f


nil-amrutlal-dept:
Small note: When searching for a different fungible on the account page, such as 'free,' I initially get no results. However, if I click on the non-fungible tag and then return to the fungible tab, the search value resets to the default 'coin,' but the page retains the results for the 'free' search. 

image

nil-amrutlal-dept (approved):

PR #1560, opened 2024-01-31
[@kadena/graph] Alter migrations

MRVDH:
Alter migrations to support the correct format for devnet. Also removed unnecessary indexes.

https://app.asana.com/0/1205801361945248/1206475741288313/f


alber70g (approved):

nil-amrutlal-dept:
Had to delete the volume to get the new migrations to run successfully. Just making sure this is expected

nil-amrutlal-dept (approved):

PR #1567, opened 2024-02-01
[@tools-website] Add link to Block Explorer on Faucet pages

KristinaSpasevska:
Adds Notification after the submit button with a link to the Kadena Block Explorer with the appropriate Request Key


packages/apps/tools/src/pages/transactions/cross-chain-transfer-finisher/index.tsx
```diff
@@ -302,23 +304,41 @@ const CrossChainTransferFinisher: FC = () => {
...
   const renderNotification =
     txError.toString() === '' && receiverRequestKey ? (
       <FormStatusNotification
         status="successful"
         title={t('Notification title success')}
         body={t('XChain transfer has been successfully finalized!')}
       >
-        <p
-          className={noticationKeyStyle}
-        >{`Request key: ${receiverRequestKey}`}</p>
```
sanderlooijenga, line 314: So this is now also `"Target Chain Request key:"` (instead of `"Request key:"`), that is okay?

sanderlooijenga (approved):

packages/apps/tools/src/pages/transactions/cross-chain-transfer-finisher/index.tsx
KristinaSpasevska: it's already cahnged

sanderlooijenga (approved):

PR #1579, opened 2024-02-05
[@react-ui] Fix: Cursor pointer on whole viewport when modal is open

KristinaSpasevska:


packages/libs/react-ui/src/components/Dialog/Dialog.css.ts
```diff
@@ -34,6 +34,9 @@ export const overlayClass = style([
       maxHeight: '75vh',
     },
   }),
+  {
+    cursor: 'default',
+  },
```
eileenmguo, line 39: is this necessary?

eileenmguo (approved):

PR #1585, opened 2024-02-06
feat(account): update get account details sub command

realdreamer:
https://app.asana.com/0/1205969628270714/1206113181762549/f

`account get-details` sub command


packages/tools/kadena-cli/src/account/commands/accountAddFromWallet.ts
```diff
@@ -66,7 +65,6 @@ export const createAddAccountFromWalletCommand = createCommandFlexible(
   ],
 
   async (option, values) => {
-    await ensureNetworksConfiguration();
```
realdreamer, line 69: Move into `NetworkSelectPrompt` makes more sense. so moved there

packages/tools/kadena-cli/src/networks/utils/networkHelpers.ts
```diff
@@ -120,8 +120,6 @@ export async function ensureNetworksConfiguration(): Promise<void> {
     mkdirSync(defaultNetworksPath, { recursive: true });
   }
 
-  if (readdirSync(defaultNetworksPath).length !== 0) return;
```
realdreamer, line 123: Moved this check also as part of network select prompt.

As part of init config also this method is getting called. So, if user (assume he already have its own custom network) wants to create a default networks config later then this check makes it impossible to create it.

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+  return {
+    name: account?.name ?? '',
+    alias: accountAlias,
+  };
+};
+
+export async function ensureAccountExists(): Promise<void> {
+  if (!(await services.filesystem.directoryExists(ACCOUNT_DIR))) {
+    console.error(`No account created yet. Please create an account first.`);
+    process.exit(1);
```
barthuijgen, line 31: Don't `process.exit()` in helpers, or preferably at all. I know it's been used in some other places but they need to go away as well.  If there is an error just throw.

barthuijgen: Also no console output in helpers.

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+  alias: string;
+}
+
+const readAccountFromFile = async (accountFile: string): Promise<IAccount> => {
+  const accountAlias = accountFile.split('.')[0];
+  const content = await services.filesystem.readFile(
+    join(ACCOUNT_DIR, accountFile),
+  );
+  const account =
+    content !== null ? (yaml.load(content) as IAccountAliasFile) : null;
```
barthuijgen, line 21: please use a zod schema instead of `as Interface`

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -0,0 +1,52 @@
...
+  publicKeys: z.array(z.string()),
+  predicate: z.string(),
+});
+
+const readAccountFromFile = async (accountFile: string): Promise<IAccount> => {
+  const accountAlias = accountFile.split('.')[0];
+  const content = await services.filesystem.readFile(
+    join(ACCOUNT_DIR, accountFile),
+  );
+  const account = content !== null ? yaml.load(content) : null;
```
nillo, line 27: When account is null, don't you want to throw?
accountAliasFileSchema.parse(account)) does not handle null gracefully, so either it throws an error upon encountering null or it doesn't return an object with a name property, then trying to access .name on the result will lead to a runtime error because you're trying to access a property on null

nillo (approved):
lgtm except that one comment

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -0,0 +1,52 @@
...
+  const parsedContent = accountAliasFileSchema.parse(account);
```
barthuijgen, line 28: Make sure zod `.parse()` is wrapped in a try{}catch(){} that gracefully handles when it errors with a good error message.

barthuijgen: or there is `.safeParse()` if that makes it easier

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
realdreamer: zod parse is also taking care of null gracefully, but yeah it's not providing meaningful error message to the user. so wrapped it with `try` and `catch` like Bart suggested.

alber70g (approved):

barthuijgen (approved):

PR #1597, opened 2024-02-07
Feat/expand maskvalue component

sanderlooijenga:
By request of @isaozler I have expanded some of the masking options for the MaskedValue component, please check it out 🙏 

Preview; https://react-ui-git-feat-expand-maskvalue-component-kadena-js.vercel.app/?path=/docs/patterns-maskedvalue--docs

### DIFF
--- a/packages/libs/react-ui/src/components/MaskedValue/utils.ts
+++ b/packages/libs/react-ui/src/components/MaskedValue/utils.ts
@@ -25,7 +25,7 @@ export const maskValue = (
   };
 
   if (_character.length > 1) {
-    console.info('Only one character is allowed');
+    console.warn('Only one character is allowed');
   }
 
   const character = _character.slice(0, 1);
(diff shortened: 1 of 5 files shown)


packages/libs/react-ui/src/components/MaskedValue/utils.ts
```diff
@@ -24,10 +24,12 @@ export const maskValue = (
     ...options,
   };
 
-  if (character.length > 1) {
-    throw new Error('Only one character is allowed');
+  if (_character.length > 1) {
+    console.info('Only one character is allowed');
```
salamaashoush, line 28: I think warn would be more suitable here, since it is a warning

salamaashoush (approved):

packages/libs/react-ui/src/components/MaskedValue/utils.ts
sanderlooijenga: Done!

PR #1611, opened 2024-02-09
feat(client-utils): query all chains

javadkh2:
Introducing accountDiscovery and queryAllChainsClient

```
queryAllChainsClient: accepts a query and do a local call on all chains
accountDiscovery: it calls `details` on all chains


alber70g (approved):
Minor comment

PR #1616, opened 2024-02-09
[@kadena/graph] Account by public key query (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206523433657930


packages/apps/graph/src/graph/query/fungible-account-by-public-key.ts
```diff
@@ -0,0 +1,95 @@
...
+  const result = await prismaClient.transaction.findFirst({
+    where: {
+      AND: [
+        {
+          code: {
+            contains: 'transfer-create',
+          },
+        },
+        {
+          data: {
```
MRVDH, line 72: This is neat 😄

packages/apps/graph/src/graph/query/fungible-account-by-public-key.ts
```diff
@@ -0,0 +1,95 @@
+import { prismaClient } from '@db/prisma-client';
+import { getFungibleChainAccount } from '@services/account-service';
+import { chainIds } from '@utils/chains';
+import { normalizeError } from '@utils/errors';
+import { builder } from '../builder';
+import FungibleAccount from '../objects/fungible-account';
+import type { FungibleChainAccount } from '../types/graphql-types';
+import { FungibleAccountName } from '../types/graphql-types';
+
+builder.queryField('fungibleAccountByPublicKey', (t) =>
```
MRVDH, line 10: Do we want to add the same but on the chain level?

`fungibleChainAccountByPublicKey`

MRVDH (approved):

packages/apps/graph/src/graph/query/fungible-account-by-public-key.ts
nil-amrutlal-dept: Yep, in fact I think it makes sense to also align with the rest of the existing queries. I'll update it

alber70g (approved):

PR #1622, opened 2024-02-12
feat(account): implement fund commands for existing / new account

realdreamer:
`account fund` command
https://app.asana.com/0/1205969628270714/1206113181762551/f

removed account create and account transfer-create commands 

Screenshot reference for how alias and account name displayed to the user

image


packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -1,73 +1,147 @@
...
+  'fund an existing/new account',
   [
-    globalOptions.accountName(),
+    globalOptions.accountSelect(),
     globalOptions.amount(),
-    globalOptions.fungible(),
-    globalOptions.network(),
+    networkSelect(),
     globalOptions.chainId(),
+    deployDevnet(),
```
nillo, line 56: Devnet should be disabled, in this pr, since we are not shipping devnet support with the first release

packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -1,73 +1,147 @@
...
       console.log(
         chalk.red(
-          `\nFailed to fund the account on "${config.network}".\n`,
+          `\nNetwork "${network}" of id "${networkConfig.networkId}" is not supported.\n`,
         ),
       );
+      return;
     }
+
+    if (networkConfig.networkId === 'fast-development') {
```
nillo, line 83: disable

packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -1,73 +1,147 @@
...
+        console.log(
+          chalk.yellow(
+            `\nFaucet module is not available on chain "${chainId}" in "${networkConfig.network}".\n`,
+          ),
+        );
+
+        // TODO: Verify this use case with "ALBERT"
+        if (config.accountConfig.fungible !== 'coin') return;
+
+        const { deployDevnet } = await option.deployDevnet();
```
nillo, line 111: disable

packages/tools/kadena-cli/src/account/utils/__tests__/getAccountDetails.test.ts
```diff
@@ -81,13 +81,13 @@ describe('getAccountDetailsFromChain', () => {
...
 
-describe('getAccountDetailsForAddAccount', () => {
+describe('getAccountDetails', () => {
   afterEach(() => {
     server.resetHandlers();
   });
 
   it('should return account details from chain when account is available on chain', async () => {
-    const result = await getAccountDetailsForAddAccount({
+    const result = await getAccountDetails({
```
nillo, line 90: add test for other chain (to see if url is constructed correcrly)

packages/tools/kadena-cli/src/account/utils/transferFund.ts
```diff
@@ -0,0 +1,80 @@
...
+      `${networkConfig.networkHost}/chainweb/0.0/${networkConfig.networkId}/chain/${chainId}/pact`,
+    );
+
+    const requestKeys = await submit(signedTx);
+
+    const response = await listen(requestKeys);
+
+    return response;
+  } catch (error) {
+    throw new Error(error.message);
```
nillo, line 78: if you are throwing the same error again just 

throw error

nillo (changes requested):
i havent looked at the devnet code since its outside of the scope of the PR, but great you already did some work on it!

packages/tools/kadena-cli/src/account/utils/__tests__/fund.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      },
+      amount: '100',
+      networkConfig: testNetworkConfigMock,
+      chainId: '1',
+    });
+
+    expect(result).toStrictEqual({
+      success: false,
+      errors: [
+        'Failed to transfer fund : "Cannot read properties of undefined (reading \'map\')"',
```
barthuijgen, line 149: I don't think this error should be expected, this error means there is no defensive programming used. Please fix it so this gives a proper error message

packages/tools/kadena-cli/src/account/utils/transferFund.ts
```diff
@@ -0,0 +1,85 @@
...
+  config,
+}: {
+  accountName: string;
+  config: {
+    amount: string;
+    contract: string;
+    chainId: ChainId;
+    networkConfig: INetworkCreateOptions;
+  };
+}): Promise<ICommandResult | string> {
```
barthuijgen, line 26: Please don't mix `ICommandResult` with something else, either use only `ICommandResult<string>` or just `string`

packages/tools/kadena-cli/src/devnet/faucet/deploy/faucet.ts
```diff
@@ -0,0 +1,72 @@
...
+}: {
+  chainId: ChainId;
+  upgrade: boolean;
+  namespace: string;
+}): Promise<Record<string, ICommandResult>> => {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const __filename = fileURLToPath(import.meta.url);
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  const __dirname = path.dirname(__filename);
+  const faucetContract = fs.readFileSync(
```
barthuijgen, line 29: this won't work when we publish. Please just put the contract in a `.ts` file and do 
```
export const contract = `
<contract>
`;
```

packages/tools/kadena-cli/src/prompts/account.ts
```diff
@@ -51,7 +55,21 @@ export const accountKdnNamePrompt: IPrompt<string> = async () =>
...
+
+      if (isNaN(parsedValue)) {
+        return 'Please enter a valid amount.';
+      }
+
+      if (parsedValue <= 0) {
+        return 'Please enter a positive amount.';
+      }
+
+      if (parsedValue > 100) {
```
barthuijgen, line 68: Why is there a 100 limit on `amountPrompt`? This seems like a generic prompt, if there is some specific scenario where 100 limit makes sense, please name the method accordingly

packages/tools/kadena-cli/src/account/utils/__tests__/fund.test.ts
realdreamer: actually this error is from the chain itself, i'll have a look is it because of my mocks

nillo (approved):
lgtm i didnt take a deep dive into the devnet part since, devnet is up for change anyways.

other then that looks good

PR #1626, opened 2024-02-12
New Icon package base on design system repo

salamaashoush:
~99% code of the PR is autogenerated, just review this [script](https://github.com/kadena-community/kadena.js/pull/1626/files#diff-b7f33cf88351dfcc55c8c5e4b5eec792e2c3689878f84dd39da1b8ea988193b2)

Icons have been created from the design system repository with the following considerations:

1. The default fill color is set to `currentColor`.
2. The dimensions for width and height are established at `1em`, while the default `fontSize` is `1.5em`.
3. These icons are compiled into a distinct package, which is designed without knowledge of the design tokens. This approach ensures that properties such as fill and fontSize remain generic, eliminating circular dependencies between the `@kadena/react-ui` and allowing the icons to be universally compatible with any build-system and react styling solutions.
4. Icons are grouped in different folders and package entries eg `@kadena/react-icons/system`.


sstraatemans (changes requested):
Can we maybe have a search option in storybook?
and the names visible of the icons. So when i CTRL+F I can search for a particular icon. 
Impossible to find the correct one atm

sstraatemans (approved):

PR #1636, opened 2024-02-14
Tools dApp - Ledger hooks

sanderlooijenga:


packages/apps/tools/src/hooks/use-ledger-public-key.ts
```diff
@@ -0,0 +1,57 @@
+import { getTransport } from '@/utils/getTransport';
+import AppKda from '@ledgerhq/hw-app-kda';
+import { useAsyncFn } from 'react-use';
+
+export const derivationModes = [
+  'current',
+  'legacy',
+  // 'kip26'
+] as const;
```
KristinaSpasevska, line 9: Are we having only modes for the derivation path? The reason I'm asking is that in the UI part I've built a toggle (isLegacy on/off) and it could work if they're only two ways, but not working for 3 or more

KristinaSpasevska (approved):

packages/apps/tools/src/hooks/use-ledger-public-key.ts
sanderlooijenga: For now yes, if and when the KIP 26 will be accepted then we might have to add this as an option as well and update the UI for it as well. But for now we should be okay with having only these two, which is the same as the current Transfer ledger page

PR #1638, opened 2024-02-14
[@kadena/graph] Remove coin related data from non-fungible account (Fix)

nil-amrutlal-dept:

### DIFF
--- a/packages/apps/graph/src/services/account-service.ts
+++ b/packages/apps/graph/src/services/account-service.ts
@@ -48,20 +48,13 @@ export async function getNonFungibleChainAccount({
   chainId: string;
   accountName: string;
 }): Promise<NonFungibleChainAccount | null> {
-  const [accountDetails, tokenDetails] = await Promise.all([
-    accountDetailsLoader.load({ accountName, chainId, fungibleName: 'coin' }),
-    tokenDetailsLoader.load({ accountName, chainId }),
-  ]);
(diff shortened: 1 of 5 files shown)


packages/apps/graph/src/services/account-service.ts
```diff
@@ -48,20 +48,13 @@ export async function getNonFungibleChainAccount({
...
   accountName: string;
 }): Promise<NonFungibleChainAccount | null> {
-  const [accountDetails, tokenDetails] = await Promise.all([
-    accountDetailsLoader.load({ accountName, chainId, fungibleName: 'coin' }),
-    tokenDetailsLoader.load({ accountName, chainId }),
-  ]);
+  const tokenDetails = await tokenDetailsLoader.load({ accountName, chainId });
 
-  return accountDetails !== null && tokenDetails !== null
+  return tokenDetails !== null && tokenDetails.length !== 0
```
nil-amrutlal-dept, line 53: @alber70g If no tokens are found, it will return null. If there is a `create-account` like action for marmalade, I can check if there's a way to check (not from reconcile view)

MRVDH (approved):

packages/apps/graph/src/services/account-service.ts
alber70g: For now we can leave it. An empty account isn't that interesting anyway.

alber70g (approved):

PR #1641, opened 2024-02-15
fix(account): add default values when user use --quiet option

realdreamer:
account commands with `--quiet` flag needs few defaults and additional check. Fixing those as part of this PR.

### DIFF
--- a/packages/tools/kadena-cli/src/utils/globalOptions.ts
+++ b/packages/tools/kadena-cli/src/utils/globalOptions.ts
@@ -96,7 +96,7 @@ export const globalOptions = {
     ),
     expand: async (publicKeys: string) => {
       return publicKeys
-        .split(',')
+        ?.split(',')
         .map((value) => value.trim())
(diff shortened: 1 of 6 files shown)


packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -89,7 +89,7 @@ export const globalOptions = {
     ),
     expand: async (publicKeys: string) => {
       return publicKeys
-        .split(',')
+        ?.split(',')
```
barthuijgen, line 92: how come it was able to be undefined? validation schema also is `z.string()` so this should not be possible right

barthuijgen (approved):

packages/tools/kadena-cli/src/utils/globalOptions.ts
realdreamer: it's an optional field so it could be undefined

PR #1643, opened 2024-02-15
[@kadena/graph] Add TokenInfo to Token (Feat)

nil-amrutlal-dept:
image

### DIFF
--- a/packages/apps/graph/src/devnet/simulation/marmalade/get-token-info.ts
+++ b/packages/apps/graph/src/devnet/simulation/marmalade/get-token-info.ts
@@ -1,0 +1,52 @@
+import type { ChainId, IPactModules, PactReturnType } from '@kadena/client';
+import { Pact } from '@kadena/client';
+import { dirtyReadClient } from '@kadena/client-utils/core';
+import { composePactCommand, execution, setMeta } from '@kadena/client/fp';
+import { dotenv } from '@utils/dotenv';
+
+interface TokenInfo {
+  supply: number;
+  precision: number;
+  uri: string;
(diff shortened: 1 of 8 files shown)


packages/apps/graph/src/devnet/simulation/marmalade/get-token-info.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  if ('precision' in tokenInfo) {
+    if (
+      typeof tokenInfo.precision === 'object' &&
+      tokenInfo.precision !== null
+    ) {
+      tokenInfo.precision = (tokenInfo.precision as { int: number }).int;
+    }
+  }
```
nil-amrutlal-dept, line 49: The property precision can be 
`precision: {int: number}`
or 
`precision: int`

Hence this

alber70g (approved):

packages/apps/graph/src/graph/objects/token.ts
```diff
@@ -6,5 +8,21 @@ export default builder.objectType('Token', {
     id: t.exposeID('id'),
     balance: t.exposeInt('balance'),
     chainId: t.exposeInt('chainId'),
+    info: t.field({
+      type: 'TokenInfo',
+      nullable: true,
+      resolve: async (parent) => {
```
MRVDH, line 14: this can be `async resolve(parent) {` I think

MRVDH, line 14: Could you try catch the whole function with the error normaliser that we also use in the other queries?

PR #1649, opened 2024-02-15
[@kadena/graph] Optimize non-fungible queries and add marmalade-v1 token support (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206601617273555


packages/apps/graph/generated-schema.graphql
```diff
@@ -384,7 +384,7 @@ type Subscription {
 """The token identifier and its balance."""
 type Token {
   balance: Int!
-  chainId: Int!
+  chainId: String!
```
nil-amrutlal-dept, line 387: This is to match the rest of the schema

packages/apps/graph/src/graph/objects/token.ts
```diff
@@ -1,27 +1,32 @@
...
-        }
+          if (!tokenInfo) {
+            return null;
+          }
 
-        return tokenInfo;
+          return tokenInfo;
+        } catch (error) {
+          throw normalizeError(error);
+        }
```
nil-amrutlal-dept, line 29: https://github.com/kadena-community/kadena.js/pull/1643#discussion_r1490918741

packages/apps/graph/src/graph/objects/non-fungible-account.ts
```diff
@@ -51,18 +47,19 @@ export default builder.node(
...
-                }),
-              )
-            ).filter(
-              (chainAccount) => chainAccount !== null,
-            ) as NonFungibleChainAccount[];
+            const chainIds = await nonFungibleChainCheck.load({
+              accountName: parent.accountName,
+            });
+
+            return chainIds.map(async (chainId) => {
```
MRVDH, line 54: doesn't need to be async I think

alber70g (approved):

PR #1650, opened 2024-02-15
[@kadena/graph] Gas Limit Estimations v2

MRVDH:
You can get the gas limit estimation for any transaction by using the `gasLimitEstimate` query. The input accepts a JSON object and based on the parameters passed it will determine what type of format it is and return the gas limit estimation. The following types are supported:

- `full-transaction`: A complete transaction object. Required parameters: `cmd`, `hash` and `sigs`.
- `parsed-command`: A JSON stringified command. Required parameters: `cmd`. It also optionally accepts `sigs`.
- `full-command`: A full command. Required parameters: `payload`, `meta` and `signers`.
- `partial-command`: A partial command. Required parameters: `payload` and either `meta` or `signers`. In case `meta` is not given, but `signers` is given, you can also add `chainId` as a parameter.
- `payload`: A just the payload of a command. Required parameters: `payload` and `chainId`.
- `code`: The code of an execution. Required parameters: `code` and `chainId`.

Every type accepts an optional parameter called `networkId` to override the default value from the environment variables.


packages/apps/graph/README.md
```diff
@@ -23,8 +23,10 @@ A GraphQL endpoint that interacts with chainweb-data and chainweb-node.
   - [Simulate traffic on the devnet](#simulate-traffic-on-the-devnet)
     - [Coin simulation](#coin-simulation)
     - [Marmalade simulation](#marmalade-simulation)
+    - [Flood devnet](#flood-devnet)
```
nil-amrutlal-dept, line 26: Ooops. Good catch 👍🏽

nil-amrutlal-dept (approved):

PR #1659, opened 2024-02-16
fix(cli): minor bugs + improvements

realdreamer:
- Improved error message for account details not found in chain
- instead of network name, `undefined` is shown in the network delete confirmation prompt
- Fix keys gen-plain --key-alias="alpha22" --key-amount="5" 
the file name shown in the console and the key's generated are not same 
- keys gen-plain --quiet   => `keyAlias`  and `key message` set required field
- Updated Wallet `verify password` option to `-vp` . `-p` already used by password
- Make network select option as required field as part of network manage and delete command

### DIFF
--- a/packages/tools/kadena-cli/src/utils/globalOptions.ts
+++ b/packages/tools/kadena-cli/src/utils/globalOptions.ts
@@ -520,7 +520,7 @@ export const globalOptions = {
     prompt: security.securityPasswordVerifyPrompt,
     validation: z.string(),
     option: new Option(
-      '-p, --security-verify-password <securityVerifyPassword>',
+      '-vp, --security-verify-password <securityVerifyPassword>',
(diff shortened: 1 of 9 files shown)


packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -520,7 +520,7 @@ export const globalOptions = {
     prompt: security.securityPasswordVerifyPrompt,
     validation: z.string(),
     option: new Option(
-      '-p, --security-verify-password <securityVerifyPassword>',
+      '-vp, --security-verify-password <securityVerifyPassword>',
```
barthuijgen, line 523: we need to remove all multi-character options with a single `-`, in this case just remove `-vp` entirely I think.

barthuijgen (approved):

PR #1660, opened 2024-02-16
fix(keys): loader and prompt exit

realdreamer:
image

fixing the loader state remain struck for wrong password


nillo (approved):
lgtm, i wonder when this issue was introduced.

barthuijgen (approved):

PR #1662, opened 2024-02-17
feat(pou): framermotion

sstraatemans:


packages/apps/proof-of-us/src/components/List/ListItem.tsx
```diff
@@ -28,15 +29,18 @@ interface ITempToken extends Token {
...
+  const uri = (token as ITempToken).info?.uri;
   const { data, isLoading } = useSWR(uri, fetchManifestData, {
     revalidateOnFocus: false,
   });
 
   return (
-    <li className={listItemClass}>
+    <motion.li
+      className={listItemClass}
+      initial={{ opacity: 0, left: '500px' }}
```
ash-vd, line 40: If possible, it's better to use transforms so the animations are you accelerated. You can try

```suggestion
      initial={{ opacity: 0, x: '500px' }}
```

ash-vd (approved):

packages/apps/proof-of-us/src/components/List/ListItem.tsx
sstraatemans: you are correct of course.

Ghislain89 (approved):

PR #1684, opened 2024-02-20
[@kadena/graph] Added more filters to Events query and subscription

MRVDH:
### Prisma JSON field queries

Some columns in the database are of type `jsonb`. To query these columns, you can supply a stringified JSON object that matches the [JSON object property filters](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields#filter-on-object-property) from Prisma.

An example of such a filter parameter value could be: `events(parametersFilter: "{\"array_starts_with\": \"k:abcdefg\"}")`, in which case, the `parameters` column is a `jsonb` type column which is filtered to only include rows where the value contains an array that has the string `k:abcdefg` on index 0.

Queries that allow such filters:

- `parametersFilter` on the `events` query and `events` subscription.

---
  - https://app.asana.com/0/0/1206592009540818

### DIFF
--- a/packages/apps/graph/src/graph/query/events.ts
+++ b/packages/apps/graph/src/graph/query/events.ts
@@ -39,6 +39,8 @@ builder.queryField('events', (t) =>
     edgesNullable: false,
     args: {
       qualifiedEventName: t.arg.string({ required: true }),
+      chainId: t.arg.string(),
+      parametersFilter: t.arg.string(),
     },
     type: 'Event',
     cursor: 'blockHash_orderIndex_requestKey',
@@ -69,6 +71,12 @@ builder.queryField('events', (t) =>
             transaction: {
               NOT: [],
             },
+            ...(args.chainId && {
+              chainId: parseInt(args.chainId),
+            }),
+            ...(args.parametersFilter && {
+              parameters: JSON.parse(args.parametersFilter),
+            }),
           },
           orderBy: {
             id: 'desc',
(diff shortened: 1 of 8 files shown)


packages/apps/graph/src/graph/query/events.ts
```diff
@@ -69,6 +71,12 @@ builder.queryField('events', (t) =>
             transaction: {
               NOT: [],
             },
+            ...(args.chainId && {
+              chainId: parseInt(args.chainId),
+            }),
+            ...(args.parametersFilter && {
+              parameters: JSON.parse(args.parametersFilter),
+            }),
```
nil-amrutlal-dept, line 79: The `totalCount` for events does not consider these newly added arguments.
Eg.
image

packages/apps/graph/src/graph/query/events.ts
nil-amrutlal-dept, line 1: Maybe we could also implement a type of error regarding Syntax Errors. It'd be easier for a user to detect that the provided parameter is invalid. 
eg.
image

packages/apps/graph/src/graph/query/events.ts
MRVDH: Good catch, fixed!

nil-amrutlal-dept (approved):

packages/apps/graph/src/graph/query/events.ts
MRVDH: Done, let me know what you think

PR #1685, opened 2024-02-20
Fix styles to make camera for the whole page

Likita:


packages/apps/proof-of-us/src/components/UserLayout/UserLayout.tsx
```diff
@@ -19,7 +19,7 @@ const UserLayout: FC<PropsWithChildren> = ({ children }) => {
 
   return (
     <motion.div
-      style={{ position: 'absolute', width: '100%' }}
+      style={{ position: 'relative', width: '100%' }}
```
sstraatemans, line 22: this needs to be on absolute. otherwise the router layout animations will not work correctly

sstraatemans (approved):

PR #1686, opened 2024-02-20
chore(pact-project-proposal): initial proposal

alber70g:

### DIFF
--- a/packages/pocs/pact-project-proposal/README.md
+++ b/packages/pocs/pact-project-proposal/README.md
@@ -1,0 +1,43 @@
+# Concept of project management for Pact projects
+
+Inspired by npm package- and tsconfig project-management.
+
+A `pact-project.json` needs to be in every package.
+
+> When that doesn't exist, it can be made by a depending project. It'll then
+> include a "source" property that points to the source files.
+
+The `pact-project.json` has properties that'll describe the project:
+
+- `name` (optional): is used to reference the project, or show in logs or error
+  messages
+- `description` (optional): description of the contents of the pact project
+- `dependencies` (optional): an list of dependencies that need to be made
+  available for the project to work properly
+- `dependencyTree`: an list of objects that define a smart contract and its
+  dependencies (recursively through `dependencies`)
+  - `path`: the path to the contract that needs to be available for the
+    depending project to work
+  - `dependencies`: a list of object that define this contracts' dependencies.
+
+# How should this be used
+
+A tool can interpret the `pact-project.json` to do various tasks
+
+## `kadena project`
+
+- `install` make dependencies available through downloading or other mechanism
+- `verify` verify that all dependencies are available
+- `deploy` will check whether the target network has the contracts available,
+  and deploys them where needed (new or upgrade deployment)
+
+# Example files
+
+There are 3 files in this project
+
+- [`./pact-project.json`](./pact-project.json) an example of a project config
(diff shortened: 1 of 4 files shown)


packages/pocs/pact-project-proposal/README.md
```diff
@@ -0,0 +1,43 @@
...
+- `name` (optional): is used to reference the project, or show in logs or error
+  messages
+- `description` (optional): description of the contents of the pact project
+- `dependencies` (optional): an list of dependencies that need to be made
+  available for the project to work properly
+- `dependencyTree`: an list of objects that define a smart contract and its
+  dependencies (recursively through `dependencies`)
+  - `path`: the path to the contract that needs to be available for the
+    depending project to work
+  - `dependencies`: a list of object that define this contracts' dependencies.
```
emilypi, line 21: So this makes sense, but we need to tease out some structure here: 

1. What does it mean to be a _unit of distribution_, as in, what is considered a singular, self-contained unit that can be delivered via its source code and deployed on the blockchain in an automatic fashion? I think this is a good start. You'll also probably want to include the install script (`tx.yaml` or smth) as a field in this distribution. I imagine `deploy` would only understand how to upload and deploy to a network if that were present. 
2. What does it mean to be a _project_, distinct from a unit of distribution? I would deploy a unit of distribution, but what would i build upon? A project, to me, should contain the following: 
    - A list of dependencies, and a specification for where they're being stored, whether we have a default location local to the project, or user-specified. 
    - A distinction between source and test directories, and tools to test that i could compose with deployment to build a CI/CD pipeline.
    - directories I could specify to dump FV or coverage artifacts

Together, i think 1 and 2 would be enough of a start to talk about a fully fledged "Pact Project" . 

HTH

packages/pocs/pact-project-proposal/chainweb-pact-project.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "chainweb-pact",
+	"description": "Chainweb Pact Project",
+  "source": "git@github.com:kadena-io/chainweb-pact.git",
+	"dependencies": [
+		"git@github.com:other-user/my-other-dependency.git"
```
emilypi, line 6: I like the idea of content address-ability here. We should figure out how to reference contract hashes directly with some nice syntax. Having github support would be killer to mix with that.

packages/pocs/pact-project-proposal/chainweb-pact-project.json
```diff
@@ -0,0 +1,18 @@
...
+	],
+  "contracts": [
+    {
+      "path": "./utils.pact",
+      "dependencies": [
+        {
+          "path": "./some-other-contract.pact"
+        }
```
emilypi, line 14: hash or checksum would be nice to include here.

packages/pocs/pact-project-proposal/chainweb-pact-project.json
alber70g: I'd like this to be similar to how NPM allows to use dependencies. This includes various ways of specifying versions, git urls, github urls (`kadena-community/a-repo#commit-sh`) etc

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#dependencies

emilypi: That's fine by me, but we should specify the variations in syntax that we plan to support beforehand.

PR #1688, opened 2024-02-20
Add checkStatus to Ecko signing functions

### LINKED ISSUE #1613: Give a way to retrieve the account from eckoWallet
### Please confirm the following:

- [X] I have read the [Code of Conduct](https://github.com/kadena-community/kadena.js/blob/main/CODE_OF_CONDUCT.md)
- [X] The feature was not suggested before

### Is your feature request related to a problem?

While eckoWallet gives an API to retrieve the account of the connected wallet, Kadena-JS doesn't wrap it and doesn't give any standard way to retrieve it.
IMO, this is a very important use case… that should be handled by Kadena-JS client.

For example, in my frontend I had to add a small and ugly workaround:

https://github.com/Marmalade-NG/marmalade-ng-explorer/blob/06ee5e2dd82b2351abcd88a2e57916395009d62a/src/Buying.jsx#L21

Instead of using 100% the kadena-JS abstraction.

### Describe the solution you'd like

I would propose to add a new "side function" in the eckoWallet module: 
I don't know exactly how should look like the API..

https://github.com/kadena-community/kadena.js/blob/7d06bd2d9b4a0e0172983bc48d1ae2f86b51120d/packages/libs/client/src/signing/eckoWallet/quicksignWithEckoWallet.ts#L85

### Describe alternatives you've considered

_No response_

### Do you have any additional context?

_No response_

### Participation

- [ ] I am willing to submit a pull request for this feature request.

ash-vd:
Adds a wrapper around the `kda_checkStatus` function from Ecko Wallet, which returns the currently connected account.

Closes #1613

### DIFF
--- a/.changeset/quiet-berries-appear.md
+++ b/.changeset/quiet-berries-appear.md
@@ -1,0 +1,5 @@
+---
+'@kadena/client': minor
+---
+
+Add `eckoStatus` to `createEckoWalletQuicksign` and `createEckoWalletSign`.
--- a/packages/libs/client/etc/client.api.md
+++ b/packages/libs/client/etc/client.api.md
@@ -75,6 +75,9 @@ export const createWalletConnectQuicksign: typeof createQuicksignWithWalletConne
 export const createWalletConnectSign: typeof createSignWithWalletConnect;
 
 // @public
+export type EckoStatus = 'success' | 'fail';
+
+// @public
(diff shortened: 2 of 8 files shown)


alber70g (approved):
Note for later: we need to add documentation for the various signWith-methods to the readme

PR #1690, opened 2024-02-21
[@kadena/graph] Added Events to Block and removed obsolete Block fields

MRVDH:
- Removed Block.parentHash
- Removed Block.predicate: String, predicate can be accessed through MinerAccount
- Added Block.events property to access all events from block
- Added Events to block view in graph-client

https://app.asana.com/0/1205801361945248/1206591792021826

### DIFF
--- a/packages/apps/graph/src/graph/objects/block.ts
+++ b/packages/apps/graph/src/graph/objects/block.ts
@@ -28,12 +28,10 @@ export default builder.prismaNode('Block', {
         'The moment the difficulty is adjusted to maintain a block validation time of 30 seconds.',
     }),
     height: t.expose('height', { type: 'BigInt' }),
-    parentHash: t.exposeString('parentBlockHash'),
     payloadHash: t.exposeString('payloadHash'),
     powHash: t.exposeString('powHash', {
       description: 'The proof of work hash.',
     }),
-    predicate: t.exposeString('predicate'),
     minerAccount: t.field({
       type: FungibleChainAccount,
       complexity: COMPLEXITY.FIELD.PRISMA_WITH_RELATIONS,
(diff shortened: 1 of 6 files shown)


packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -131,6 +129,47 @@ export default builder.prismaNode('Block', {
...
+              blockHash: (parent as Block).hash,
+            },
+            orderBy: {
+              orderIndex: 'asc',
+            },
+          });
+        } catch (error) {
+          throw normalizeError(error);
+        }
+      },
```
nil-amrutlal-dept, line 170: I was thinking. If we on the query level select events, then here we wouldn't need to make another query. We could just get the events from the parent object. 
So on `query.object` I'd look like this:

```ts
  try {
    const block = await prismaClient.block.findUnique({
      ...query,
      where: {
        hash: args.hash,
      },
      select: {
        ...query.select,
        events: true,
      },
    });
```

I am not sure if this is the best approach, let me know your thoughts :)

packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -131,6 +129,47 @@ export default builder.prismaNode('Block', {
...
+      edgesNullable: false,
+      complexity: (args) => ({
+        field: getDefaultConnectionComplexity({
+          first: args.first,
+          last: args.last,
+        }),
+      }),
+      select: {
+        hash: true,
+      },
```
nil-amrutlal-dept, line 144: Why don't we include events here? That way we can access it on the resolver via the parent object.

packages/apps/graph/src/graph/objects/block.ts
nil-amrutlal-dept: Forget it, found a better approach. See https://github.com/kadena-community/kadena.js/pull/1690/files#r1497421049

packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -131,6 +129,47 @@ export default builder.prismaNode('Block', {
...
+      async totalCount(parent) {
+        try {
+          return await prismaClient.event.count({
+            where: {
+              blockHash: (parent as Block).hash,
+            },
+          });
```
nil-amrutlal-dept, line 151:
```suggestion
           return (
            parent as Prisma.BlockGetPayload<{ include: { events: true } }>
          ).events.length;
```

nil-amrutlal-dept (approved):
The commented changes are going to be studied further and discussed outside of this PR's scope

PR #1691, opened 2024-02-21
[@kadena/graph], [@kadena/graph-client] Refactor Confirmation Depth (Feat)

nil-amrutlal-dept:
In this PR:

- Removed confirmation depth from block object
- Removed max confirmation depth from graph configuration and dotenv
- Implemented block from depth query
- Implemented new blocks from depth subscription
- Removed confirmation depth occurrences in client (including from block and graph configuration)

---
  - https://app.asana.com/0/0/1206591792021822


MRVDH (changes requested):
in graph-client .env.example can be removed and env.ts as well I think, right?

MRVDH (approved):

PR #1694, opened 2024-02-21
feat(account): implement account list sub command

realdreamer:
https://app.asana.com/0/1205969628270714/1206644161687865/f

`account list`

Screenshot reference for single `alias`  and `all` selection 

image

image

### DIFF
--- a/packages/tools/kadena-cli/src/account/commands/accountList.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountList.ts
@@ -68,8 +68,6 @@ export const createAccountListCommand: (
 
     const tabularData = generateTabularData(accounts);
 
-    log.output(
-      log.generateTableString(tabularData.header, tabularData.data, true, true),
-    );
+    log.output(log.generateTableString(tabularData.header, tabularData.data));
   },
 );
--- a/packages/tools/kadena-cli/src/account/commands/accountList.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountList.ts
@@ -1,4 +1,3 @@
-import Table from 'cli-table3';
 import type { Command } from 'commander';
 import { createCommand } from '../../utils/createCommand.js';
 import { globalOptions } from '../../utils/globalOptions.js';
--- a/packages/tools/kadena-cli/src/account/commands/accountList.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountList.ts
@@ -1,0 +1,74 @@
+import Table from 'cli-table3';
+import type { Command } from 'commander';
+import { createCommand } from '../../utils/createCommand.js';
+import { globalOptions } from '../../utils/globalOptions.js';
+import {
+  maskStringPreservingStartAndEnd,
+  truncateText,
+} from '../../utils/helpers.js';
+import { log } from '../../utils/logger.js';
+import type { IAliasAccountData } from '../types.js';
+import { getAllAccounts } from '../utils/accountHelpers.js';
+
+async function generateAccountTabularData(config: {
+  account: string;
+  accountConfig: IAliasAccountData | undefined;
+}): Promise<{ header: string[]; data: string[][] } | undefined> {
+  const header = [
+    'Alias',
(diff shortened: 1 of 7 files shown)


packages/tools/kadena-cli/src/account/commands/accountList.ts
```diff
@@ -0,0 +1,75 @@
...
+    log.debug('account-list:action', { config });
+
+    const tabularData = await generateAccountTabularData(config);
+
+    if (!tabularData) {
+      return log.error(`Selected account "${config.account}" not found.`);
+    }
+
+    log.output(
+      log.generateTableString(tabularData.header, tabularData.data, true, true),
```
nillo, line 72: log.generateTableString(tabularData.header, tabularData.data) @alber70g adviced me to switch lines off

packages/tools/kadena-cli/src/utils/tableDisplay.ts
```diff
@@ -31,11 +34,55 @@ export function displayTable(
...
+        return cell.map((data: string) => data.padEnd(columnWidth));
+      });
+
+      const maxRows = Math.max(
+        ...rowStrings.map((cellData) => cellData.length),
+      );
+
+      //  We need to pad the rows with empty strings to make sure all rows have the same length
+      // so that new line data from the same column aligned properly to the header
+      const contentRows = [];
```
nillo, line 61: i like it, but maybe you can modularise the complexity, by extracting the logic for caculating the column with, and the formatting of the row

calculateColumWidth fn
formatRowfn

since its now became a bit large, and its less readible. 

Other then that nice addition

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -672,6 +672,30 @@ export const globalOptions = {
...
+    ): Promise<IAliasAccountData | undefined> => {
+      try {
+        if (accountAlias === 'all') {
+          return;
+        }
+
+        const accountDetails = await readAccountFromFile(accountAlias);
+        return accountDetails;
+      } catch (error) {
+        throw new Error(error.message);
```
nillo, line 695: @barthuijgen we probably shouldnt throw errors from the createOptions, it happens in 5 more places toughts

nillo (changes requested):
lgtm but see if my comments with some refactor tips look usefull to you.

packages/tools/kadena-cli/src/account/commands/accountList.ts
nillo: should still be false false.

nillo (approved):
lgtm but switch of lines in table hor: false, ver: false

packages/tools/kadena-cli/src/utils/globalOptions.ts
barthuijgen: Now that I'm looking at it more, especially in `expand` I would not want to see an error, the return type is already `| undefined` so that makes it even weirder.

packages/tools/kadena-cli/src/utils/tableDisplay.ts
```diff
@@ -1,41 +1,118 @@
...
-      ? ''
-      : row
-          .map((item, index) => String(item).padEnd(columnWidths[index], ' '))
-          .join(includeVerticalSeparator ? ' | ' : ' ');
+const formatHeaderRow = (
+  row: TableRow,
+  columnWidths: number[],
+  includeVerticalSeparator: boolean,
+): string =>
+  row.length === 0
```
barthuijgen, line 77: the length check doesn't really add anything here. You can do `.map` on an empty array just fine, and doing `.join()` on an empty array will always result in `""` which is what you want

packages/tools/kadena-cli/src/utils/globalOptions.ts
nillo: agreed, we will make a new ticket to remove the throwing of errors from the prompts

PR #1696, opened 2024-02-22
feat(pou): Add squared capture & styles for camera mode

Likita:
For vertical devices like smartphones


packages/apps/proof-of-us/src/pages/user/proof-of-us/[id]/index.tsx
```diff
@@ -11,11 +11,11 @@ interface IProps {
 
 const Page: NextPage<IProps> = ({ params }) => {
   return (
-    <UserLayout>
+    // <UserLayout>
```
Comment (sstraatemans) [inline packages/apps/proof-of-us/src/pages/user/proof-of-us/[id]/index.tsx:14]: we need this for layout animations

sstraatemans (approved):

PR #1697, opened 2024-02-22
Feat/kadena cli/command tests

barthuijgen:


packages/tools/kadena-cli/src/tx/tests/createTransaction.test copy.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, it } from 'vitest';
+import { mockPrompts, runCommand } from '../../utils/test.util.js';
```
nillo, line 2: fix filename? copy.ts?

packages/tools/kadena-cli/src/utils/logger.ts
```diff
@@ -77,7 +77,7 @@ export const defaultTransport: Transport = (record) => {
     [LEVELS.info]: stdErrChalk.white,
```
nillo, line 77: still doubting these color levels, and maybe use / create something for the inversion? when using black / white terminal. 

Maybe just black on white bg, and white on black bg

nillo (approved):
lgtm

packages/tools/kadena-cli/src/tx/tests/createTransaction.test copy.ts
barthuijgen: Oh huh, never even noticed that 🤣 thanks

nillo (approved):

PR #1712, opened 2024-02-23
[@kadena/graph] Subscriptions over WebSockets instead of SSE. Replaced ID return types with full Objects

MRVDH:
https://app.asana.com/0/1205801361945248/1206590017183507/f

### DIFF
--- a/packages/apps/graph/src/graph/subscription/new-blocks-from-depth.ts
+++ b/packages/apps/graph/src/graph/subscription/new-blocks-from-depth.ts
@@ -1,5 +1,5 @@
 import { prismaClient } from '@db/prisma-client';
-import { createID } from '@utils/global-id';
+import type { Block } from '@prisma/client';
 import { nullishOrEmpty } from '@utils/nullish-or-empty';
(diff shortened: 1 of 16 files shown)


packages/apps/graph/src/graph/subscription/new-blocks-from-depth.ts
```diff
@@ -1,5 +1,5 @@
 import { prismaClient } from '@db/prisma-client';
-import { createID } from '@utils/global-id';
```
nil-amrutlal-dept, line 2: Maybe it was left on purpose, but the `createID` function is not being used anywhere

nil-amrutlal-dept:
Looks good 🙌🏽

nil-amrutlal-dept (approved):

packages/apps/graph/src/graph/subscription/new-blocks-from-depth.ts
MRVDH: Yep I want to keep it since we will need it again if we want to return an ID somewhere

PR #1717, opened 2024-02-23
Test/improve e2e turbo setup

Ghislain89:
This PR introduces two things:

- Restructure of the E2E test packages

In essence, we have the following scenario's:

1.  When a package has a dependency change, it should be built and tested again
2. When a dependency has a peer-dependency change, the package should be built and tested again
3. When the tests for the package change, the package should be retested, but does not need to be rebuilt.
4. When there are no dependency or peer dependency changes, the package does not need to be rebuilt or tested.

~To achieve this i've added a synthetic Turbo task (as described [here](https://turbo.build/repo/docs/core-concepts/monorepos/task-dependencies#dependencies-outside-of-a-task)) that should allow us to more accurately use Turbo's caching mechanism while maintaining the possibility to run tasks in parallel.~ Out of scope for now.

I've also restructured the E2E test packages so that they can directly depend on the package under test, e.g. `e2e-graph` depends on `graph`. If a change in the graph is made, it will force test execution and otherwise ignore.

This should improve build times as well as ensure we actually re-run tests when we need to.


sanderlooijenga (approved):

turbo.json
alber70g, line 1: We discussed to leave `topo` out for now, and see whether the dependency on subject-under-test fixes caching issues

alber70g (approved):

turbo.json
alber70g: And change `packages/e2e-tests` to `packages/e2e`

packages/e2e-tests/e2e-tools/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@kadena-dev/e2e-tools",
```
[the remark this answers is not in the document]
alber70g: For packages that have no dependants.
For `@kadena-dev/e2e-base` I would leave it like that, as it has dependants, wdyt?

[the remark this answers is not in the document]
Ghislain89: That's actually a good idea. Will do it like that!

turbo.json
Ghislain89: I updated the PR to reflect these changes.

PR #1719, opened 2024-02-23
[@kadena/graph], [@kadena/graph-client] Adapt Transactions to Domain Model (Feat)

nil-amrutlal-dept:
- Adjusted transactions to represent the domain model
- Adjusted the client to consume the new query format

``` graphql
        hash
        cmd {
          meta {
            chainId
            creationTime
            gasLimit
            gasPrice
            sender
            ttl
          }
          payload {
            ... on ExecPayload {
              data
              code
            }
            ... on ContPayload {
              data
              pactId
              proof
              rollback
              step
            }
          }
          networkId
          nonce
          signers {
            publicKey
            capabilities
          }
        }
        result {
          badResult
          continuation
          eventCount
          gas
          goodResult
          height
          logs
          metadata
          transactionId
        }
```

---
  - https://app.asana.com/0/0/1206591792021818


packages/apps/graph/src/graph/index.ts
```diff
@@ -6,14 +6,19 @@ import './objects/fungible-chain-account';
...
+import './objects/payload';
 import './objects/signer';
 import './objects/token';
 import './objects/token-info';
 import './objects/total-count';
-import './objects/transaction';
+// import './objects/transaction';
+import './objects/cmd';
+import './objects/transaction-result';
+import './objects/transactions';
```
MRVDH, line 21: I think we want to keep it named `transaction`, since everything else is singular as well

packages/apps/graph/generated-schema.graphql
```diff
@@ -47,6 +47,41 @@ type BlockTransactionsConnectionEdge {
   node: Transaction!
 }
 
+"""A transaction cmd."""
+type Cmd {
```
MRVDH, line 51: If I remember correctly, we wanted to already start using fully written out names, since that's where the domain will shift to in the future.

cmd -> command
exec -> execution
cont -> continuation

But please confirm @alber70g

MRVDH (approved):

packages/apps/graph/generated-schema.graphql
alber70g: Yes, lets do that. 👍

packages/apps/graph/generated-schema.graphql
```diff
@@ -239,6 +297,9 @@ type PageInfo {
   startCursor: String
 }
 
+"""The payload of a transaction."""
+union Payload = ContPayload | ExecPayload
```
alber70g, line 301: We might want to be a little more specific on this type name and call it `TransactionPayload` as in GraphQL it's hard to get around too generic names

packages/apps/graph/generated-schema.graphql
alber70g, line 51: Maybe we should make the type name for `Cmd` more specific to what it is in the context of the transaction. I think this would then be `TransactionCommand`

alber70g (approved):
Comments on naming. It's best to be as specific as possible on the type names, so there will not be a conflict with similar "things" in the future

PR #1720, opened 2024-02-23
[@kadena/graph] System Status Checker

MRVDH:
A succesful start:
image

In case of failures:
image


packages/apps/graph/src/services/systems-check.ts
```diff
@@ -0,0 +1,96 @@
...
+                  ),
+                  {
+                    preflight: false,
+                    signatureVerification: false,
+                  },
+                );
+              },
+            },
+            {
+              title: 'Checking if the mempool and reachable.',
```
MRVDH, line 80:
```suggestion
              title: 'Checking if the mempool is reachable.',
```

PR #1724, opened 2024-02-26
feat(docs): introduce new scripts to detect broken links + fix broken internal links

realdreamer:
---
  - https://app.asana.com/0/0/1206752931510119


lsgunnlsgunn (approved):
I kinda wish that the build didn't chop up sentences and create diffs just for line breaks (sigh).

PR #1740, opened 2024-02-27
feat(kadena-cli) option to delete single or all keys

nillo:
add support for deleting plain generated keys


barthuijgen (approved):

packages/tools/kadena-cli/src/keys/commands/keysDelete.ts
```diff
@@ -0,0 +1,101 @@
...
+    await services.filesystem.deleteFile(`${PLAIN_KEY_DIR}/${key}`);
+    return { success: true, data: {} };
+  } catch (error) {
+    return { success: false, errors: ['Failed to delete key'] };
+  }
+};
+
+export const deleteAllKeys = async (): Promise<CommandResult<{}>> => {
+  try {
+    await services.filesystem.deleteDirectory(PLAIN_KEY_DIR);
```
realdreamer, line 26: some how i feel it's okay to keep the directory and delete only files but i'm also not sure which one we wanna follow here. if we choose to delete directory then it makes sense to do it for all sub commands delete.

packages/tools/kadena-cli/src/keys/commands/keysDelete.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    log.debug('delete-keys:action', { config });
+    if (config.keyFiles === 'all') {
+      const result = await deleteAllKeys();
+      assertCommandError(result);
+      log.info(log.color.green('\nAll keys have been deleted.\n'));
+    } else {
+      if (config.keyFiles === null) {
+        throw new Error(`Key: ${config.keyFiles} does not exist.`);
```
realdreamer, line 87:
```suggestion
        throw new Error(`Key: ${config.keyFiles} is invalid`);
```

packages/tools/kadena-cli/src/keys/commands/keysDelete.ts
nillo: it could be done, but it will always check for the keys folder existence anyways, so it will always be created, so i see no value in keeping the empty dir?

packages/tools/kadena-cli/src/keys/commands/keysDelete.ts
nillo: why is invalid?

realdreamer: here the input is invalid because it's null.

PR #1742, opened 2024-02-27
[@kadena/graph] Naming consistency and chainId as stringList

MRVDH:
* All magic strings have been replaced with their types/objects.
* ChainIds are now defined everywhere as string lists

https://app.asana.com/0/1205801361945248/1206643280157055/f


packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -10,11 +11,10 @@ import type { Guard } from '../types/graphql-types';
 import { FungibleChainAccountName } from '../types/graphql-types';
 import FungibleChainAccount from './fungible-chain-account';
 
-export default builder.prismaNode('Block', {
+export default builder.prismaNode(Prisma.ModelName.Block, {
```
nil-amrutlal-dept, line 14: Nicee

nil-amrutlal-dept (approved):

PR #1744, opened 2024-02-27
Feat/kadena cli/txlist

nillo:
added listing of transactions


barthuijgen (approved):

packages/tools/kadena-cli/src/tx/utils/txDisplayHelper.ts
```diff
@@ -1,12 +1,33 @@
...
+
+  for (const key of transactions) {
+    rows.push([key ?? 'N/A', key.includes('signed') === true ? 'Yes' : 'No']);
+  }
+
+  if (rows.length > 0) {
+    log.output(log.generateTableString(header, rows));
+  } else {
+    log.info('No transactions found');
+  }
```
realdreamer, line 28: this `else` part seems redundant since you're already checking the empty `transactions` length above

PR #1752, opened 2024-02-27
Chore/build esm version of client utils + lazy load key sources in dev-wallet

javadkh2:
importing cjs with vite causes some issues (in dev mode) so I have to updated the build script of client utils to generate esm output as well


packages/libs/client-utils/package.json
```diff
@@ -15,58 +15,52 @@
   ],
   "exports": {
     ".": {
-      "import": "./lib/index.js",
-      "require": "./lib/index.js",
-      "default": "./lib/index.js",
-      "types": "./lib/index.d.ts"
+      "import": "./lib/esm/index.js",
+      "require": "./lib/cjs/index.js",
+      "types": "./lib/esm/index.d.ts"
```
alber70g, line 20: I have read somewhere that "types" should be first in this object to get it to work.

javadkh2: interesting, I'll try it

ferreroltd (approved):

packages/libs/client-utils/package.json
javadkh2: it works regardless of the "types" place, but maybe there is an edge case. I will move it to top, its harmless

PR #1754, opened 2024-02-27
Document database compaction (part 1-sqllite) using the chainweb-node cwtool compact command

lsgunnlsgunn:
---
  - https://app.asana.com/0/0/1205972211842104


packages/apps/docs/src/docs/chainweb/compact-a-database.md
```diff
@@ -0,0 +1,99 @@
...
+   
+   For example, if you run the node in a Docker container, open a terminal in the container.
+   If you installed chainweb-node from a release binary or built it from source, open a terminal or secure shell on the computer where the binary is installed.
+
+   For information about installing chainweb-node from a release binary or building it from source code, see the [chainweb-node README](https://github.com/kadena-io/chainweb-node#README).
+
+2. Verify that you have access to the `cwtool`  command-line program by running the following command:
+   
+   ```bash
+   cwtool
```
chessai, line 32:
```suggestion
   cwtool --help
```

packages/apps/docs/src/docs/chainweb/compact-a-database.md
```diff
@@ -0,0 +1,99 @@
...
+   ```
+
+3. Create a backup copy of your current pact `sqlite` database to save all current state by running a command similar to the following:
+   
+   ```bash
+   cp -r /data/state/chainweb/db/0/sqlite /tmp/sqlite-backup-dd-mm-yyyy
+   ```
+   
+   Creating a backup copy of the database requires you to have more disk space available to store both the active database and the backup until you complete the compaction. 
+   However, creating a backup ensures that the node continues to run uninterrupted while the database is being compacted.
```
chessai, line 73: The other benefit, which is also very important, is that if something goes wrong, you didn't ruin your only copy.

packages/apps/docs/src/docs/chainweb/compact-a-database.md
```diff
@@ -0,0 +1,99 @@
...
+   ```
+   
+   Creating a backup copy of the database requires you to have more disk space available to store both the active database and the backup until you complete the compaction. 
+   However, creating a backup ensures that the node continues to run uninterrupted while the database is being compacted. 
+
+4. Compact your backup `sqlite` database by running the `cwtool compact` command with the following arguments:
+   
+   - `--target-blockheight` to keep enough state available to continue validating blocks after the database has been compacted. An appropriate value depends on the network your node is connected to. For example, if you are compacting your local development netowrk for testing purposes, you would set a lower block height than a node connected to the Kadena test network or main network. If this parameter isn't computed automatically, you should specify a value of at least 3.5 million (”3500000”) for a node connected to the Kadena main network.
+   - `--pact-database-dir` to specify the path to the backup pact `sqlite` state where all of the `*.sqlite`  files are located. The `cwtool compact` command creates one log file for each chain in this directory.
```
chessai, line 78: The wording about the log files I find confusing. At first I read it to mean that the log files would go in this directory.  I realise now that's not what it means, but I don't really think logging is worth mentioning in this paragraph.

packages/apps/docs/src/docs/chainweb/compact-a-database.md
```diff
@@ -0,0 +1,99 @@
...
+   - `--log-dir` to specify the directory where you want `cwtool compact` to put the log files it creates. If the directory doesn’t exist, `cwtool compact` creates it. These logs can be useful if something goes wrong.
```
chessai, line 79:
```suggestion
   - `--log-dir` to specify the directory where you want `cwtool compact` to put the log files it creates (one per chain). If the directory doesn’t exist, `cwtool compact` creates it. These logs can be useful if something goes wrong.
```

packages/apps/docs/src/docs/chainweb/compact-a-database.md
```diff
@@ -0,0 +1,99 @@
...
+
+5. Stop your node, replace the original pact `sqlite` state with the newly compacted `sqlite` state from the backup by running a command similar to the following:
+   
+   ```bash
+   mv /tmp/sqlite-backup-dd-mm-yyyy /data/state/chainweb/db/0/sqlite
+   ```
+
+6. Restart your node. 
+   
+   Your node should start normally and continue running with the reduced database size as though nothing has changed.
```
chessai, line 97: I think there should be a recommendation to keep around the old state for a while in case something goes wrong. It can be compressed, though, to save space in the meantime.

chessai (changes requested):

packages/apps/docs/src/docs/chainweb/compact-a-database.md
lsgunnlsgunn: to put the log files it creates, one for each chain.

chessai (approved):

sstraatemans (approved):

realdreamer (approved):

PR #1757, opened 2024-02-28
refactor(cli): move errors from helper to action

realdreamer:
- Removed the directory not exists error from `getAllAccounts` helper method and moved to specific actions. Because this was making an prompt to thrown an error.
- Made account details output as tabular data

### DIFF
--- a/packages/tools/kadena-cli/src/account/commands/accountFund.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountFund.ts
@@ -9,9 +9,11 @@ import ora from 'ora';
 import { assertCommandError } from '../../utils/command.util.js';
 import { createCommand } from '../../utils/createCommand.js';
 // import { createOption } from '../../utils/createOption.js';
+import { NO_ACCOUNTS_FOUND_ERROR_MESSAGE } from '../../constants/account.js';
 import { globalOptions } from '../../utils/globalOptions.js';
 import { log } from '../../utils/logger.js';
 import { accountOptions } from '../accountOptions.js';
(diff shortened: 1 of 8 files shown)


packages/tools/kadena-cli/src/constants/account.ts
```diff
@@ -13,5 +13,5 @@ export const NAMESPACES_MAP: { [key: string]: string } = {
   testnet04: NAMESPACES.TEST_NET,
 };
 
-export const NO_ACCOUNT_ERROR_MESSAGE =
-  'No valid accounts found. To create an account use `account add-manual` or `account add-from-wallet` command.';
+export const NO_ACCOUNTS_FOUND_ERROR_MESSAGE =
```
nillo, line 16: Looking at this we might want to consider moving all error messages to CONSTANTS

nillo (approved):

packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -34,6 +36,11 @@ export const createAccountFundCommand = createCommand(
     // deployDevnet(),
   ],
   async (option) => {
+    const allAccounts = await getAllAccounts();
+    if (allAccounts.length === 0) {
```
barthuijgen, line 40: seems a bit odd to get all accounts just to check if there are any. Wasn't there some `ensureAccountsExists` method or something that was used inside `getAllAccounts()` before?

realdreamer: yes, makes sense! updated

barthuijgen (approved):

PR #1761, opened 2024-02-28
Feat/initial layout profiles

ferreroltd:
This creates an inital setup for the select profiles page


packages/apps/dev-wallet/src/pages/select-profile/select-profile.tsx
```diff
@@ -11,15 +14,46 @@ export function SelectProfile() {
...
+          {profileList.map((profile) => {
+            return (
+              <div key={profile.uuid}>
+                <Link
+                  to={`/unlock-profile/${profile.uuid}`}
+                  style={{ textDecoration: 'none' }}
+                >
+                  <div
+                    className={cardClass}
+                    style={{ backgroundColor: idToColor(profile.uuid) }}
```
javadkh2, line 33: I would say instead of having this call here, we can do it in profile creation. and then store it in database as profile.accentColor. then we use it here. later we let users to pick another color themselves.

packages/apps/dev-wallet/src/utils/id-to-color.ts
```diff
@@ -0,0 +1,15 @@
+export function idToColor(id: string) {
```
javadkh2, line 1: good for now but we might refine this approach later for example maybe designers prefer to only use a limited list of colors instead of creating a random one

javadkh2 (changes requested):
Add `color` to the `IProfile` interface and set the value in profile creation.
update the `createProfile` function from `useWallet` to accept color.

packages/apps/dev-wallet/src/pages/select-profile/select-profile.tsx
```diff
@@ -11,15 +14,46 @@ export function SelectProfile() {
...
+                  >
+                    <div className={imgClass}>
+                      <InitialsAvatar name={profile.name} />
```
javadkh2, line 36: need to set a color here in contrast with background

PR #1762, opened 2024-02-28
[@kadena/graph] Account by public key 2.0 (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206580767959293


nil-amrutlal-dept:
I was doing some testing and I can see that by including the possibility of code being `coin.create-account` we sometimes we results like this from the selection (selected only `to_acct` and `amount`:

``` 
[
  { to_acct: 'one-1707914783471', amount: 100 },
  { to_acct: 'one-1707903605959', amount: 100 },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  {
    to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
    amount: 100
  },
  { to_acct: 'utils-one-1707914783469', amount: 1 },
  { to_acct: 'utils-one-1707903605959', amount: 1 },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000709
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000709
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000708
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000708
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000705
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000705
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000705
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000705
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.000007
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.000007
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.000007
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000427
  },
  {
    to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
    amount: 0.00000427
  }
]
```

nil-amrutlal-dept:
1. Distinct accounts
2. Get guards

nil-amrutlal-dept:
Small issue with this: a scenario on testnet. 
- Public key: `9be19442151c880492ec0fddc5bdbe9eccd243b8d723f4673b317f10b2e5d515`

This public key has 4 accounts. Each of them can potentially contain 20 chain accounts. For each chain account, we're calling the data loader. This can result in too many calls to the cw-node making the socket hang up

packages/apps/graph/src/utils/errors.ts
```diff
@@ -12,6 +12,7 @@ import { PrismaJsonColumnParsingError } from './prisma-json-columns';
  * Checks what type of error it is and returns a normalized GraphQLError with the correct type, message and a description that clearly translates to the user what the error means.
  */
 export function normalizeError(error: any): GraphQLError {
+  console.log(error);
```
MRVDH, line 15: do we need to keep this?

packages/apps/graph/src/services/chainweb-node/account-details.ts
```diff
@@ -21,6 +21,12 @@ export async function getAccountDetails(
 ): Promise<FungibleChainAccountDetails | null> {
   let result;
   try {
+    console.log(
```
MRVDH, line 24: do we need to keep this?

packages/apps/graph/src/graph/query/fungible-chain-account-by-public-key.ts
```diff
@@ -11,68 +12,67 @@ builder.queryField('fungibleChainAccountByPublicKey', (t) =>
...
 
-        return await getFungibleChainAccount({
-          chainId: args.chainId,
-          fungibleName: 'coin',
-          accountName,
-        });
+        const fungibleChainAccounts = (
+          await Promise.all(
+            accountNames.map(async (accountName: string) => {
+              return await getFungibleChainAccount({
```
MRVDH, line 31: this await can be removed

packages/apps/graph/src/graph/query/fungible-account-by-public-key.ts
```diff
@@ -13,83 +13,79 @@ builder.queryField('fungibleAccountByPublicKey', (t) =>
...
-        return {
-          __typename: FungibleAccountName,
-          accountName: accountName,
-          fungibleName: 'coin',
-          chainAccounts,
-          totalBalance: 0,
-          transactions: [],
-          transfers: [],
-        };
+        console.log('returning', accountsWithChainAccounts);
```
MRVDH, line 52: do we need to keep this?

packages/apps/graph/generated-schema.graphql
```diff
@@ -304,15 +304,15 @@ type Query {
   fungibleAccount(accountName: String!, fungibleName: String!): FungibleAccount
 
   """Retrieve an account by public key."""
-  fungibleAccountByPublicKey(publicKey: String!): FungibleAccount
+  fungibleAccountByPublicKey(publicKey: String!): [FungibleAccount!]
```
MRVDH, line 307: we should probably rename these to `...Accounts...`

nil-amrutlal-dept:
> * 9be19442151c880492ec0fddc5bdbe9eccd243b8d723f4673b317f10b2e5d515

This was solved by the recent update that implemented retries. Whenever we reach the rate limit, there's a small timeout followed by another call ensuring we have a much greater chance of getting the data we're looking for

nil-amrutlal-dept:
> Small issue with this: a scenario on testnet.
> 
> * Public key: `9be19442151c880492ec0fddc5bdbe9eccd243b8d723f4673b317f10b2e5d515`
> 
> This public key has 4 accounts. Each of them can potentially contain 20 chain accounts. For each chain account, we're calling the data loader. This can result in too many calls to the cw-node making the socket hang up

This was solved by the recent update that implemented retries. Whenever we reach the rate limit, there's a small timeout followed by another call ensuring we have a much greater chance of getting the data we're looking for

packages/apps/graph/generated-schema.graphql
```diff
@@ -384,11 +384,17 @@ type Query {
   """
   fungibleAccount(accountName: String!, fungibleName: String): FungibleAccount
 
+  """Retrieve an account by public key."""
+  fungibleAccountsByPublicKey(publicKey: String!): [FungibleAccount!]
```
MRVDH, line 388: Don't we want the fungiblename as a parameter here as well?

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const accountsFromTransactions = (await prismaClient.$queryRaw`
+    SELECT DISTINCT to_acct
+    FROM transfers AS tr
+    INNER JOIN transactions AS tx
+      ON tx.block = tr.block AND tx.requestkey = tr.requestkey
+    WHERE
+      tx.data::text LIKE ${searchPubKey}
+      AND
+        (tx.code LIKE '%coin.transfer-create%'
```
MRVDH, line 84: here you can then replace `coin` with the fungibleName

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,95 @@
...
+        OR tx.code LIKE '%coin.create-account%')
+  `) as { to_acct: string }[];
+
+  const result = accountsFromTransactions.map((account) => account.to_acct);
+
+  if (result.length === 0) {
+    return [];
+  }
+
+  return result;
```
MRVDH, line 94: if result.length === 0 then it's an empty array and you can just return result right away.

```suggestion
  return accountsFromTransactions.map((account) => account.to_acct);
```

MRVDH (changes requested):

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
nil-amrutlal-dept: You mean also add `fungibleName` as an argument?

packages/apps/graph/generated-schema.graphql
nil-amrutlal-dept: The thing with that is that a fungible can have its own code to create accounts. I don't think its guaranteed that we can find created account using `${$fungibleName}.transfer-create`.

Can you confirm this @alber70g

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
alber70g: We should probably look for both `transfer-create` and `create-account`

alber70g:
> was doing some testing and I can see that by including the possibility of code being `coin.create-account` we sometimes we results like this from the selection (selected only `to_acct` and `amount`:
> 
> ```
> [
>   { to_acct: 'one-1707914783471', amount: 100 },
>   { to_acct: 'one-1707903605959', amount: 100 },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   {
>     to_acct: 'k:5a2afbc4564b76b2c27ce5a644cab643c43663835ea0be22433b209d3351f937',
>     amount: 100
>   },
>   { to_acct: 'utils-one-1707914783469', amount: 1 },
>   { to_acct: 'utils-one-1707903605959', amount: 1 },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000709
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000709
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000708
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000708
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000705
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000705
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000705
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000705
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.000007
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.000007
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.000007
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000427
>   },
>   {
>     to_acct: 'k:f89ef46927f506c70b6a58fd322450a936311dc6ac91f4ec3d8ef949608dbf1f',
>     amount: 0.00000427
>   }
> ]
> ```
> 
> What is weird here is the `one-1707914783471` and other similar values. Should these be there? If not, do we have to apply any additional filtering? @alber70g

Yes, these are now called vanity-accounts. It's possible to do `(coin.create-account  "albert" (read-keyset 'ks))` and supply an arbitrary keyset. The `k:`, `w:` (and other)-accounts are so-called "principal" accounts, and require you to use a specific form of name that matches with the keyset that's given. This is to prevent squatting of accounts, either on other chains, or by monitoring the mempool

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
nil-amrutlal-dept: We do! In the file, the line after this one is `OR tx.code LIKE '%coin.create-account%')`

nil-amrutlal-dept:
I would like to discuss if we need to find a way to go around this nuance and if so, how it could be done.

Giving an example, (using key `0cef7a289b5fe19fac8f6ec86b7f7f4721b028e959edc81181dbf85860c753b9`.

image

Here we can see that by searching for this public key and using `fungibleName = coin` we get no results. However, if I do search without providing a `fungibleName`, we get this

image


What is happening here:
- The account QfQ07sSzeiP9N9AFnvosJ662Ykt1 was created through the transaction free.aos-token-n_vxsqc09.transfer-create.
- I am assuming it paid gas to `k:4ba39bd19b9f3ecae0f03375f9bbe370b66e0101f1ffce5dbff0f3e9f9290967`, resulting in having a transfer to this account
- The query returned this account due to meeting the provided conditions.


**Possible solution**
To verify if the segment preceding `.transfer-create` matches the transfer's `moduleName`. This approach ensures that accounts to which gas was paid are not included when dealing with non-coin fungible accounts.
@alber70g @MRVDH what do you think?

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (fungible) {
+    return (await prismaClient.$queryRaw`
+    SELECT DISTINCT to_acct AS accountname, moduleName as fungiblename
+    FROM transfers AS tr
+    INNER JOIN transactions AS tx
+      ON tx.block = tr.block AND tx.requestkey = tr.requestkey
+    WHERE
+    tx.data::text LIKE ${`%${publicKey}%`}
+      AND
+      (tx.code LIKE ${`%${fungible ?? ''}.transfer-create%`}
```
MRVDH, line 95: empty string should not be allowed here, it breaks the query. Besides that, we already check the fungible above in the `if (fungible) {`.

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,109 @@
...
+  fungible?: string,
+): Promise<{ accountname: string; fungiblename: string }[]> {
+  const regex = /^[a-zA-Z0-9]+$/;
+
+  if (publicKey.length !== 64 || !regex.test(publicKey)) {
+    throw new Error('Invalid public key');
+  }
+
+  // This is required because prisma queryRaw does not support string interpolation for entire query fragments.
+  if (fungible) {
```
MRVDH, line 86: Is there a reason we use raw queries here? Distinct selection is possible with Prisma: https://www.prisma.io/docs/orm/prisma-client/queries/aggregation-grouping-summarizing#select-distinct

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,109 @@
...
+      }),
+    },
+    type: [FungibleAccount],
+    nullable: true,
+    async resolve(__parent, args) {
+      try {
+        const accountAndFungibleNames =
+          await getAccountAndFungibleNamesByPublicKey(
+            args.publicKey,
+            args.fungibleName ?? undefined,
```
MRVDH, line 34: can't we do just this here since the function param already is defined as `fungible?`

```suggestion
            args.fungibleName,
```

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
```diff
@@ -0,0 +1,109 @@
...
+          );
+
+        if (accountAndFungibleNames.length === 0) {
+          return null;
```
MRVDH, line 38: In terms of consistency, for all the other queries that return arrays, do we return an empty array on no results or null?

MRVDH (commented):
Comments in that one file also apply to the other file.

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
nil-amrutlal-dept: Its this weird case where the the argument can be `string | undefined | null`. I will make the param of the function `fungible?: string | null` since it might be cleaner
image

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
nil-amrutlal-dept: Good catch. It is indeed an array so we should return an empty array and make the query non-nullable. Changes done 👍🏽

packages/apps/graph/src/graph/query/fungible-accounts-by-public-key.ts
nil-amrutlal-dept: We are using a raw query because it allows us to look for a `publicKey` in the transaction field `data::text`. Since this is a JSON field, when not using raw query we are obliged to provide a path. This path is dynamic and can be different from transaction to transaction

MRVDH (approved):

PR #1768, opened 2024-02-28
ag/chore/add spirekey example app

alber70g:
- chore(spirekey-example): initial commit
- chore(spirekey-example): add localstorage


packages/apps/spirekey-example/src/spirekey.ts
```diff
@@ -0,0 +1,167 @@
...
+  get transactions() {
+    return this._transactions;
+  }
+
+  update(location: Location) {
+    const searchParams = new URLSearchParams(location.search);
+
+    if (searchParams.has('user')) {
+      const userSearch = searchParams.get('user');
+      if (userSearch && userSearch?.length > 0) {
```
nillo, line 33: `userSearch?.length > 0 `is sufficient to check existence and non empty

packages/apps/spirekey-example/src/spirekey.ts
```diff
@@ -0,0 +1,167 @@
...
+          'retrieved transaction from querystring parameters',
+          JSON.stringify(transaction, null, 2),
+        );
+
+        this._transactions[transaction.hash] = transaction;
+      }
+    }
+    this._saveToLocalStorage();
+    // clear querystring parameters
+    history.pushState({}, '', location.pathname);
```
nillo, line 56: this will break backbutton navigation if any

packages/apps/spirekey-example/src/spirekey.ts
```diff
@@ -0,0 +1,167 @@
...
+        // decode base64
+        this._user = JSON.parse(decodeBase64(userSearch));
+        console.log('retrieved user from querystring parameters', this._user);
+      }
+    }
+
+    if (searchParams.has('transaction')) {
+      const transactionSearch = searchParams.get('transaction');
+      if (transactionSearch && transactionSearch?.length > 0) {
+        const transaction: IUnsignedCommand = JSON.parse(
```
nillo, line 43: what if parsing fails ( for all parses btw)

packages/apps/spirekey-example/src/spirekey.ts
```diff
@@ -0,0 +1,167 @@
...
+      transaction: encodeBase64(JSON.stringify(tx)),
+      returnUrl: this._returnUrl,
+      optimistic: false,
+    };
+    const queryString = new URLSearchParams(query as any).toString();
+    location.href = `${this._spireKeyHostname}/sign?${queryString}`;
+  }
+
+  private _saveToLocalStorage() {
+    sessionStorage.setItem('user', JSON.stringify(this._user));
```
nillo, line 79: u say save to localStorage but you store to session storage?
- session storage is cleared when closing browser / tab ( do we want that? )
- session storage can be accessed by other scripts... (risk)

packages/apps/spirekey-example/src/spirekey.ts
```diff
@@ -0,0 +1,167 @@
+import { IUnsignedCommand } from '@kadena/client';
+
+export class KadenaSpireKey {
+  private _returnUrl: string;
+  private _spireKeyHostname: string;
+  private _user: IUser | null = null;
+  private _transactions: Record<string, IUnsignedCommand> = {};
+
```
nillo, line 8:
```
  private _storage: Storage;
  private _location: Location;
  private _history: History;
```

As abstractions?

  ```
constructor(spireKeyHostname: string, returnUrl: string, storage: Storage, location: Location, history: History) {
    this._spireKeyHostname = spireKeyHostname;
    this._returnUrl = returnUrl;
    this._storage = storage;
    this._location = location;
    this._history = history;
    this._loadFromLocalStorage();
    this.update(this._location);
  }
```

for easier testing

packages/apps/spirekey-example/src/spirekey.ts
Randynamic, line 1: I really like this should we move this into the SpireKey or wallet repo and release it as a separate lib

Randynamic (approved):

packages/apps/spirekey-example/src/spirekey.ts
alber70g: LocalStorage can be accessed by other scripts as well, right?

packages/apps/spirekey-example/src/spirekey.ts
alber70g: I want to remove the querystring parrameters as they're loaded. Otherwise it'll get into weird states

packages/apps/spirekey-example/src/spirekey.ts
alber70g: Yes, but typing doesn't accept that as when `userSearch === null` `userSearch.length = == false` and `false > 0` is not a valid statement

PR #1776, opened 2024-02-29
fix(pou): small fixes

sstraatemans:
fixes:
- when you first get on the page (when scanned QR). after you log in, you went to frontpage. now you go to the page you scanned. so you can continue.

- when you made a connection and start a new one, you used to see old info. now it starts over as expected

-


Ghislain89 (approved):

Ghislain89:
tested, works

PR #1779, opened 2024-03-04
chore(client-utils): esm

javadkh2:
using unbuild to create correct esm modules

### DIFF
--- a/packages/libs/client-utils/build.config.ts
+++ b/packages/libs/client-utils/build.config.ts
@@ -1,24 +1,4 @@
-import { readdirSync } from 'fs';
-import { BuildEntry, defineBuildConfig } from 'unbuild';
-
-const entries: BuildEntry[] = readdirSync('./src').flatMap((entry) => {
-  return [
-    {
-      builder: 'mkdist',
-      input: `./src/${entry}`,
-      outDir: `./dist/${entry}/esm`,
-      cleanDist: true,
-      format: 'esm',
-    },
-    {
-      builder: 'mkdist',
-      input: `./src/${entry}`,
-      outDir: `./dist/${entry}/cjs`,
-      cleanDist: true,
-      format: 'cjs',
-    },
-  ];
-});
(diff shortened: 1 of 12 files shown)


packages/libs/client-utils/build.config.ts
alber70g, line 1: Does this build also do type checking etc, or is it just transpiling?

packages/libs/client-utils/etc/client-utils-core.api.md
```diff
@@ -75,13 +75,13 @@ export const crossChainClient: <T = PactValue>(args_0: IClientConfig, args_1?: I
     data: ICommandResult;
     }, {
     event: "spv-proof";
-    data: {
+    data: Partial<{
     pactId: string;
     step: number;
     rollback: boolean;
     data?: Record<string, unknown> | undefined;
     proof?: string | null | undefined;
```
alber70g, line 83: If partial is used, we don't need to specify `?` or `|undefined`, correct?

Why is this partial? It's fairly sure what `spv-proof` should look like?

alber70g (approved):
Some comments

packages/libs/client-utils/build.config.ts
javadkh2: its a good question, I will check it

javadkh2: It didn't work and I couldn't find any confing for unbuild to enable it so I have added tsc with "noEmit" for type checking.

packages/libs/client-utils/etc/client-utils-core.api.md
javadkh2: I have addressed this

alber70g (approved):
LGTM

packages/libs/client-utils/build.config.ts
salamaashoush: it runs tsc to generate declaration files since there is no other way, so it would catch some types error if you that's what you mean

barthuijgen (approved):

PR #1789, opened 2024-03-05
fix(cli): networks empty + tx list on unsigned

realdreamer:
- Ensure networks configuration for non-existent networks directory or networks directory empty
- List only unsigned transactions on - `tx sign`

### DIFF
--- a/packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
+++ b/packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
@@ -62,12 +62,14 @@ export function getSignersStatus(
 /**
  * Retrieves all transaction file names from the transaction directory based on the signature status.
  * @param {boolean} signed - Whether to retrieve signed or unsigned transactions.
+ * @param {boolean} all - Whether to retrieve all transactions (signed and unsigned).
  * @returns {Promise<string[]>} A promise that resolves to an array of transaction file names.
  * @throws Throws an error if reading the transaction directory fails.
  */
 export async function getTransactions(
   signed: boolean,
   directory: string,
+  all: boolean = false,
 ): Promise<string[]> {
   try {
     const files = await services.filesystem.readDir(directory);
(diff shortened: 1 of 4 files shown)


packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -62,12 +62,14 @@ export function getSignersStatus(
...
  * Retrieves all transaction file names from the transaction directory based on the signature status.
  * @param {boolean} signed - Whether to retrieve signed or unsigned transactions.
+ * @param {boolean} all - Whether to retrieve all transactions (signed and unsigned).
  * @returns {Promise<string[]>} A promise that resolves to an array of transaction file names.
  * @throws Throws an error if reading the transaction directory fails.
  */
 export async function getTransactions(
   signed: boolean,
   directory: string,
+  all: boolean = false,
```
realdreamer, line 72: I thought make it as a short term fix for now! maybe will fix it properly after the demo

packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -81,7 +83,15 @@ export async function getTransactions(
           if (content === null) return null;
           // signed=false can still return already signed transactions
           const schema = signed ? ICommandSchema : IUnsignedCommandSchema;
-          const parsed = schema.safeParse(JSON.parse(content));
+          const JSONParsedContent = JSON.parse(content) as
+            | ICommand
+            | IUnsignedCommand;
+          const parsed = schema.safeParse(JSONParsedContent);
+          if (parsed.success && !signed && !all) {
+            const isSigned = JSONParsedContent.sigs.every((sig) => !!sig);
```
barthuijgen, line 91: would prefer to just introduce a 3rd schema variant to have `ICommandSchema` and `IUnsignedCommandSchema`. The problem now was that `IUnsignedCommandSchema` also accepts signed commands.

If it's in a rush I can accept this for now.

barthuijgen (approved):

PR #1796, opened 2024-03-05
Feat/dev wallet sig builder

javadkh2:


packages/apps/dev-wallet/src/modules/wallet/wallet.service.ts
```diff
@@ -58,6 +63,21 @@ export function sign(
...
+export function getRelevantKeys(
+  keySources: IKeySource[],
+  Tx: IUnsignedCommand,
+) {
+  return keySources.flatMap((keySource) => {
+    const { keys } = keySource;
+    const cmd: IPactCommand = JSON.parse(Tx.cmd);
+    const publicKeys = cmd.signers
+      .map((signer) => keys.find((key) => key.publicKey === signer.pubKey))
+      .filter((index) => index !== undefined) as Array<IKeyItem>;
```
ferreroltd, line 75: why do we need the cast here?

packages/apps/dev-wallet/src/pages/sig-builder/style.css.ts
```diff
@@ -0,0 +1,6 @@
+import { style } from '@vanilla-extract/css';
+
+export const codeArea = style({
+  width: `100%`,
```
ferreroltd, line 4: do we need this one? could be solved with flexing the parent, i can take a look in another pr

ferreroltd (approved):

packages/apps/dev-wallet/src/pages/sig-builder/style.css.ts
javadkh2: I just focused on functionality in this PR
the whole page need to be refactored. its better to do it when we have design for it.

packages/apps/dev-wallet/src/modules/wallet/wallet.service.ts
javadkh2: TS is not smart enough to grasp the code filters out undefined items, the type it infers is  `Array<IKeyItem | undefined>` so I cast it explicitly to the correct type `Array<IKeyItem>`

alber70g (approved):

PR #1805, opened 2024-03-07
fix(pou): use graph on dashboard

sstraatemans:
use graphql data for the dashboard. instead of relying on firebase.
we want to use firebase as less as possible. and rely on data from chain.

https://app.asana.com/0/1206452987729603/1206749540780134/f


packages/apps/proof-of-us/src/utils/getIPFSLink.ts
```diff
@@ -2,10 +2,10 @@ import { env } from './env';
```
sstraatemans, line 2: removing all the extra data in `proofs`.
this was needed for having a leaderboard

packages/apps/proof-of-us/src/utils/socket/store.ts
```diff
@@ -40,43 +40,15 @@ const ProofOfUsStore = () => {
       signees: [{ ...account, signerStatus: 'init', initiator: true }],
```
sstraatemans, line 40: removing all the extra data in proofs.
this was needed for having a leaderboard

packages/apps/proof-of-us/src/components/ConnectView/ConnectView.tsx
```diff
@@ -18,10 +21,22 @@ interface IProps {
 export const ConnectView: FC<IProps> = ({ proofOfUs }) => {
   const { signToken } = useSignToken();
 
+  const router = useRouter();
+
   const handleJoin = async () => {
     signToken();
   };
 
+  useEffect(() => {
```
sstraatemans, line 30: we wait for the minting on that page. so we always have the same loading state

Ghislain89 (approved):

PR #1808, opened 2024-03-07
Feat/add confirmation depth to pollStatus

javadkh2:
`/poll` endpoint accepts `confirmationDepth` and returns the result if the request meets that length. It means waiting for a certain amount of blocks, then returning the result of the request. It's useful when you need to make sure the data is persisted permanently on the blockchain.

In this PR, we will add the support in the client library.

You can pass it to the `createClient` function to use it as the value on applicable functions.

```
const client = createClient("http://host-url", { confirmationDepth: 5 })
```

You can also pass it with poll options for the `pollStatus` function (which overrides the one you passed to the `createClient`).

```
const result = pollStatus(request, { confirmationDepth: 5 })
```

### DIFF
--- a/packages/libs/chainweb-node-client/src/poll.ts
+++ b/packages/libs/chainweb-node-client/src/poll.ts
@@ -17,10 +17,16 @@ import { stringifyAndMakePOSTRequest } from './stringifyAndMakePOSTRequest';
 export async function poll(
   requestBody: IPollRequestBody,
   apiHost: string,
+  confirmationDepth = 0,
 ): Promise<IPollResponse> {
   const request = stringifyAndMakePOSTRequest(requestBody);
   const pollUrl = new URL(`${apiHost}/api/v1/poll`);
-
+  if (confirmationDepth > 0) {
+    pollUrl.searchParams.append(
+      'confirmationDepth',
+      confirmationDepth.toString(),
+    );
+  }
   const response = await fetch(pollUrl.toString(), request);
   return parseResponse<IPollResponse>(response);
 }
--- a/packages/libs/client/etc/client.api.md
+++ b/packages/libs/client/etc/client.api.md
@@ -106,6 +106,7 @@ export interface IClient extends IBaseClient {
     dirtyRead: (transaction: IUnsignedCommand) => Promise<ICommandResult>;
     // @deprecated
     getPoll: (transactionDescriptors: ITransactionDescriptor[] | ITransactionDescriptor) => Promise<IPollResponse>;
+    pollOne: (transactionDescriptor: ITransactionDescriptor, options?: IPollOptions) => Promise<ICommandResult>;
     preflight: (transaction: ICommand | IUnsignedCommand) => Promise<ILocalCommandResult>;
     runPact: (code: string, data: Record<string, unknown>, option: INetworkOptions) => Promise<ICommandResult>;
     // @deprecated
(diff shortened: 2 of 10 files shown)


packages/libs/chainweb-node-client/src/poll.ts
```diff
@@ -17,10 +17,16 @@ import { stringifyAndMakePOSTRequest } from './stringifyAndMakePOSTRequest';
 export async function poll(
   requestBody: IPollRequestBody,
   apiHost: string,
+  confirmationDepth = 0,
```
EnoF, line 20: we should probably add it to 3 or 5, I forgot what the common consensus is on this. @Takadenoshi ?

packages/libs/client/src/client/client.ts
```diff
@@ -239,7 +252,7 @@ export interface ICreateClient {
    * Useful when you are working with a single network and chainId.
    * @param hostUrl - the URL to use in the client
    */
-  (hostUrl: string): IClient;
+  (hostUrl: string, defaults?: { confirmationDepth?: number }): IClient;
```
eileenmguo, line 255: Can we add some jsdoc for defaults to make it really clear what is to be expected when passing this value? I assume when you createClient with certain defaults it is applied to all applicable the functions returned

packages/libs/chainweb-node-client/src/poll.ts
javadkh2: What is the default value in the server?
What if I just make this optional

packages/libs/client/src/client/client.ts
javadkh2: I will add the docs. yes in this level it is applied to all applicable functions

packages/libs/client/etc/client.api.md
```diff
@@ -106,6 +106,7 @@ export interface IClient extends IBaseClient {
     dirtyRead: (transaction: IUnsignedCommand) => Promise<ICommandResult>;
     // @deprecated
     getPoll: (transactionDescriptors: ITransactionDescriptor[] | ITransactionDescriptor) => Promise<IPollResponse>;
+    pollOne: (transactionDescriptor: ITransactionDescriptor, options?: IPollOptions) => Promise<ICommandResult>;
```
alber70g, line 109: Can you add a message for the changelog to indicate that we also added an alternative to poll that can be used for a single transaction?

packages/libs/client/etc/client.api.md
```diff
@@ -432,6 +443,11 @@ export class Literal {
 // @public
 export const literal: (value: string) => Literal;
 
+// @alpha (undocumented)
+export type Milliseconds = number & {
```
alber70g, line 447: Is this something that we want and use more often for primitives that are used in a specific context?

I'd like to talk about this a little bit, as there are various ways to do this. Maybe we want to use symbols?

Does a standard exist in the community that we can follow?

packages/libs/client/src/client/client.ts
```diff
@@ -238,20 +253,23 @@ export interface ICreateClient {
    *
    * Useful when you are working with a single network and chainId.
    * @param hostUrl - the URL to use in the client
+   * @param defaults - default options for the client it includes confirmationDepth that is used for polling
    */
-  (hostUrl: string): IClient;
+  (hostUrl: string, defaults?: { confirmationDepth?: Milliseconds }): IClient;
```
alber70g, line 258: Is this a mistake? Should this be number?

alber70g (commented):
Some questions

packages/libs/client/etc/client.api.md
javadkh2: Yes using `brand` is a kind of standard in TS community. I made that optional here, so it works mostly like a hint to user.
Since number is very generic having this alias will make the type more clear.

packages/libs/client/src/client/client.ts
javadkh2: yes will fix it

alber70g (approved):

PR #1811, opened 2024-03-07
fix(account): fund command additional feedback for sending transactions + check

realdreamer:
- Print request key and explorer url
- Improve help text after funding an account
- Before sending transactions to chain validate using `local` (for account fund and tx send commands)

---
  - https://app.asana.com/0/0/1206794459798420
  - https://app.asana.com/0/0/1206794459798425

### DIFF
--- a/packages/tools/kadena-cli/src/account/commands/accountFund.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountFund.ts
@@ -130,9 +130,13 @@ export const createAccountFundCommand = createCommand(
     const result = await fund(config);
     assertCommandError(result);
 
+    const explorerURL = networkConfig.networkExplorerUrl.endsWith('/')
+      ? networkConfig.networkExplorerUrl
+      : `${networkConfig.networkExplorerUrl}/`;
+
     log.info(
       log.color.green(
-        `Explorer URL: ${networkConfig.networkExplorerUrl}\nRequest Key: ${result.data.requestKey}`,
+        `Transaction explorer URL: ${explorerURL}${result.data.requestKey}`,
       ),
     );
     const { pollStatus } = createClient(
--- a/packages/tools/kadena-cli/src/account/commands/accountFund.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountFund.ts
@@ -1,5 +1,6 @@
 // import { describeModule } from '@kadena/client-utils/built-in';
 // import { Option } from 'commander';
+import { createClient } from '@kadena/client';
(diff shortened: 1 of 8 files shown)


packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -126,15 +127,40 @@ export const createAccountFundCommand = createCommand(
...
-    const loader = ora('Funding account...\n').start();
-
     const result = await fund(config);
-    assertCommandError(result, loader);
+    assertCommandError(result);
 
     log.info(
       log.color.green(
-        `"${accountConfig.name}" account funded with "${amount}" ${accountConfig.fungible} on chain ${chainId} in ${networkConfig.networkId} network.`,
+        `Explorer URL: ${networkConfig.networkExplorerUrl}\nRequest Key: ${result.data.requestKey}`,
```
barthuijgen, line 135: I think this url should be including the request key:

`https://explorer.chainweb.com/testnet/tx/${result.data.requestKey}`

packages/tools/kadena-cli/src/account/utils/createAndTransferFunds.ts
```diff
@@ -78,15 +78,16 @@ export async function createAndTransferFund({
...
     }
 
-    const { submit, listen } = createClient(
+    const { submit, local } = createClient(
       `${networkConfig.networkHost}/chainweb/0.0/${networkConfig.networkId}/chain/${chainId}/pact`,
     );
 
-    const transactionDescriptor = await submit(signedTx);
+    // Validate the transaction locally before sending it to the network
+    await local(signedTx);
```
barthuijgen, line 86: does `local()` throw if something is wrong? or should the result be actually validated?

realdreamer: yes, it does throws error but for some use cases needs to be validated. so added that check as well.

packages/tools/kadena-cli/src/tx/commands/txSend.ts
```diff
@@ -139,6 +139,11 @@ export const sendTransactionAction = async ({
 
       const client = getClient(details);
 
+      const localResponse = await client.local(command);
+      if (localResponse.result.status === 'failure') {
+        throw localResponse.result.error;
```
nillo, line 144: throw a command error (check)

barthuijgen (approved):

nillo (approved):

nillo (approved):

PR #1812, opened 2024-03-07
Add What is Marmalade and an updated Mint a basic NFT tutorial

lsgunnlsgunn:
Removed the content from the misleading "Marmalade tutorial" topic.
Added "What is Marmalade" and "Get started with Marmalade" topics.
Updated the config.yaml to include the new files.
Updated the Marmalade landing page to not just repeat the same content and links.


lsgunnlsgunn:
@ggobugi27 @jermaine150
I made some changes to the current Marmalade doc (more are coming in Docs 2.0, but this is a start).
Please take a look, and add your comments when you get a chance.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+- Transfers NFT1 to an escrow account associated with a specific sale identifier—`sale:1`—in the token ledger. 
+
+![Bob starts a sale by offering NFT1](/assets/marmalade/offer.png)
+
+By initiating the sale, Bob loses custody of the asset for a period of time.
+The `sale:1` pact guard takes ownership of the asset.
+
+### Buy
+
+The buyer who wants to acquire the asset that's been offered for sale sends a **continuation message** for the identifier assigned to the SALE event.
```
ggobugi27, line 122: I'd suggest "continuation transaction" over "continuation message"

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+
+By initiating the sale, Bob loses custody of the asset for a period of time.
+The `sale:1` pact guard takes ownership of the asset.
+
+### Buy
+
+The buyer who wants to acquire the asset that's been offered for sale sends a **continuation message** for the identifier assigned to the SALE event. 
+
+For example, Alice sends a continuation message for the `sale:1` pact identifier to transfer NFT1 out of the escrow account in the token ledger and into her account. 
+The token policy then debits the sale price of 50.0 from Alice’s KDA account to credit Bob and executes any other offsetting transaction needed to release the escrowed NFT.
```
ggobugi27, line 125: It'd be the Marmalade's ledger that debits the price, to be exact.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+For example, the token policy pays any royalty fees or returns any funds as specified in the policy. 
+
+![Alice continues the execution of the sale pact](/assets/marmalade/buy.png)
+
+With this separation of duties, the `sale` pact is responsible for transferring the NFT out of escrow and enforcing the timeout.
+Any other transaction requirements are handled by token policies.
+
+### Rollback and withdraw
```
ggobugi27, line 133: We can simply call it "Withdraw", to be consistent with above? 
Rollback is more of a Pact term.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+
+The timeout captured in the **offer** step allows a seller to withdraw from the sale if there are no interested buyers for the token.
```
ggobugi27, line 135: One note we can also add is that timeout is optional. Other than timeout, sale-contract can also control the withdraw.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+Before you can mint a token on the blockchain, you must create the digital item and upload it to at least one accessible location such as InterPlanetary File System (IPFS), Arweave, or another content delivery or hosting service.
+
+After you have a uniform resource identifier (URI) for locating the digital item, you must also create a metadata file using the JSON schema to record the name, location, and other properties about the item to be minted.
+You can then create and mint the token for the digital item, generating an on-chain token identifier with a `t:` prefix. 
+The token identifier includes the hash for the metadata that describes the token to ensure that a digital item can be cryptographically verified to be the actual NFT. 
+
+Depending on how you apply or configure token policies, you might be the sole owner of the token you mint or share ownership with others. 
+If you're the owner of a token, you have the option to offer the token for sale or transfer the token to another owner. 
+
+## Token sales and trustless escrow
```
ggobugi27, line 93: In this section we don't exactly describe the sale contract's role. The sale can b e "quoted" sale or "non-quoted" sale, and when the seller uses the "quoted-sale", then they need to choose the "sale-contract". The "sale-contract" then has the ability to update the final price (such as the final price of the auction) or block withdraw until a condition is met. If the seller does not choose a quoted sale, sale-contract is not needed. Escrow part applies to any kind of sale, non-quoted or quoted. I think we can assume users are using quoted sale in this section, and briefly add the role of sale-contract.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+- Use the **royalty policy** to specify the account for depositing royalty payments from sales.
+- Use the **collection policy** to create token collections for a marketplace.
+
+To illustrate how you can use these policies, assume that you want to offer a non-fungible token.
+You want to ensure that you are the only person authorized to mint the token and that only one token can be minted and owned.
+You also want to require that the token can only be transferred to a new owner through a sale contract and you'll be paid a royalty any time the token ownership changes.
+
+To accomplish this, you can create the token using the `marmalade-v2.guard-policy-v1`,`marmalade-v2.non-fungible-policy-v1`, and `marmalade-v2.royalty-policy-v1` policies.
+
+These policies ensure that you are the only person authorized to mint the token, only one token can be minted, and any time the token is transferred to a new owner, you're paid the  royalty you specify.
```
ggobugi27, line 171: To be picky, adding guard-policy doesn't automatically add the mint-guard, but it requires the token creator to add "mint-guard" information at creation.

ggobugi27 (changes requested):
The newly added articles are very well put! Let me know if some of my comments require extra explanation!

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+
+In the simplest sense, a sale is typically a three-step process:
+
+1. A seller lists an item for sale with a price.
+
+2. A buyer agrees to pay the seller for the item at a mutually agreed-upon price.
+
+3. The seller receives payment and transfers the item to the buyer.
+
+On the Kadena network, Pact smart contracts use `pact` functions to define multi-step operations like cross-chain transfers and automatic gas payments.
```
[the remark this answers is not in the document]
lsgunnlsgunn: I'm trying to avoid using the terminology that way, at least in this introductory part, treating `pact` as a feature of the language instead of a unique entity unto itself.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -0,0 +1,180 @@
...
+All of the sequencing and cryptography required to enforce a sales contract is handled automatically by the Pact language itself.
+
+To complete a sale, the `pact` function defines the following steps:
+
+- Offer: A seller who wants to list a digital asset for sale creates an **offer**. 
+  This operation puts the asset into a _trustless escrow_ account that can only be released by later steps in the `pact` definition. 
+
+- Buy: A buyer initiates a **buy** operation with a promised payment and the escrowed asset waits for the step to be completed successfully by the buyer. 
+
+- Rollback: If the buyer doesn't complete the transaction within a certain period of time, the `pact` function can roll back the transaction and the seller can **withdraw** the sale.
```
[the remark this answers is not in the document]
lsgunnlsgunn: I went back and forth on which one to use. I guessed wrong 😁.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
lsgunnlsgunn: I'm not sure I captured this in my revision.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
lsgunnlsgunn: This is still kind of confusing to me. The source material maybe predates auction type sales. It seems like there are three types of sales:
Simple sale: I put up an offer with an asking price, a buyer pays the asking price (from your description, I think you are calling this a quoted sales, the terminology seems a little odd to me).
Conventional auction sale: with a low starting price and a well-defined window of time for bidders to act.
Dutch auction sale: with a high starting price and a step-down interval for the price to drop until someone acts (or the seller withdraws?)

Randynamic (approved):

Randynamic:
❤️

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
ggobugi27: Yes, I think I requires some more context, I'll try to explain a bit better. I used the term "quoted sale" because a non-quoted sale is also available. 

In marmalade contract there are 3 types of sales 
1. sales that don't use quote 
2. sales that use quote without sale contract sale-type "" 
3. sales that use quote with sale contract i.e.) dutch-auction, conventional-auction

Case 1: 
 In this case, the sale's role is only the transfer of token from seller -> escrow -> buyer. 
There "could be" different custom logics that could be involved in token policies that custom it's own sale mechanism, but on the top level marmalade contract, we don't do anything. 

Case 2: 
Requires a "quote" from the seller, where the seller has to specify the fungible price of the sale. The buyer can process the "buy" by paying the quote price. Technically, it would be done by adding an empty sale-type of "" in the quote.

Case 3: 
Requires a "sale-contract" to be specificied with the quote. The sale-contract is granted the ability to "update" the quote price via its logic. 
Technical context:All sale contracts must provide `enforce-withdrawal`, `enforce-quote-update` that needs to pass at both "withdraw" step and "buy" step, giving the sale-contract authority to only allow buy's to be processed for the winning bids and winner accounts, or withdaws after only a certain condition. Conventional auction and dutch auction each store its logic. 

I think here we do describe the case 2 ,(quoted sales without sale contracts), and for an overview document it seems good enough.

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
```diff
@@ -13,7 +13,7 @@ Marmalade is the name of the Kadena **token standard**.
 This token standard defines the interfaces for minting digital items with Kadena **smart contracts** and protecting those digital items with Kadena **token policies**.
 
 Digital items that are unique, one-of-a-kind assets are most commonly referred to as non-fungible tokens (NFTs).
-However, the Kadena token standard—influenced by the Ethereum [multi-token standard](https://ethereum.org/developers/docs/standards/tokens/erc-1155)—enables you to define, mint, and secure tokens that are fully fungible as interchangeable coins, partially fungible as a limited edition with a fixed supply, or completely non-fungible as unique items.
+However, the Kadena token standard is implemented as am multi-token standard that enables you to define, mint, and secure tokens that are fully fungible as interchangeable coins, partially fungible as a limited edition with a fixed supply, or completely non-fungible as unique items.
```
ggobugi27, line 16: I spotted a typo "a multi-token standard"

ggobugi27 (approved):

packages/apps/docs/src/docs/marmalade/what-is-marmalade.md
ggobugi27: I suggest we use the "poly-fungible standard" term instead of "multi-token standard" because we name our token interface as `poly-fungible-v3`

ggobugi27 (approved):

PR #1813, opened 2024-03-08
[@kadena/graph] Integration of Mempool data for tx's (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206309584555266


Ghislain89 (approved):

packages/apps/graph/src/graph/mappers/signer-mapper.ts
```diff
@@ -0,0 +1,23 @@
+import type { Signer } from '@prisma/client';
```
MRVDH, line 1: What do we need these mapper functions for exactly? Doesn't pothos do this already?

packages/apps/graph/src/graph/mappers/transaction-mapper.ts
```diff
@@ -0,0 +1,133 @@
...
+  const prismaSigners = await prismaClient.signer.findMany({
+    where: { requestKey: { in: requestKeys } },
+  });
+
+  return Promise.all(
+    prismaTransactions.map(async (prismaTransaction) => {
+      const transactionSigners = prismaSigners.filter(
+        (s) => s.requestKey === prismaTransaction.requestKey,
+      );
+      return await prismaTransactionMapper(
```
MRVDH, line 22: The Promise.all wont have much effect if you put the await here.

packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -86,34 +88,30 @@ export default builder.prismaNode(Prisma.ModelName.Block, {
     }),
 
     // relations
-    transactions: t.prismaConnection({
-      type: Prisma.ModelName.Transaction,
+    transactions: t.field({
```
MRVDH, line 91: t.connection won't work here I assume right? Due to that totalCount issue?

packages/apps/graph/src/graph/resolvers/transaction-connection.ts
```diff
@@ -0,0 +1,84 @@
...
+  const totalCount = await prismaClient.transaction.count({
+    where: {
+      ...whereCondition,
+    },
+  });
+
+  const connection = await resolveCursorConnection(
+    {
+      args,
+      toCursor: (transaction) => {
```
MRVDH, line 22: We could do this here:
```suggestion
      toCursor(transaction) {
```

packages/apps/graph/src/services/chainweb-node/mempool.ts
```diff
@@ -0,0 +1,59 @@
...
+    super(message);
+    this.mempoolError = mempoolError;
+  }
+}
+
+export async function mempoolGetPending() {
+  try {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+    const res = await fetch(
+      'https://localhost:1789/chainweb/0.0/development/chain/1/mempool/getPending',
```
MRVDH, line 16: The host needs to be an env var

packages/apps/graph/src/services/chainweb-node/mempool.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor(message: string, mempoolError?: any) {
+    super(message);
+    this.mempoolError = mempoolError;
+  }
+}
+
+export async function mempoolGetPending() {
+  try {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
MRVDH, line 14: This feels a bit hacky. Is this only an issue when talking to localhost? Also changing an nev var means it will have an effect on the whole environment as well

packages/apps/graph/src/services/systems-check.ts
```diff
@@ -81,9 +82,15 @@ export async function runSystemsCheck(networkConfig: Promise<NetworkConfig>) {
             },
             {
               title: 'Checking if the mempool is reachable.',
-              skip: () =>
-                'Skipping: There is currently no mempool implementation.',
-              task: async () => {},
+              // skip: () =>
```
MRVDH, line 85: can be removed

packages/apps/graph/src/services/systems-check.ts
```diff
@@ -81,9 +82,15 @@ export async function runSystemsCheck(networkConfig: Promise<NetworkConfig>) {
...
+              //   'Skipping: There is currently no mempool implementation.',
+              task: async () => {
```
MRVDH, line 87: Can be 
```suggestion
              async task() {
```

packages/apps/graph/src/graph/objects/block.ts
nil-amrutlal-dept: Yep exactly

packages/apps/graph/src/graph/mappers/signer-mapper.ts
nil-amrutlal-dept: Pothos doesnt because we map the capabilities to clist (capabilities list). However, I managed to couple signers with prisma again so I completely removed this mapper function

PR #1814, opened 2024-03-08
refactor: get signed/unsigned transactions list

realdreamer:
- Removed the temporary fix of using `all` param
- Improved the `tx list` display instead of relying on filename extension using the transaction signed key

https://app.asana.com/0/1205969628270714/1206751469972661/f

### DIFF
--- a/packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
+++ b/packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
@@ -17,7 +17,6 @@ import type { ICommand, IKeyPair, IUnsignedCommand } from '@kadena/types';
 import type { IWallet } from '../../keys/utils/keysHelpers.js';
 import { getWalletKey } from '../../keys/utils/keysHelpers.js';
 import type { IKeyPair as IKeyPairLocal } from '../../keys/utils/storage.js';
-import { tx } from '../../prompts/index.js';
 import { ICommandSchema } from '../../prompts/tx.js';
 import { services } from '../../services/index.js';
 import type { CommandResult } from '../../utils/command.util.js';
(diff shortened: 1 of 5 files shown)


packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -81,18 +72,18 @@ export async function getTransactions(
...
-          const JSONParsedContent = JSON.parse(content) as
-            | ICommand
-            | IUnsignedCommand;
-          const parsed = schema.safeParse(JSONParsedContent);
-          if (parsed.success && !signed && !all) {
-            const isSigned = JSONParsedContent.sigs.every((sig) => !!sig);
-            if (isSigned) return null;
+          const JSONParsedContent = JSON.parse(content);
+          const isSignedTx =
+            'sigs' in JSONParsedContent &&
```
nillo, line 77: this is incorrect, beause a transaction can also be partially signed, where it has a sigs array with a sig, and null for example if there was one signed and one unsigned and that would still make it unsigned.

realdreamer: yes, you're right! partially signed transaction needs to be considered as unsignedso that when user wants to sign with another set of key will still list them as part of sign command. 

Even when we're running `tx list` that is the current behaviour partial signed transaction considered as unsigned as well. 

in case if i'm missing we can have a quick call and discuss it

barthuijgen: Hmm.. I'm wondering if using the schemas like this still makes sense if the method needs to return all transactions anyway.

I had set up the two schemas as a filter between signed and unsigned, but since you are now accepting all transactions I don't think that's needed anymore.

If you already used `isSignedTransaction` there is no extra value to using the `ICommandSchema` over the `IUnsignedCommandSchema`, this is just double-checking the same thing right?

barthuijgen (approved):

packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -81,18 +71,16 @@ export async function getTransactions(
...
-            | ICommand
-            | IUnsignedCommand;
-          const parsed = schema.safeParse(JSONParsedContent);
-          if (parsed.success && !signed && !all) {
-            const isSigned = JSONParsedContent.sigs.every((sig) => !!sig);
-            if (isSigned) return null;
+          const JSONParsedContent = JSON.parse(content);
+          const parsed = ICommandSchema.safeParse(JSONParsedContent);
+          if (parsed.success) {
+            const isSignedTx = isSignedTransaction(JSONParsedContent);
```
barthuijgen, line 77: Why are you not using the parsed data instead? `parsed.data`

PR #1818, opened 2024-03-08
Fix Select and Combobox aria labeling

salamaashoush:


eileenmguo (approved):

packages/apps/tools/src/components/Global/ChainSelect/index.tsx
```diff
@@ -36,12 +33,10 @@ const ChainSelect: FC<ChainSelectProps> = ({
   return (
     <Select
       {...rest}
-      label="Chain ID"
-      id={ELEMENT_ID}
```
sanderlooijenga, line 40: Out of curiosity; why remove the `id` prop?

sanderlooijenga (approved):

packages/apps/tools/src/components/Global/ChainSelect/index.tsx
salamaashoush: it is redundant

PR #1820, opened 2024-03-11
fix: add animations and small improvements for readability and access…

r-mulder:
This MR contains:
- Animations for opening and closing.
- Fix a layout issue where content had no padding.
- Storybook story had no select for the options.
- Added comment for known issue.

### DIFF
--- a/.changeset/stupid-buckets-carry.md
+++ b/.changeset/stupid-buckets-carry.md
@@ -1,0 +1,5 @@
+---
+'@kadena/react-ui': patch
+---
+
+small fixes for the accordion component
--- a/.changeset/stupid-buckets-carry.md
+++ b/.changeset/stupid-buckets-carry.md
@@ -2,4 +2,4 @@
 '@kadena/react-ui': patch
 ---
 
-small fixes for the accordion component
+Add expand animation to accordion component
--- a/packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
+++ b/packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
@@ -37,18 +37,36 @@ export const accordionButtonClass = style([
 
 export const accordionContentClass = style([
   atoms({
-    display: 'none',
+    display: 'grid',
     color: 'text.subtle.default',
     fontSize: 'base',
     margin: 'no',
     overflow: 'hidden',
+    paddingInlineStart: 'sm',
+    height: '100%',
+    paddingBlock: 'no',
+  }),
+  {
+    gridTemplateRows: '0fr',
+    transition: 'all 300ms ease',
+  },
+]);
+
+export const accordionContentOpenClass = style([
+  atoms({
     paddingBlock: 'sm',
   }),
   {
-    selectors: {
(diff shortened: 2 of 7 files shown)


packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -37,18 +37,36 @@ export const accordionButtonClass = style([
...
-    selectors: {
-      "&[data-open='true']": {
-        display: 'block',
-      },
-    },
+    gridTemplateRows: '1fr',
   },
 ]);
+
+export const accordionContentWrapperClass = style([{ minHeight: '0' }]);
```
salamaashoush, line 64: please ensure that the tools app is working fine, I had the animation before and reverted it because there was an issue with tools and didn't had time to fix back then

r-mulder: I'll test it to be sure, do you remember what the issue was?

packages/libs/react-ui/src/components/Accordion/useAccordionItem.ts
```diff
@@ -29,16 +24,15 @@ export interface AccordionItemAria {
 
 //TODO: Replace with `useAccordionItem` from @react-aria/accordion when it is fixed.
 export function useAccordionItem<T>(
-  props: AccordionItemAriaProps<T>,
+  item: Node<T>,
```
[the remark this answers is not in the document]
r-mulder: I understand, I don't like it from them as they destruct only the item from props in the first line but to make it easier for us I reverted the change.

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
```diff
@@ -37,18 +37,36 @@ export const accordionButtonClass = style([
...
     fontSize: 'base',
     margin: 'no',
     overflow: 'hidden',
+    paddingInlineStart: 'sm',
+    height: '100%',
+    paddingBlock: 'no',
+  }),
+  {
+    gridTemplateRows: '0fr',
+    transition: 'all 300ms ease',
```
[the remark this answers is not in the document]
r-mulder: True, had struggles finding the correct way to only animate padding and grid template rows, but I fixed it in the next commit.

packages/libs/react-ui/src/components/Accordion/Accordion.css.ts
r-mulder: Checked and works like a charm!

salamaashoush (approved):

.changeset/stupid-buckets-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@kadena/react-ui': patch
+---
+
+small fixes for the accordion component
```
eileenmguo, line 5: Changsets will end up making up the release notes so I recommend writing them with more detail/presentation for consumers in mind. For example: 
`Added expand animation to the Accordion component`
...whatever else you think consumers would be important to communicate

PR #1823, opened 2024-03-11
ag/test/git session

alber70g:
- feat(client-utils): added hi to marmalade
- feat(client-utils): added hello to marmalade


packages/libs/client-utils/src/marmalade/index.ts
```diff
@@ -1 +1,9 @@
 export * from './create-token-id';
+
+export function hi() {
+  return 'hi';
```
alber70g, line 4: This should return `Hi!`

```suggestion
  return 'Hi!';
```

PR #1827, opened 2024-03-13
Tokens upgrade

salamaashoush:
This MR will update the tokens from the design system package. 

**Changes made:**
- All references to non existing tokens are fixed with a new reference.
- Added fonts 
- Fix for the crackle bundler which removes client directives from components


eileenmguo (approved):
Not sure how precise we are trying to be in this PR, but here are some inconsistencies I noticed:

`Combobox` arrow doesn't chance colors in dark/light mode like the `Select` component
image
`NumberField` color do switch, but I think the tag colors and the arrow button colors may be flipped? 
image
`TextField` and `TextAreaField` tag colors look like they may be flipped as well 
image
`Tag` close buttons are not visible in dark mode 
image
`Table` background colors are not very visible in dark mode 
image

I'm also not seeing the glass effect. Are we no longer using transparent backgrounds? @isaozler

sstraatemans (approved):

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
```diff
@@ -60,8 +60,8 @@ export const expandedDrawerTitleClass = style([
     display: 'flex',
     justifyContent: 'space-between',
     alignItems: 'center',
-    fontWeight: 'bodyFont.bold',
-    backgroundColor: 'layer-1.default',
+    fontWeight: 'secondaryFont.bold',
+    // backgroundColor: 'layer-1.default',
```
sanderlooijenga, line 64: What will happen with these commented out styles?

sanderlooijenga: It needs backgroundColors (otherwise the sidebar will be transparent)

isaozler:
> Not sure how precise we are trying to be in this PR, but here are some inconsistencies I noticed:
> 
> `Combobox` arrow doesn't chance colors in dark/light mode like the `Select` component image `NumberField` color do switch, but I think the tag colors and the arrow button colors may be flipped? image `TextField` and `TextAreaField` tag colors look like they may be flipped as well image `Tag` close buttons are not visible in dark mode image `Table` background colors are not very visible in dark mode image
> 
> I'm also not seeing the glass effect. Are we no longer using transparent backgrounds? @isaozler

Considering the tag and combo box, I agree. The glass effect is part of a larger refactoring that is currently breaking. I am working to resolve all the issues.

KristinaSpasevska (commented):
Since there are no more layers between different background colors, do we need to separate the Header with a border
Screenshot 2024-03-18 at 13 32 52

KristinaSpasevska (commented):
Icon colors inside FormFields is not clearly visible
Screenshot 2024-03-18 at 13 40 20
Screenshot 2024-03-18 at 13 40 27

KristinaSpasevska (commented):
Tooltips are not visible on dark / light mode 
Screenshot 2024-03-18 at 13 53 45
Screenshot 2024-03-18 at 13 53 52

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
r-mulder: When we implement the new styles we will see if this can be removed.

r-mulder:
@KristinaSpasevska, great comments. I'll will fix the tooltip's background color.
For the icons, we have a story in our sprint to replace the old icon package with the new one. We will start in the tools app and then do the other apps/libs after. @sanderlooijenga This will also fix your comment send via Slack.

packages/apps/tools/src/components/Common/DrawerToolbar/styles.css.ts
r-mulder: When I wrote this only one comment was visible, I'll check.

KristinaSpasevska (approved):

PR #1833, opened 2024-03-14
Test/tools ledger tests

Ghislain89:
This PR adds tests for the Ledger feature for tools


packages/e2e/e2e-tools/package.json
```diff
@@ -25,6 +26,7 @@
     "@kadena-dev/markdown": "workspace:*",
     "@kadena-dev/shared-config": "workspace:*",
     "@kadena/types": "workspace:*",
+    "@ledgerhq/hw-transport-mocker": "6.28.4",
```
sanderlooijenga, line 29: We can now remove this dependency

sanderlooijenga (approved):

packages/e2e/e2e-tools/package.json
Ghislain89: Good catch.

PR #1835, opened 2024-03-15
feat(network): implement network set-default command

realdreamer:
  - https://app.asana.com/0/0/1206807019750943

`kadena network set-default` command

- It sets default network in a specific file `.kadena/defaults/networks/__default__.yaml` 
- When user list the network they could able to see which is the default network

image

- When user deletes the network if it's a default network it will prompted with a warning including it will be deleted from defaults
- Whenever account gets prompted it shows the default network at the top of the list

------------------
When the `--quiet` flag is used:
- If a default network is set and no network argument is provided, the default network should be used without errors.
- If no default network is set and no network argument is provided, an error should be thrown for the missing network argument.

### DIFF
--- a/packages/tools/kadena-cli/src/networks/commands/networkDelete.ts
+++ b/packages/tools/kadena-cli/src/networks/commands/networkDelete.ts
@@ -1,8 +1,12 @@
 import { globalOptions } from '../../utils/globalOptions.js';
-import { removeNetwork } from '../utils/networkHelpers.js';
+import {
+  removeDefaultNetwork,
+  removeNetwork,
+} from '../utils/networkHelpers.js';
 
 import type { Command } from 'commander';
 import { createCommand } from '../../utils/createCommand.js';
+import { getDefaultNetworkName } from '../../utils/helpers.js';
 import { log } from '../../utils/logger.js';
 import { networkOptions } from '../networkOptions.js';
 
@@ -17,8 +21,12 @@ export const deleteNetworksCommand: (
     networkOptions.networkDelete(),
   ],
   async (option) => {
+    const defaultNetworkName = await getDefaultNetworkName();
     const networkData = await option.network();
-    const deleteNetwork = await option.networkDelete();
+    const isDefaultNetwork = networkData.network === defaultNetworkName;
+    const deleteNetwork = await option.networkDelete({
+      isDefaultNetwork,
+    });
 
     log.debug('delete-network:action', {
       ...networkData,
@@ -35,6 +43,7 @@ export const deleteNetworksCommand: (
     }
 
     await removeNetwork(networkData.networkConfig);
+    await removeDefaultNetwork();
 
     log.info(
       log.color.green(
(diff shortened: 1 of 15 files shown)


packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
```diff
@@ -31,6 +36,19 @@ export const createAddAccountManualCommand = createCommand(
 
     const chainId = (await option.chainId()).chainId;
 
+    // when --quiet is passed
+    // network is not provided and
+    // default network is also not found
+    if (!isNotEmptyString(network)) {
```
barthuijgen, line 42: if `--quiet` is passed, it will throw an error only if the flag is not passed AND the option is required. So in this case I assume `network` is not required, and I see you repeating this check in several places.

I'm not sure how complex it would be, but have you looked into making network required, and solving this once at the place of the `quiet` checks?

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -9,6 +9,18 @@ import { services } from '../../services/index.js';
 import { notEmpty } from '../../utils/helpers.js';
 import { isEmpty } from './addHelpers.js';
 
+export const accountResolveAllowedNetworks = [
```
barthuijgen, line 12: I'm not so sure about this, if we hard-code this what is the point of having user-defined networks? A user could add their own network that allows us te resolve accounts and everything.

packages/tools/kadena-cli/src/networks/commands/networkDelete.ts
```diff
@@ -12,19 +24,34 @@ export const deleteNetworksCommand: (
...
+  [globalOptions.networkSelect(), networkOptions.networkDelete()],
   async (option) => {
-    const networkData = await option.network();
-    const deleteNetwork = await option.networkDelete();
+    const defaultNetworkName = await getDefaultNetworkName();
+    const networkData = await option.network({
+      default: defaultNetworkName,
+    });
+
+    networkData.network = networkData.network ?? defaultNetworkName;
```
barthuijgen, line 34: you're already passing `defaultNetworkName` to `option.network()` and here still using it again to set the default, what is it used for in the option then?

realdreamer: missed to remove this! will update it

packages/tools/kadena-cli/src/account/commands/accountAddManual.ts
realdreamer: Yeah, I agree even I'm not happy with current approach with multiple duplication of doing it all place. 

Yes, I tried out introducing default but I was getting same error of missing required arguments initially. Maybe I'll try out something tomorrow morning first thing. Thank you for the feedback

nillo (approved):

barthuijgen (approved):

PR #1840, opened 2024-03-15
[@tools-dapp] Add notification if browser is not supported

KristinaSpasevska:
- Check browser support for Ledger library
- Handle not supported browsers by adding notification with a link to check the compatibility


packages/apps/tools/src/pages/transactions/transfer/index.tsx
```diff
@@ -81,6 +85,22 @@ const TransferPage = () => {
         paddingBlockEnd={'xxxl'}
         gap={'lg'}
       >
+        {!browserSupported ? (
```
sanderlooijenga, line 88: Should we also check whether the Ledger signing method is used to show this notification or not?

sanderlooijenga (approved):

PR #1844, opened 2024-03-18
Update metadata and storage for the current site

lsgunnlsgunn:
Consolidated some information from multiple pages as a preliminary move to the new Docs 2.0 structure.


lsgunnlsgunn:
@ggobugi27, If you get a chance, here's a revised version of the metadata and storage sections. In the Docs 2.0 version, I'll have walk-throughs of create an NFT and create a collection that use IPFS so that section is likely to be much smaller, but for now I kept these mostly the way they were.

lsgunnlsgunn:
Closed in favor of docs2.0 content and structure.

PR #1859, opened 2024-03-20
feat: refactor text and heading components and upgrade storybook

r-mulder:
In this MR I've refactored the Heading and Text component to a new standard. 

Also storybook has been upgraded to version 8.


README.md
```diff
@@ -38,10 +38,10 @@ Overview of the main packages maintained in this repository:
 | [@kadena/cryptography-utils][32]     | [![version][34]][33] |
 | [@kadena/graph][35]                  | [![version][37]][36] |
 | [@kadena/hd-wallet][38]              | [![version][40]][39] |
-| [@kadena/kadena-cli][41]             | [![version][43]][42] |
```
eileenmguo, line 41: was this removal intentional?

packages/libs/react-ui/src/components/Typography/Text/Text.tsx
```diff
@@ -14,70 +15,89 @@ import {
...
+  children: string | ReactNode;
+  className?: string;
   color?: keyof typeof colorVariants;
+  bold?: boolean;
+  size?: 'small' | 'smallest' | 'base';
   transform?: keyof typeof transformVariants;
-  children: React.ReactNode;
+  variant?: 'body' | 'code' | 'ui';
+  ariaLabel?: HTMLAttributes<HTMLSpanElement>['aria-label'];
+  id?: HTMLAttributes<HTMLSpanElement>['id'];
```
eileenmguo, line 38: Since you are extending the span and paragraph elements, I don't think you need to define them here. 

Also, I noticed you are not extending the heading elements in the `Heading` component. I would recommend to keep the approach the same for the `Text` and `Heading` components for predictability

packages/libs/react-ui/src/components/Typography/Text/Text.tsx
```diff
@@ -14,70 +15,89 @@ import {
...
 }
 
-export const Text: FC<ITextProps> = ({
-  as = 'span',
-  variant = 'base',
-  bold = false,
-  color = 'default',
-  transform = 'none',
+const fontMap = {
```
eileenmguo, line 41: nit: I believe you can pull this out of the component itself

packages/libs/react-ui/src/components/Typography/Heading/Heading.tsx
```diff
@@ -11,56 +11,52 @@ import {
...
+ * @param variant - Typography variant defaults to the "as" param
+ */
+
+export const Heading = ({
+  as = 'h2',
   children,
   className,
-  ...props
-}) => {
+  color = 'default',
```
eileenmguo, line 48: Is this intentional? I believe the headings are supposed to be in the "emphasized" color unless that was a design change

eileenmguo (commented):
Looking good! 

Nit: I think the storybook descriptions could be improved. The information provided looks like jsdoc, but I think this information is already represented in the docs. I would suggest providing more succinct descriptions on the purpose of the component/something a bit more readable in the paragraph format.

README.md
r-mulder: No this is all done by the tooling we use. I can try to put it back.

packages/libs/react-ui/src/components/Typography/Text/Text.tsx
r-mulder: Yhea this is kind of an oopsie, I intended to not extend these types to keep the API a bit more strict. Hence why I added the props like this. I'll remove the extend.

packages/libs/react-ui/src/components/Typography/Text/Text.tsx
r-mulder: I'll create a constants file for this one. This will probably happen at other components as well, then I can keep it consistent.

r-mulder:
> Looking good!
> 
> Nit: I think the storybook descriptions could be improved. The information provided looks like jsdoc, but I think this information is already represented in the docs. I would suggest providing more succinct descriptions on the purpose of the component/something a bit more readable in the paragraph format.

Yes I provided jsDocs for more feedback while using the components in your editor. The Storybook is something I want to work on when the first batch of components is ready. I think a consistent way and being able to reference to certain documented guidelines can help with explaining why decisions are made. 
The ticket for more documentation is on the backlog, I'll add some more info on that after some investigation.

packages/libs/react-ui/src/components/Typography/Heading/Heading.tsx
r-mulder: Good catch, I'll check with design.

r-mulder: The color props is removed from the component. The color prop should only be useable on body and UI texts. Which are part of the Text component.

README.md
r-mulder: Yes the package has been set to private like two weeks ago.

eileenmguo (approved):

sstraatemans (approved):

KristinaSpasevska (approved):

PR #1860, opened 2024-03-20
Draft of How-to topics for Docs2

lsgunnlsgunn:
@ggobugi27 @wooglie @jermaine150, I started drafting how-to topics for Create an NFT (with policies) and Create a collection. This is still rough and unfinished but I could use some early guidance on the content.
Especially for creating a collection. I think there's more to it than this lets on:

> in shorts these steps are as followed:
> 
> The Birth of the Collection: Establishing a new digital space to host unique tokens, similar to naming an art exhibition. It's as simple as calling the create-collection function and setting the parameters.
> Crafting the Tokens: Assigning each token a unique ID and defining its rules, like ownership and transfer rights.
> Minting - The Final Touch: Embedding the token onto the kadena blockchain, utilizing marmalade, marking its official entry into the digital world.
> 
😁


lsgunnlsgunn:
Closing this PR because the how-to topics are incorporated into the Docs2.0 structure.

PR #1865, opened 2024-03-21
feat(account): account details and fund command to support multi chain ids

realdreamer:
---
  - https://app.asana.com/0/0/1206807019750929

---

`account details` and account fund both supports now multiple chainids `1,2` or `1-5` or `all` 

`account details` with partial success
image

`account details` not available on any chains
image

`account fund` with success

image

`account fund` with error

image

`account fund` with partial success

image


packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
```diff
@@ -145,3 +174,48 @@ export const fundAmountValidation = z
...
+  input: string,
+  separator?: ',' | '-',
+): number[] | undefined => {
+  if (isEmpty(separator)) {
+    return [parseInt(input.trim(), 10)];
+  } else if (separator === ',') {
+    const splitValue = new Set(input.split(','));
+    return [...splitValue]
+      .map((id) => parseInt(id.trim(), 10))
+      .filter(notEmpty);
```
barthuijgen, line 196: parseInt failing will result in `NaN`, I don't know if `notEmpty` filters that correctly

packages/tools/kadena-cli/src/account/utils/fundHelpers.ts
```diff
@@ -0,0 +1,125 @@
...
+      .map((obj) => Object.keys(obj))
+      .flat()
+      .join(', ');
+    log.info(
+      log.color.green(
+        `\nAccount "${accountName}" funded with ${amount} ${fungible}(s) on Chain ID(s) "${chainIds}" in ${networkId} network.`,
+      ),
+    );
+    log.info(
+      log.color.green(`Use "account details" command to check the balance.`),
```
barthuijgen, line 115: I think it should show the full command, if you're going to print it `kadena account details`. It could even include the `--account` flag already

packages/tools/kadena-cli/src/constants/config.ts
```diff
@@ -48,3 +48,5 @@ export const PLAIN_KEY_LEGACY_EXT = '.legacy.key';
 export const IS_DEVELOPMENT = process.env.NODE_ENV === 'development';
 export const IS_TEST = process.env.VITEST === 'true';
 export const MAX_CHARACTERS_LENGTH = 80;
+
+export const MAX_CHAIN_IDS: number = 20;
```
barthuijgen, line 52: you define max chains constant here, but I see `19` hardcoded on a lot of other places in this PR

packages/tools/kadena-cli/src/account/utils/accountHelpers.ts
realdreamer: It doesn't filter out the `NaN` it just returns the values and i'm validating against ZodSchema here https://github.com/kadena-community/kadena.js/blob/feat/support-multi-chain-ids/packages/tools/kadena-cli/src/prompts/account.ts#L275-L281

realdreamer: image

nillo (commented):
other then my other comment lgtm,

packages/tools/kadena-cli/src/account/commands/accountDetails.ts
```diff
@@ -1,55 +1,113 @@
...
+interface IAccountDetailsConfig
+  extends Omit<IGetAccountDetailsParams, 'chainId'> {
+  chainId: ChainId[];
+}
+
+interface IAccountDetails {
+  [key: string]: IAccountDetailsResult;
+}
+
+const formatWarnings = (warnings: string[]): string => {
```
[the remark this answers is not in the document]
realdreamer: `warnings.length === 0` `return null` this i just fixed it.

nope! it's not only the `chainIds` i'm concatenating at the end `return `${prefix} ${sortedChainIds.join(',')}`; 

The error message looks like this 
`Account "account name" is not available on\nfollowing chain(s) of the "testnet04" network: 6,7,8`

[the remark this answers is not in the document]
nillo: 🙏🫡

packages/tools/kadena-cli/src/account/utils/fundHelpers.ts
realdreamer: I think keeping it generic `kadena account details` make simpler. if we're adding --account flag then it makes sense to add other flags as well.

PR #1867, opened 2024-03-21
Test/expect to pass

Ghislain89:
The latest version of the devnet sandbox has instant mining enabled. During our tests we've noticed that there is a slight delay of around 100ms before chainweb data is able to catch up with the burst of blocks that are being mined. 

To ensure our graph tests are repeatable, the expects are now wrapped in a wrapper expect that retries the assertion to ensure chainweb data has enough time to catch up. Small example can be found below

```TS
await expect(async () => {
  const response = await page.request.get('https://api.example.com');
  expect(response.status()).toBe(200);
}).toPass();
```

While testing these changes, a bug was discovered in Chainweb data that certain pieces of information are only indexed _after a certain block height_. Since mining is now on demand, a fewer amount of blocks is being mined and our tests actually manage to execute transfers before this block height. This caused our tests to fail.

In our effort to debug this, two composite actions have been created that archive the sandbox logs as well as the sandbox volumes. 

These can be downloaded as artifacts and to debug any issues with the sandbox. I've decided to leave these in, as it also opens up the door for future improvements such as: "Start a sandbox container in the cloud and use the state that was used in CI"


packages/e2e/e2e-graph/tests/account.test.ts
nil-amrutlal-dept, line 1: The graph test file changes are going to conflict with the changes I made here: https://github.com/kadena-community/kadena.js/pull/1813

Ghislain89: Yeah, i'm waiting on your PR to be merged.

.gitignore
```diff
@@ -102,3 +102,5 @@ temp
 
 # parcel
 .parcel-cache
+
+.github/actions/devnet/l1/state/*
```
Ghislain89, line 106: can be removed

PR #1873, opened 2024-03-22
[@kadena/graph] Transaction Refactor with Mempool Data (feat)

nil-amrutlal-dept:
- Inclusion of mempool data in transaction - created union on the transaction result type
- Refactor transaction relations - now found in `TransactionInfo`
- Adjustments in `graph-client`
- Adjustments in `e2e-graph`

---
  - https://app.asana.com/0/0/1206918504101417
  - https://app.asana.com/0/0/1206309584555266


Ghislain89 (approved):

packages/apps/graph/src/graph/mappers/transaction-mapper.ts
```diff
@@ -0,0 +1,57 @@
+import type { Transaction } from '@prisma/client';
+
+export function mempooTransactionMapper1(mempoolData: any): Transaction {
```
MRVDH, line 3:
```suggestion
export function mempoolTransactionMapper(mempoolData: any): Transaction {
```

packages/apps/graph-client/src/graphql/fields/signer.graph.ts
```diff
@@ -3,7 +3,10 @@ import type { DocumentNode } from 'graphql';
 
 export const CORE_SIGNER_KEY_FIELDS: DocumentNode = gql`
   fragment CoreSignerKeyFields on Signer {
-    capabilities
+    clist {
```
MRVDH, line 6: Don't we want to call this `capabilitiesList`?

packages/apps/graph/src/services/chainweb-node/mempool.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  constructor(message: string, mempoolError?: any) {
+    super(message);
+    this.mempoolError = mempoolError;
+  }
+}
+
+export async function mempoolGetPending() {
+  try {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
MRVDH, line 16: I mentioned this in the other PR as well but is there no other way around this? This feels like a bad idea

packages/apps/graph/src/services/chainweb-node/mempool.ts
```diff
@@ -0,0 +1,100 @@
...
+  hash: string,
+  chainId?: string,
+): Promise<string | undefined> {
+  const mempoolData = await mempoolLookup(hash, chainId);
+
+  if (mempoolData && mempoolData[0].tag === 'Pending') {
+    return 'Pending';
+  }
+
+  return undefined;
```
MRVDH, line 73: we can remove this line I think

packages/apps/graph-client/src/graphql/fields/signer.graph.ts
nil-amrutlal-dept: I think here we want to respect the existing model which has this field called clist (please confirm @alber70g). You can check the model here: https://api.chainweb.com/openapi/pact.html#tag/model-payload

packages/apps/graph/src/services/chainweb-node/mempool.ts
nil-amrutlal-dept: I think I would need to have a certificate if we don't want to disable TSL/SSL certificate verification. @enobayram do you maybe know if this would be possible to obtain? Without it I can't make requests to the mempool

enobayram: As far as I know, the node's service port doesn't serve the mempool endpoint, but I'm not sure why, (might be worth asking the chainweb team to expose it there).

As long as we need to get that information from the p2p port, and if the node isn't configured with a domain name and a valid certificate for that domain, you need to accept an unauthorized certificate. That said, if I'm not mistaken, this code snippet globally disables the TLS authority check in a rather ad-hoc manner, on which I agree with @MRVDH that it seems very shady.

My 2 cents here would be to make a check at `graph` start up whether the configured p2p endpoint is reachable. If the `graph` operator wants to make `graph` talk securely with their `chainweb-node`, then they can set up a proper domain and a signed SSL certificate for it, then `graph` can just talk to it without messing with TLS settings. Or maybe the `graph` operator is running the node and `graph` in a secure network, in which case they don't care about a signed SSL certificate. In this case, your reachability check would fail at `graph` start up and at that time you can inform the user about the situation and tell them that they should either configure their node with a domain or explicitly run `graph` with `NODE_TLS_REJECT_UNAUTHORIZED=0` in the environment.

packages/apps/graph/src/services/chainweb-node/mempool.ts
```diff
@@ -0,0 +1,140 @@
...
+    const options = {
+      hostname: dotenv.MEMPOOL_HOSTNAME,
+      port: dotenv.MEMPOOL_PORT,
+      path: '/chainweb/0.0/development/chain/1/mempool/getPending',
+      method: 'POST',
+      rejectUnauthorized: false, // This disables certificate verification
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    };
```
nil-amrutlal-dept, line 26: This fixes it for now. We disable certificate verification for this specific request. Will create a ticket to discuss the possibility of implementing an optional certificate integration

packages/apps/graph/src/graph/types/graphql-types.ts
```diff
@@ -123,9 +123,16 @@ export interface TransactionMeta {
...
 // eslint-disable-next-line @typescript-eslint/naming-convention
+export interface MempoolInfo {
+  status: string;
+}
 
-export interface TransactionResult {
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export interface TransactionInfo {
+  hash: string;
+  chainId: bigint;
```
nil-amrutlal-dept, line 133: I added these two here because I need this data to be able to get `Block`, `Events`, and `Transfers` in the `TransactionInfo` object. However, these are not exposed in the `TransactionInfo` output

packages/apps/graph/src/graph/objects/transaction-result.ts
```diff
@@ -1,6 +1,23 @@
...
+  description: 'The mempool information.',
+  fields: (t) => ({
+    status: t.exposeString('status', {
+      description: 'The status of the mempool.',
+      nullable: true,
+    }),
+  }),
+});
+
+const TransactionInfo = builder.objectType('TransactionInfo', {
```
nil-amrutlal-dept, line 20: Is `eventCount` needed here? Since we included the `events` object? 
The same question regarding `blockHash`. 
Here, the problem with removing these is that if we query for these fields via the corresponding objects we will be querying the database an additional time. @alber70g

MRVDH (approved):

packages/apps/graph/src/graph/objects/transaction-result.ts
alber70g: We can remove them from the schema but query them from Prisma right?

alber70g (approved):
Minor comments

packages/apps/graph/generated-schema.graphql
```diff
@@ -534,32 +541,21 @@ type TransactionCommand {
...
 
-"""The payload of a transaction."""
-union TransactionPayload = ContinuationPayload | ExecutionPayload
-
 """The result of a transaction."""
-type TransactionResult {
+type TransactionInfo {
   """The transaction result when it was successful. Formatted as raw JSON."""
   badResult: String
+  block: Block!
```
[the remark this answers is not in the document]
nil-amrutlal-dept: At the moment `TransactionResult = MempoolInfo | TransactionInfo`. Any suggestions for the overall field name? I could swap them, so it'd become `TransactionInfo = MempoolTransactionInfo | TransactionResult`

packages/apps/graph/src/graph/mappers/transaction-mapper.ts
```diff
@@ -0,0 +1,57 @@
...
+    gasLimit: BigInt(mempoolTx.cmd.meta.gasLimit),
+    gasPrice: mempoolTx.cmd.meta.gasPrice,
+    nonce: mempoolTx.cmd.nonce,
+    pactId: mempoolTx.cmd.payload.pactId,
+    rollback: mempoolTx.cmd.payload.rollback,
+    proof: mempoolTx.cmd.payload.proof,
+    senderAccount: mempoolTx.cmd.meta.sender,
+    step: BigInt(mempoolTx.cmd.payload.step),
+    ttl: BigInt(mempoolTx.cmd.meta.ttl),
+  };
```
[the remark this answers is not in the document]
nil-amrutlal-dept: There are some fields that do not overlap on name, such as `requestKey` which is `hash` in our case or `sender` which is `senderAccount` in our case. We could do indeed the spread but we'd have to map the non-identical fields.

PR #1877, opened 2024-03-25
feat(network): network unset with none

realdreamer:
---
  - https://app.asana.com/0/0/1206877879294040

### DIFF
--- a/packages/tools/kadena-cli/src/prompts/network.ts
+++ b/packages/tools/kadena-cli/src/prompts/network.ts
@@ -256,6 +256,7 @@ export const networkSelectWithNonePrompt: IPrompt<string> = async (
   args,
   isOptional,
 ): Promise<string> => {
+  const defaultNetwork = previousQuestions.defaultNetwork;
   const existingNetworks = await getEnsureExistingNetworks();
 
   const choices: ICustomNetworkChoice[] = existingNetworks.map((network) => ({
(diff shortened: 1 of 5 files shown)


packages/tools/kadena-cli/src/prompts/network.ts
```diff
@@ -247,6 +251,31 @@ export const networkSelectOnlyPrompt: IPrompt<string> = async (
...
+  const existingNetworks = await getEnsureExistingNetworks();
+
+  const choices: ICustomNetworkChoice[] = existingNetworks.map((network) => ({
+    value: network.value,
+    name: network.name,
+  }));
+
+  choices.unshift({
+    value: 'none',
+    name: 'None',
```
barthuijgen, line 268: The `name` could be more descriptive like "Remove default network", and does the option only show up if a default network was already set?

packages/tools/kadena-cli/src/prompts/network.ts
```diff
@@ -247,6 +251,34 @@ export const networkSelectOnlyPrompt: IPrompt<string> = async (
...
+  const existingNetworks = await getEnsureExistingNetworks();
+
+  const choices: ICustomNetworkChoice[] = existingNetworks.map((network) => ({
+    value: network.value,
+    name: network.name,
+  }));
+
+  if (isNotEmptyString(defaultNetwork)) {
+    choices.unshift({
+      value: 'none',
```
nillo, line 269: if you do this: `const action = network === 'none' ? 'unset' : 'set';`

why not just make the value 'unset' instead of none, since its not visible to the end user anyways?
i would make a choice between internal value usage choice,

Pick none or pick unset:D

nillo (approved):

packages/tools/kadena-cli/src/prompts/network.ts
realdreamer: it makes sense if it's only with prompts but when user wants to pass it directly with args. `--network=unset` looks bit odd.

barthuijgen (approved):

PR #1879, opened 2024-03-25
[@kadena/graph] Fungible queries - make fungible name not required (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206918504101440

### DIFF
--- a/packages/apps/graph/src/utils/default.ts
+++ b/packages/apps/graph/src/utils/default.ts
@@ -1,0 +1,1 @@
+export const defaultFungibleName: string = 'coin';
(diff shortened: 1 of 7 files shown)


packages/apps/graph/src/utils/default.ts
```diff
@@ -0,0 +1 @@
+export const defaultFungibleName: string = 'coin';
```
MRVDH, line 1: I'd say let's make this an env var. For example when people want to deploy their own graph specific for their own fungible. Otherwise they have to edit and build the code from scratch

MRVDH (changes requested):

packages/apps/graph/src/utils/default.ts
nil-amrutlal-dept: Makes sense! Will make the change

MRVDH (approved):

PR #1880, opened 2024-03-25
fix(pou): Fix landscape camera mode

Likita:
- Fix landscape camera mode
- Adding Dashboard button in case of error after signing

### DIFF
--- a/packages/apps/proof-of-us/src/components/Status/ErrorStatus.tsx
+++ b/packages/apps/proof-of-us/src/components/Status/ErrorStatus.tsx
@@ -1,14 +1,16 @@
 import { MonoClose } from '@kadena/react-icons';
 import { Stack } from '@kadena/react-ui';
+import Link from 'next/link';
 import type { FC, PropsWithChildren } from 'react';
 import { Button } from '../Button/Button';
 import { MessageBlock } from '../MessageBlock/MessageBlock';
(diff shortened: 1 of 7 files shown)


packages/apps/proof-of-us/src/components/Status/ErrorStatus.tsx
```diff
@@ -1,14 +1,16 @@
...
 import { Stack } from '@kadena/react-ui';
+import Link from 'next/link';
 import type { FC, PropsWithChildren } from 'react';
 import { Button } from '../Button/Button';
 import { MessageBlock } from '../MessageBlock/MessageBlock';
 import { Heading } from '../Typography/Heading';
 
 interface IProps extends PropsWithChildren {
-  handleClose?: () => void;
+  handleClose?: string;
```
sstraatemans, line 10: if we make it a url, maybe better for readability to rename.
`handleClose` kind of implies a function.
maybe `closeUrl` or something

sstraatemans (changes requested):
some small remarks

can we make it a bit bigger maybe? feels small? I understand that the title maybe goes over the view.
if that is the case. maybe we should hide the title on landscape mode. camerview is more important.

Ghislain89:
Tested it, seems to work fine!

sstraatemans (approved):
if ghislain is happy. Im happy

PR #1882, opened 2024-03-26
feat(kadena-cli) added support for displaying missing holes, and upda…

nillo:
[x] Added support for displaying holes within a template
[x] Updated docs for templating

### DIFF
--- a/packages/tools/kadena-cli/README.md
+++ b/packages/tools/kadena-cli/README.md
@@ -801,8 +801,9 @@ signing and submission.
 ### Available Templates
 
 Currently, two default templates are provided: `transfer` and `safe-transfer`.
-These templates cover the most common transaction types, allowing for
-straightforward transfers of tokens between accounts.
+(Default templates are stored at `.kadena/transaction-templates`) These
+templates cover the most common transaction types, allowing for straightforward
+transfers of tokens between accounts.
 
 ### Command Usage
 
@@ -826,7 +827,7 @@ options available:
 ### Example Command
 
 ```
-kadena tx add --template="transfer.yaml" --template-data="data.yaml" (--account-from="k:account" --account-to="k:toaccount" --amount="1" --chain-id="1" --pk-from="publicKey") --network-id="testnet04" --out-file="transaction.json"
+kadena tx add --template="transfer.yaml" --template-data="data.yaml" --network-id="testnet04" --out-file="transaction.json"
 ```
 
 In this example, `transfer.yaml` is the template used to construct the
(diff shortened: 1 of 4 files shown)


packages/tools/kadena-cli/README.md
```diff
@@ -716,27 +736,121 @@ Tool for creating and managing transactions
...
+| `--template-data`       | File path for the data used to fill the template.         | No       |
+| `--network-id`          | Specifies the network ID (e.g., `testnet04`).             | Yes      |
+| `--out-file`            | Path for saving the generated transaction file.           | No       |
+| `--holes`               | Displays a list of required template variables.           | No       |
+| Custom options          | Generated based on the chosen template's required fields. | Varies   |
+
+### Example Command
+
+```
+kadena tx add --template="transfer.yaml" --template-data="data.yaml" (--account-from="k:account" --account-to="k:toaccount" --amount="1" --chain-id="1" --pk-from="publicKey") --network-id="testnet04" --out-file="transaction.json"
```
realdreamer, line 773:
```suggestion
kadena tx add --template="transfer.yaml" --template-data="data.yaml"  --network-id="testnet04" --out-file="transaction.json"
```

I think the values within the brackets could be removed from here and we can explain it below. because someone copy and wanna try out it would be easier

packages/tools/kadena-cli/README.md
```diff
@@ -716,27 +736,121 @@ Tool for creating and managing transactions
...
+```
+
+In this example, `transfer.yaml` is the template used to construct the
+transaction. `data.yaml` contains the user-supplied values for the template
+variables. The `--network-id` specifies which network the transaction is
+intended for, and `--out-file` determines where the generated transaction file
+will be saved.
+
```
realdreamer, line 781: before or after this, we can show an example of how transfer.yaml file and data.yaml looks like

packages/tools/kadena-cli/README.md
```diff
@@ -716,27 +736,121 @@ Tool for creating and managing transactions
...
+        args: []
+  - public: { { key:to } }
+    caps:
+      - name: 'coin.TRANSFER'
+        args: [{ { { account:to } } }, { { { account:from } } }, 0.000000000001]
+networkId: '{{network:networkId}}'
+type: exec
+```
+
+This template exemplifies how prefixes guide users in providing the correct
```
realdreamer, line 837: (just question) do we also want to mention about where default templates are stored ie. `.kadena/transaction-templates`

realdreamer (approved):

packages/tools/kadena-cli/README.md
```diff
@@ -716,27 +736,121 @@ Tool for creating and managing transactions
...
+  gasLimit: 2000
+  gasPrice: 0.00000001
+  ttl: 7200
+signers:
+  - public: { { key:from } }
+    caps:
+      - name: 'coin.TRANSFER'
+        args:
+          [
+            { { { account:from } } },
```
barthuijgen, line 823: pretty sure these spaces will kill the template

packages/tools/kadena-cli/src/tx/commands/txCreateTransaction.ts
```diff
@@ -109,6 +115,13 @@ export const createTransactionCommandNew = createCommand(
       );
     }
     const template = await option.template({ stdin });
+    const showHoles = await option.holes();
+
+    if (showHoles.holes === true) {
+      const holes = await getVariablesByTemplate(template.template, { stdin });
```
barthuijgen, line 121: the holes already exist in `template.templateConfig.variables` right?

packages/tools/kadena-cli/README.md
nillo: yeah i like that, no magic idd lets be specific, ill add it

packages/tools/kadena-cli/src/tx/commands/txCreateTransaction.ts
nillo: lol missed that, ill update

packages/tools/kadena-cli/README.md
```diff
@@ -722,28 +742,125 @@ Tool for creating and managing transactions
...
+  chainId: '{{chain-id}}'
+  sender: '{{{account:from}}}'
+  gasLimit: 2300
+  gasPrice: 0.000001
+  ttl: 600
+signers:
+  - public: '{{key:from}}'
+    caps:
+      - name: 'coin.TRANSFER'
+        args: ['{{{account:from}}}', '{{{account:to}}}', { { decimal:amount } }]
```
barthuijgen, line 807: still spaces

packages/tools/kadena-cli/README.md
```diff
@@ -209,26 +209,29 @@ kadena network add --network-name="mainnet" --network-id="mainnet01" --network-h
 ```
 
 ---
+
 ```
 kadena network set-default [arguments]
```
barthuijgen, line 214: set-default doesn't accept [arguments] only [options]

PR #1894, opened 2024-03-27
Feat(react UI)/new button component

r-mulder:
In this PR the button has been updated to the new styles.

**Highover changes:**
- Badge component, to support all features of the button.
- New button API to support the new variants.
- Render an anchor tag when you provide a href value.
- Reduce API clutter.
- Add documentation for in editor feedback.


eileenmguo (commented):
So in general I feel like the `Button` component has grown too complex. The approach we have taken so far is to try to use composition over configuration in most cases (there is some information about this in the CONTRIBUTING.md). The reason for this is to reduce complexity since components are decoupled

In this PR, the approach is to provided props for every pattern that @isaozler has designed to configure the end result. I would expect the `Button` component to only be concerned with `Button` specific styles and to defer any other styles/configuration to the "decorator" components themselves. Also, I believe that the different designs Isa provided are more to give examples of the patterns/compositions he would like to be applied at the moment however I can imagine this might change in the future. The current API would make maintainability of the `Button` component more difficult since additional "decorators" or changes to the existing patterns would be required in the `Button` component itself.

I think that providing a start and end icon/decorator prop that allows the user to compose their `Button` would be simpler and easier to maintain in the long run. Curious to see how you and others feel about this though

packages/apps/dev-wallet/src/pages/backup-recovery-phrase/write-down/write-down-recovery-phrase.tsx
```diff
@@ -73,7 +73,7 @@ export function WriteDownRecoveryPhrase() {
         <Text size="small">{mnemonic}</Text>
         <Button
           type="submit"
-          onClick={() => {
+          onPress={() => {
```
Randynamic, line 76: In my opinion, it's preferable to adhere more closely to the native browser APIs. Although these APIs predate touch events, we currently don't support a `press` event, which makes this counterintuitive. Moreover, using the component now requires learning its specific API, which feels redundant, especially for something as basic as a button. What was the reason moving towards `onPress`?

packages/apps/docs/src/components/BrowseSection/BrowseSection.tsx
```diff
@@ -64,7 +64,7 @@ const BrowseSection: BrowseSectionType = ({
             !React.isValidElement(child) ||
             (child.type !== LinkBlock &&
               child.type !== NextLink &&
-              child.type !== Link &&
+              child.type !== Button &&
```
Randynamic, line 67: just for my understanding because a link and a button have 2 different functionalities is this button triggering an action or just a link?

packages/apps/docs/src/docs/build/index.tsx
```diff
@@ -43,13 +43,9 @@ const Home: FC<IProps> = ({ blogPosts, popularPages }) => {
                 clicking Quick start.
               </Text>
             </Box>
-            <Link
-              endIcon={<SystemIcon.TrailingIcon />}
-              variant="contained"
-              href="/build/quickstart"
-            >
+            <Button icon={<SystemIcon.TrailingIcon />} href="/build/quickstart">
```
Randynamic, line 46: This for me really feels counter-intuitive as the button is now a Link and I can't read from the tag name what the purpose of the element is. what was the reason for changing this?

packages/libs/react-ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,35 @@
...
+  className,
+  size = 'lg',
+  style = 'default',
+}: IBadgeProps) => {
+  return (
+    <span className={classNames(badge({ size, style }), className)}>
+      {children}
+    </span>
+  );
+};
```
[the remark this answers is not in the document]
r-mulder: Yes, I should add a label and the value. Will do that.

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,264 @@
...
-    ISharedButtonProps {}
+type BaseProps = Omit<AriaFocusRingProps, 'isTextInput'> & Variants;
+
+export interface ICustomProps extends BaseProps {
+  avatarProps?: Omit<IAvatarProps, 'size'>;
+  badgeValue?: string | number;
+  children?: ReactNode | ReactNode[];
+  className?: string;
+  icon?: ReactElement;
+  iconPosition?: 'start' | 'end';
```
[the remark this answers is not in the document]
r-mulder: I totally agree but Isa wanted to force a way that only one icon can be placed inside a button. 

This kind of ties into the entire design of the button and component library, do we want full flexibility or limit options and force a look and feel with our components. Maybe we should have that discussion before we move on.

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,264 @@
...
-    props.icon
+  const startIcon = iconPosition === 'start' && icon;
+  const endIcon = iconPosition === 'end' && icon;
+  const iconOnly =
+    (icon && !children) || (loadingLabel === '' && props.isLoading);
+  const isLoading = props.isLoading && loadingLabel !== '';
+
+  const isLoadingAriaLiveLabel = `${
+    typeof children === 'string' ? children : buttonProps['aria-label'] ?? 'is'
+  } loading`.trim();
```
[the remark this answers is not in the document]
r-mulder: This is exactly the feedback I'm looking for, I missed the use case for aria label on an icon only button. I'll add the possibility in the API.

packages/apps/dev-wallet/src/pages/backup-recovery-phrase/write-down/write-down-recovery-phrase.tsx
r-mulder: This was documented in the shared type for the button & link:
``` typescript /**
   * @deprecated use `onPress` instead to be consistent with React Aria, also keep in mind that `onPress` is not a native event it is a synthetic event created by React Aria
   * @see https://react-spectrum.adobe.com/react-aria/useButton.html#props
   */
  onClick?: ComponentProps<'button'>['onClick'];

I agree on using onClick feels more intuitive for a web developer, this API design is coming from React Aria which we kind of extend.

File: packages/apps/docs/src/components/BrowseSection/BrowseSection.tsx (hunk 2 above)
Comment (r-mulder) [reply packages/apps/docs/src/components/BrowseSection/BrowseSection.tsx:67]: This is the only use case I was having doubts on this implementation. The button behaves as a link if you pass an href, if an onPress is provided it's an action on the page. 

I am missing context about this component to be sure if this is correct behaviour. I'm looking for feedback on this.

File: packages/apps/docs/src/docs/build/index.tsx (hunk 3 above)
Comment (r-mulder) [reply packages/apps/docs/src/docs/build/index.tsx:46]: Personal experience in previous libraries and talks with other developers. Some prefer a button which can handle both and some don't. 

I'm trying to get feedback on this but I'm already thinking about splitting them again due to recent insights.

File: packages/libs/react-ui/src/components/Button/Button.tsx [hunk 7]
```diff
@@ -1,87 +1,264 @@
...
+type Variants = NonNullable<RecipeVariants<typeof button>>;
 
-export interface IButtonProps
-  extends PickedAriaButtonProps,
-    ISharedButtonProps {}
+type BaseProps = Omit<AriaFocusRingProps, 'isTextInput'> & Variants;
+
+export interface ICustomProps extends BaseProps {
+  avatarProps?: Omit<IAvatarProps, 'size'>;
+  badgeValue?: string | number;
```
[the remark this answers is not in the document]
r-mulder: I think that either way we are bound to that. (my thoughts) If you would provide it as a child component and there is a breaking change you still need to change it. One way to mitigate it could be to split styling and functionality. Set up a hook for the styling and copy the component to prevent API changes from breaking but still allow consistency with styling. 
This way you have less components to replace on a breaking change in one go but don't have feature parity with the actual component and will probably have the breaking change on a later stage in the button if needed. 

Let's have a discussion about this during refinement, I would love to explore other approaches.

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,264 @@
...
-      aria-disabled={props.isLoading || undefined}
-      data-disabled={props.isDisabled || undefined}
-      data-pressed={isPressed || undefined}
-      data-hovered={isHovered || undefined}
-      data-focused={isFocused || undefined}
-      data-focus-visible={isFocusVisible || undefined}
+      {...{
+        ...sharedAttributes,
+        ...mergeProps(buttonProps, hoverProps, focusProps),
+        ref: ref as ForwardedRef<HTMLButtonElement>,
```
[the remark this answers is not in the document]
Randynamic: I agree with this. For me having the explicit use case of a link and a button makes more sense

packages/apps/dev-wallet/src/pages/backup-recovery-phrase/write-down/write-down-recovery-phrase.tsx
Randynamic: Ahh I missed this one. So this is something from React-Aria. I'm curious why Adobe chose this?

r-mulder: I think they described it in their blog post about the button [link](https://react-spectrum.adobe.com/blog/building-a-button-part-1.html). In short they wanted to create a unified way to handle all interaction on multiple platforms.

packages/libs/react-ui/src/components/Button/Button.tsx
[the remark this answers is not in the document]
r-mulder: I've split up the two components and reverted the changes which changed the existing Link components into Buttons.

packages/libs/react-ui/src/components/Button/Button.tsx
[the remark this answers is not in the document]
r-mulder: I've added an override for the icon only use case in the last commit.

packages/libs/react-ui/src/components/Badge/Badge.tsx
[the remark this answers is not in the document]
r-mulder: Added an optional role prop in case the content gets updated you can provide the 'status' role to it. Also added the aria label for a more descriptive description if needed.

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,150 @@
...
-  'href' | 'target' | 'rel' | 'elementType'
->;
+type Variants = NonNullable<RecipeVariants<typeof button>>;
 
-export interface IButtonProps
-  extends PickedAriaButtonProps,
-    ISharedButtonProps {}
+export type IButtonProps = Omit<AriaFocusRingProps, 'isTextInput'> &
+  Variants &
+  Omit<AriaFocusRingProps, 'isTextInput'> &
```
eileenmguo, line 25: duplicate

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,150 @@
...
+        ref={ref}
+      >
+        <span
+          className={classNames({
+            [centerContentWrapper]: !iconOnly,
+            [iconOnlyStyle]: iconOnly,
+            [noEndVisualStyle]: !endVisual && !iconOnly,
+            [noStartVisualStyle]: !startVisual && !iconOnly,
+            [startVisualStyle]: startVisual && !iconOnly,
+            [endVisualStyle]: endVisual && !iconOnly,
```
eileenmguo, line 121: This feels like a group of styles where some are mutually exclusive. Could this be simplified?

packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -1,87 +1,150 @@
...
+            [directionStyle]: props.isLoading && startVisual && !endVisual,
```
eileenmguo, line 122: What is this doing?

packages/libs/react-ui/src/components/Button/ToggleButton.tsx
```diff
@@ -29,7 +29,7 @@ const ToggleButtonBase = (
 
   const onlyIcon = props.icon !== undefined;
   const content = onlyIcon ? (
-    props.icon
+    <span className={iconOnlyStyle}>{props.icon}</span>
```
eileenmguo, line 32: can this be done within the Button component?

eileenmguo (approved):
looks great!

packages/libs/react-ui/src/components/Button/ToggleButton.tsx
r-mulder: Yes I've left it out of scope for now and just fix the styling which was reported by Steven. Wanted to know what the use cases would be for this button.

packages/libs/react-ui/src/components/Button/Button.tsx
r-mulder: I changed the name to "reversedDirectionStyle" does that make more sense?

packages/apps/docs/src/components/Layout/components/Header/Header.tsx
```diff
@@ -64,15 +64,17 @@ export const Header: FC<IProps> = ({ menuItems, layout = 'full' }) => {
             className={socialsClass}
             href="https://twitter.com/kadena_io"
             title="Go to our X"
-            icon={<SystemIcon.Twitter />}
-          />
+          >
+            <SystemIcon.Twitter />
```
sstraatemans, line 68: 🔥

sstraatemans (approved):

sanderlooijenga (approved):
Approved on behalf of Tools part

ferreroltd (approved):

nil-amrutlal-dept (approved):

PR #1898, opened 2024-03-28
[@kadena/graph] Add minimum depth argument to Events query and subscription (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206918504101418


packages/apps/graph/src/services/confirmation-depth-service.ts
```diff
@@ -0,0 +1,48 @@
...
+  const latestBlocks = await prismaClient.block.groupBy({
+    by: ['chainId'],
+    _max: {
+      height: true,
+    },
+    where: {
+      chainId: {
+        in: (chainIds as string[]).map((id) => parseInt(id)),
+      },
+    },
```
MRVDH, line 37: Only do this `where` if chainIds are passed.

```suggestion
    ...(chainIds?.length && { where: {
      chainId: {
        in: chainIds.map((id) => parseInt(id)),
      },
    }}),
```

packages/apps/graph/src/graph/subscription/events.ts
```diff
@@ -121,6 +134,12 @@ async function getLastEvents(
           column: 'parameters',
         }),
       }),
+      ...(minimumDepth && {
+        OR: await getConditionForMinimumDepth(
+          minimumDepth,
+          chainId ? [chainId] : chainIds,
```
MRVDH, line 140: I wouldn't pass the default chainIds here since it makes the query more complex. See other comment.

```suggestion
          chainId ? [chainId] : undefined,
```

packages/apps/graph/src/services/confirmation-depth-service.ts
nil-amrutlal-dept: Good catch

MRVDH (approved):

PR #1903, opened 2024-03-28
[@kadena/graph] Add minimum depth in Transaction queries (Feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206918504101420

### DIFF
--- a/packages/apps/graph/generated-schema.graphql
+++ b/packages/apps/graph/generated-schema.graphql
@@ -340,9 +340,9 @@ type Query {
 
   """
   Retrieve events by qualifiedName (e.g. `coin.TRANSFER`). Default page size is 20.
-  
+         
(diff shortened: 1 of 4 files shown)


packages/apps/graph/generated-schema.graphql
```diff
@@ -340,9 +340,9 @@ type Query {
 
   """
   Retrieve events by qualifiedName (e.g. `coin.TRANSFER`). Default page size is 20.
-
```
MRVDH, line 343: These should not be removed. They are required to have new lines in the GraphiQL docs, since regular newlines make the rest of the text dissapear.

packages/apps/graph/generated-schema.graphql
```diff
@@ -499,9 +499,9 @@ type Signer implements Node {
 type Subscription {
   """
   Listen for events by qualifiedName (e.g. `coin.TRANSFER`).
-
```
MRVDH, line 502: Same here

MRVDH (approved):

PR #1904, opened 2024-03-28
fix(tx-status): add poll option 

realdreamer:
`tx status` by default it uses the `getStatus` method to fetch the data. If the transaction not found returns an error message with an hint for the user to try it out with `poll` flag. 

If user uses the `--poll` option then it uses the `pollStatus` method to fetch the data to keep polling for a minute. If the request not found in a minute then request gets timedout. 

---
  - https://app.asana.com/0/0/1206932423881496
  - https://app.asana.com/0/0/1206930586643761

### DIFF
--- a/packages/tools/kadena-cli/src/tx/commands/txStatus.ts
+++ b/packages/tools/kadena-cli/src/tx/commands/txStatus.ts
@@ -49,7 +49,7 @@ export const getTxStatus = async ({
         status: 'error',
         errors: [
           notFoundErrorMessage,
-          `If the transaction is just submitted, please try with poll flag "kadena tx status --request-key=${requestKey} --chain-id=${chainId} --network=${networkConfig.network} --poll"`,
+          `If the transaction is just submitted, please try with poll flag: kadena tx status --request-key=${requestKey} --chain-id=${chainId} --network=${networkConfig.network} --poll`,
         ],
       };
     }
--- a/packages/tools/kadena-cli/src/tx/commands/txStatus.ts
+++ b/packages/tools/kadena-cli/src/tx/commands/txStatus.ts
@@ -10,33 +10,59 @@ import { globalOptions } from '../../utils/globalOptions.js';
 import { log } from '../../utils/logger.js';
 import { txOptions } from '../txOptions.js';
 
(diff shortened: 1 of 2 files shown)


nillo (approved):

nillo (approved):

packages/tools/kadena-cli/src/tx/commands/txStatus.ts
```diff
@@ -10,33 +10,59 @@ import { globalOptions } from '../../utils/globalOptions.js';
...
+      ? requestKey.slice(0, -1)
+      : requestKey;
+
+    if (result[trimmedRequestKey] === undefined) {
+      return {
+        status: 'error',
+        errors: [
+          notFoundErrorMessage,
+          `If the transaction is just submitted, please try with poll flag "kadena tx status --request-key=${requestKey} --chain-id=${chainId} --network=${networkConfig.network} --poll"`,
+        ],
```
barthuijgen, line 53:
```suggestion
          `If the transaction is just submitted, please try with poll flag: kadena tx status --request-key=${requestKey} --chain-id=${chainId} --network=${networkConfig.network} --poll`,
        ],
```

packages/tools/kadena-cli/src/tx/commands/txStatus.ts
```diff
@@ -10,33 +10,59 @@ import { globalOptions } from '../../utils/globalOptions.js';
...
+    };
+    if (txPoll) {
+      result = await pollStatus(payload, {
+        timeout: 60000,
+      });
+    } else {
+      result = await getStatus(payload);
+    }
+
+    const trimmedRequestKey = requestKey.endsWith('=')
```
barthuijgen, line 43: why is this trimming here?

barthuijgen: I'm not familiar with this behaviour, it seems odd. Is it documented anywhere? would be very helpful to have a commented link or something for future maintenance to make it clear why this is here.

barthuijgen (approved):

nillo (approved):

PR #1909, opened 2024-03-29
[@kadena/graph] Implement confirmation depth calculations (Feat)

nil-amrutlal-dept:
Added confirmation depth to queries and subscriptions where minimum depth is an argument;

---
  - https://app.asana.com/0/0/1206941884773067


MRVDH (approved):

packages/apps/graph/src/services/complexity.ts
nil-amrutlal-dept, line 1: @MRVDH let me know if this makes sense

PR #1936, opened 2024-04-04
Added Guard to non fungible chain accounts

MRVDH:
image


packages/apps/graph/src/graph/objects/non-fungible-chain-account.ts
```diff
@@ -27,21 +30,43 @@ export default builder.node(
...
+        complexity: COMPLEXITY.FIELD.CHAINWEB_NODE,
+        async resolve(parent) {
+          try {
+            const tokenDetails = await tokenDetailsLoader.load({
+              accountName: parent.accountName,
+              chainId: parent.chainId,
+            });
+
+            const accountDetails = await nonFungibleAccountDetailsLoader.load({
+              tokenId: tokenDetails[0].id,
```
MRVDH, line 55: Can we make the assumption that for this account and chain the guard is the same for all tokens? @alber70g

alber70g: No we cannot :)
Basically each token has it's own tables. So also it's own accountnames, guards and balances. So they can have a different guard as well

sstraatemans (approved):

packages/apps/graph/src/services/chainweb-node/non-fungible-account-details.ts
```diff
@@ -0,0 +1,80 @@
...
+  balance: number;
+  guard: {
+    keys: string[];
+    pred: Guard['predicate'];
+  };
+};
+
+function getClient(chainId: string, networkId: string): IClient {
+  return createClient(
+    `${dotenv.NETWORK_HOST}/chainweb/0.0/${networkId}/chain/${chainId}/pact`,
```
nil-amrutlal-dept, line 22: There's an `apiVersion` property on network that can be used to replace the hardcoded `0.0`.

packages/apps/graph/src/services/chainweb-node/non-fungible-account-details.ts
```diff
@@ -0,0 +1,80 @@
...
+  );
+}
+
+export async function getNonFungibleAccountDetails(
+  tokenId: string,
+  accountName: string,
+  chainId: string,
+): Promise<NonFungibleChainAccountDetails | null> {
+  let result;
+  const networkId = (await networkConfig).networkId;
```
nil-amrutlal-dept, line 32: Regarding the `apiVersion`

```suggestion
  const { networkId, apiVersion } = await networkConfig;
```

nil-amrutlal-dept (approved):

packages/apps/graph/src/services/chainweb-node/non-fungible-account-details.ts
MRVDH: Feedback applied 👍

packages/apps/graph/src/services/chainweb-node/non-fungible-account-details.ts
MRVDH: Feedback applied 👍

packages/apps/graph/src/graph/objects/non-fungible-chain-account.ts
MRVDH: Feedback applied 👍

PR #1940, opened 2024-04-05
feat(pou): Add a confirmation popup for Logout

Likita:
- Confirmation popup for Logout
- Bold header for Modals
- Native confirm() popups changed to Custom Confirmation Modal
- Confirm for Drop out btn

### DIFF
--- a/packages/apps/proof-of-us/src/components/Confirmation/Confirmation.tsx
+++ b/packages/apps/proof-of-us/src/components/Confirmation/Confirmation.tsx
@@ -1,4 +1,5 @@
 import { Button } from '@/components/Button/Button';
+import { showModalBtnClass } from '@/components/Confirmation/styles.css';
 import { Modal } from '@/components/Modal/Modal';
 import { Stack } from '@kadena/react-ui';
 import type { FC, PropsWithChildren } from 'react';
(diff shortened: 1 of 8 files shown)


packages/apps/proof-of-us/src/components/Confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { FC, PropsWithChildren } from 'react';
+import { useState } from 'react';
+
+interface IProps extends PropsWithChildren {
+  text: string;
+  action: () => void;
+}
+
+export const Confirmation: FC<IProps> = ({ text, children, action }) => {
+  const [isShowed, setIsShowed] = useState(false);
```
sstraatemans, line 13: isShown, setIsShown

packages/apps/proof-of-us/src/components/Modal/style.css.ts
```diff
@@ -49,4 +49,5 @@ export const dialogClass = style([
 export const headerClass = style({
   fontSize: '18px',
   width: '100%',
+  fontWeight: '700',
```
sstraatemans, line 52: i think this is a value in atoms, no?

packages/apps/proof-of-us/src/pages/user/index.tsx
```diff
@@ -36,9 +37,14 @@ const Page: FC = () => {
             <TitleHeader
               label="Dashboard"
               Append={() => (
-                <IconButton onClick={logout}>
-                  <MonoLogout />
-                </IconButton>
+                <Confirmation
```
sstraatemans, line 40: I like it!

packages/apps/proof-of-us/src/components/Modal/style.css.ts
Likita: But since atoms are wrappers, do we need to use them if we already have a custom class for the element?

sstraatemans (approved):

PR #1949, opened 2024-04-09
feat(kadena-cli) auto select network or fallback to prompt

nillo:
[x] Automatically select network (id) when sending or fallback to prompt
[x] Added logging so its more clear what transactions is being send and might fail

### DIFF
--- a/packages/tools/kadena-cli/src/prompts/tx.ts
+++ b/packages/tools/kadena-cli/src/prompts/tx.ts
@@ -1,4 +1,8 @@
-import type { ICommand, IUnsignedCommand } from '@kadena/types';
+import type {
+  ICommand,
+  ICommandPayload,
+  IUnsignedCommand,
+} from '@kadena/types';
 import { z } from 'zod';
 import {
   getTransactions,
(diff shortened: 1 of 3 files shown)


barthuijgen (approved):

packages/tools/kadena-cli/src/prompts/tx.ts
```diff
@@ -508,6 +512,19 @@ export async function selectSignMethodPrompt(): Promise<'wallet' | 'keyPair'> {
...
 
+function determineNetwork(networkId: string | null): string {
+  const id = networkId ?? '';
+
+  if (id.includes('testnet')) {
+    return 'testnet';
+  } else if (id.includes('mainnet')) {
+    return 'mainnet';
+  } else if (id.includes('devnet')) {
+    return 'devnet';
```
realdreamer, line 523: network ids are different than the alias/network names, so relying on the ids like testnet04, mainnet01 and development would be better. isn't it?

realdreamer (approved):
other than my above comment. LGTM 👍🏽

PR #1955, opened 2024-04-09
Correct non-id fields types

nil-amrutlal-dept:
There were some fields exposed as IDs which resulted in these having the same type as the node IDs which can confuse. Retyped these non-id fields in the graph to their actual primitive type

---
  - https://app.asana.com/0/0/1207034879037532

### DIFF
--- a/packages/apps/graph/generated-schema.graphql
+++ b/packages/apps/graph/generated-schema.graphql
@@ -15,7 +15,7 @@ type Block implements Node {
 
   """Default page size is 20."""
   events(after: String, before: String, first: Int, last: Int): BlockEventsConnection!
-  hash: ID!
+  hash: String!
   height: BigInt!
   id: ID!
   minerAccount: FungibleChainAccount!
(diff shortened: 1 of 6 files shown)


packages/apps/graph/generated-schema.graphql
```diff
@@ -290,7 +290,7 @@ type NonFungibleTokenBalance {
   balance: Int!
   chainId: String!
   guard: Guard!
-  id: ID!
+  id: String!
```
MRVDH, line 293: This should probably become `tokenId`. Since the id property in GraphQL is what we use as the Relay node id.

nil-amrutlal-dept: Yep, agree. Will change it

MRVDH: I'm not sure if there are other object definitions where a DB id and a Relay ID possibly clash. Might have to check that

nil-amrutlal-dept: These will imply changes in `graph-client` and PoU

nil-amrutlal-dept: Should we merge this as is, and put those changes on a different PR to be dealt with when we have a proper way of dealing with breaking changes? @MRVDH @alber70g 
Or should I do it in this PR and have it on hold

MRVDH: Let's put it in a different PR

MRVDH (approved):

PR #1956, opened 2024-04-09
feat(cli): create a wallet and account in config init command 

realdreamer:
I moved very generic helpers from helpers file to globalHelpers to avoid circular dependency.

---
  - https://app.asana.com/0/0/1206807019750911


packages/tools/kadena-cli/src/services/config/config.service.ts
```diff
@@ -37,8 +45,53 @@ export interface IConfigService {
 }
 
 export class ConfigService implements IConfigService {
-  // eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/parameter-properties
-  public constructor(private services: Services) {}
+  public directory: string | null = null;
```
barthuijgen, line 48: since there already is `getDirectory` this one should be private.

packages/tools/kadena-cli/src/services/config/config.service.ts
```diff
@@ -37,8 +45,53 @@ export interface IConfigService {
...
+    // eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/parameter-properties
+    private services: Services,
+    directory?: string,
+  ) {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.setDirectory(directory);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  private async setDirectory(directory?: string): Promise<void> {
```
barthuijgen, line 59: in the config.ts file the logic was specifically written to be sync, and while it being in a service gives somewhat more flexibility, this is still dangerous since class constructors can not be async.

doing for exmaple:
```ts
process.env.ENV_KADENA_DIR = "/whatever"; // (valid directory)
const services = new Services();
console.log(services.config.getDirectory());
```
will probably be `null`, since the async `setDirectory()` did not finish yet.

There's two options to solve it:
1. what I usually do when I want async in a constructor is introduce a `init()` method, not called within the constructor, so you have to do 
```ts
const services = new Services();
await services.init()
```
that's tricky in this case, since services is by default a singleton you can import and use directly, a top-level await could be used in ESM, but I'm not so experienced in it's downsides.

2. make `setDirectory` not async, it was already that way, so it's not hard to do. The downside is that it's less clean, we have the `directoryExists` service method available, seems odd not to use it.

---
I personally prefer option 2.

packages/tools/kadena-cli/src/tx/utils/txHelpers.ts
```diff
@@ -679,3 +680,10 @@ export async function logTransactionDetails(command: ICommand): Promise<void> {
     log.info(`No transaction details to display`);
   }
 }
+
+export const getTxTemplateDirectory = (): string | null => {
+  const kadenaDir = services.config.getDirectory();
+  return isNotEmptyString(kadenaDir)
```
barthuijgen, line 686: why did you decide to use `isNotEmptyString` everywhere on `config.getDirectory()`'s response? It's type is `string | null` so just checking non-null should be sufficient. 

I understand this works fine, but the wording of the method, and checking for something (empty string) that it can never be will cause confusion for code readability.

packages/tools/kadena-cli/src/wallets/walletOptions.ts
```diff
@@ -6,6 +6,16 @@ import { createOption } from '../utils/createOption.js';
 import { mnemonicPromptTransform } from '../utils/helpers.js';
 
 export const walletOptions = {
+  createWalletConfirmation: createOption({
+    key: 'createWallet' as const,
+    prompt: wallets.createWalletPrompt,
+    defaultIsOptional: false,
+    validation: z.string(),
+    option: new Option(
+      '-c --create-wallet <createWallet>',
```
barthuijgen, line 15: if `--create-account` is `-a` should `--create-wallet` be `-w`?

barthuijgen (approved):

nillo (approved):

PR #1965, opened 2024-04-10
Fixed incorrect IDs in schema

MRVDH:


nil-amrutlal-dept:
image

I see the same result when I query it via `node (id)`.

nil-amrutlal-dept (approved):

PR #1978, opened 2024-04-12
feat(fund): enable account funding on devnet

realdreamer:
---
  - https://app.asana.com/0/0/1206951762878097

### DIFF
--- a/packages/tools/kadena-cli/src/account/commands/accountFund.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountFund.ts
@@ -1,18 +1,14 @@
-// import { describeModule } from '@kadena/client-utils/built-in';
-// import { Option } from 'commander';
+import { describeModule } from '@kadena/client-utils/built-in';
 import ora from 'ora';
-// import { z } from 'zod';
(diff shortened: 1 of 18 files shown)


packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -78,66 +64,56 @@ export const createAccountFundCommand = createCommand(
...
+        const { deployFaucet } = await option.deployFaucet();
+
+        if (!deployFaucet) {
+          return;
+        }
+        const loader = ora(
+          `Deploying faucet on chain Id(s): "${undeployedChainIdsStr}" in "${network}" network...\n`,
+        ).start();
+
+        await deployDevNetFaucet(undeployedChainIds).catch((e) => {
```
barthuijgen, line 106: I would map over `undeployedChainIds` in here so you can handle success/error individually per chain, and if 1 deployment fails but others succeed you can give the correct output information.

nillo (approved):

barthuijgen (approved):

PR #1984, opened 2024-04-17
Network Info Endpoint (feat)

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1206590017183465


nil-amrutlal-dept:
Some issues found:
- endpoint `/coins` is not reachable in `testnet` and `mainnet`. Its is only on the local devnet
- local devnet's explorer page gets data about total transaction and circulating coins from `/stats`. The same is observed here: this endoint is not reachable on `testnet` or `mainnet`

Do we need to expose any additional ports for this? @alber70g

MRVDH (approved):

PR #1986, opened 2024-04-18
Add policies to non fungible tokens

nil-amrutlal-dept:


packages/apps/graph/src/services/token-service.ts
```diff
@@ -110,3 +117,105 @@ export async function checkAccountChains(
...
+    );
+  }
+
+  let executionCmd;
+
+  if (version === 'v1') {
+    // Note: Alternative approach left for reference
+    // executionCmd = execution(
+    //   Pact.modules['marmalade.ledger']['get-policy-info'](tokenId),
+    // );
```
nil-amrutlal-dept, line 138: Left on purpose. Same for the following

PR #1987, opened 2024-04-18
Feat(react UI)/new input components

r-mulder:
In this PR I've updated the textField component to the new branding.

### DIFF
--- a/.changeset/dry-coins-juggle.md
+++ b/.changeset/dry-coins-juggle.md
@@ -1,0 +1,6 @@
+---
+'@kadena/proof-of-us': patch
+'@kadena/react-ui': patch
+---
+
+Update the textfield and numberfield components to the new branding
--- a/packages/apps/docs/src/components/Layout/components/SideMenu/SideMenu.tsx
+++ b/packages/apps/docs/src/components/Layout/components/SideMenu/SideMenu.tsx
@@ -1,6 +1,7 @@
 import { EVENT_NAMES, analyticsEvent } from '@/utils/analytics';
 import type { IMenuItem } from '@kadena/docs-tools';
 import { Box, Heading, SystemIcon, TextField } from '@kadena/react-ui';
+import { atoms } from '@kadena/react-ui/styles';
 import classNames from 'classnames';
 import { useRouter } from 'next/router';
 import type { FC, KeyboardEvent } from 'react';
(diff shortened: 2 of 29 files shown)


ferreroltd (approved):

sanderlooijenga (approved):

sstraatemans (approved):

.changeset/dry-coins-juggle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@kadena/proof-of-us': patch
+'@kadena/react-ui': patch
+---
+
+Update the textfield and numberfield components to the new branding
```
eileenmguo, line 6: I would recommend adding more detail when writing changesets. Essentially add any information you would want someone using the library to know before they upgrade. So for example any breaking changes, and new features, etc. Otherwise someone might upgrade and not realize why something is suddenly not working

packages/apps/docs/src/components/Layout/components/SideMenu/SideMenu.tsx
```diff
@@ -71,10 +72,11 @@ export const SideMenu: FC<IProps> = ({ closeMenu, menuItems }) => {
             id="search"
             onKeyUp={handleKeyPress}
             placeholder="Search"
-            isOutlined
             type="text"
             aria-label="Search"
-            endAddon={<MagnifierIcon />}
+            endAddon={
+              <MagnifierIcon className={atoms({ paddingInline: 'n2' })} />
```
eileenmguo, line 78: Is this padding specifically for icons? if not I would recommend adding spacing to the input itself

packages/libs/react-ui/src/components/Form/ActionButtons/NumberFieldActions.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { iconSize } from '../NumberField/NumberField.css';
+
+interface INumberFieldActionsProps {
+  isDisabled?: boolean;
+  variant: string;
+  state: any;
+  size?: INumberFieldProps['size'];
+}
+
+export const NumberFieldActions = ({
```
eileenmguo, line 15: nit: I personally wouldn't categorize this as an action button since it should only be used with the number field

PR #2006, opened 2024-04-24
feat(docs): add mermaid styling of codeblock

sstraatemans:
https://app.asana.com/0/1205619636589593/1206862413948496/f


packages/apps/docs/src/components/Markdown/Code/Mermaid.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const ref = useRef<HTMLPreElement>(null);
+
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.innerHTML = ref.current.innerText;
+      mermaid.initialize({
+        theme: 'base',
+        startOnLoad: true,
+        themeVariables: {
+          primaryColor: '#2767a3',
```
sstraatemans, line 17: hardcoded values. we cant use the tokens, because they are `css var` values

PR #2015, opened 2024-04-25
fix: add account alias prompt for init + wallet add command

realdreamer:
account alias option is an optional field because it's required only when create account is passed as true

So, incase when `--quiet` flag is used with `--create-account=true` but `account-alias` is **not** passed. it still create the wallet but doesn't create a account alias fails with error 

---
  - https://app.asana.com/0/0/1207158682603351

### DIFF
--- a/packages/tools/kadena-cli/src/config/commands/configInit.ts
+++ b/packages/tools/kadena-cli/src/config/commands/configInit.ts
@@ -95,8 +95,12 @@ export const createConfigInitCommand: (
     logWalletInfo(created.words, wallet.filepath, key.publicKey);
 
     if (createAccount === 'true') {
+      // when --quiet is passed and account alias is not provided
+      // we will not create an account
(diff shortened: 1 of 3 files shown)


packages/tools/kadena-cli/src/config/commands/configInit.ts
```diff
@@ -87,6 +95,15 @@ export const createConfigInitCommand: (
     logWalletInfo(created.words, wallet.filepath, key.publicKey);
 
     if (createAccount === 'true') {
+      // when --quiet is passed and account alias is not provided
+      // we will not create an account
+      if (!notEmpty(accountAlias)) {
+        log.error(
```
barthuijgen, line 101: Not sure about the error state, since the wallet did get created successfully. My initial feeling is that this should be a warning. (another reason commands should just do 1 thing... but that's a lost battle 😅)

nillo (approved):

PR #2020, opened 2024-04-25
fix(account): when display public keys from wallet prefix index and alias

realdreamer:
Screenshot reference for `index alias public key` 

image

---
  - https://app.asana.com/0/0/1207170077788327

### DIFF
--- a/packages/tools/kadena-cli/src/prompts/account.ts
+++ b/packages/tools/kadena-cli/src/prompts/account.ts
@@ -311,18 +311,16 @@ export const chainIdPrompt: IPrompt<string> = async (
   })) as ChainId;
 };
 
-export const selectPublicKeysFromWalletPrompt: IPrompt<string> = async (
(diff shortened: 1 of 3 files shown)


packages/tools/kadena-cli/src/prompts/account.ts
```diff
@@ -306,3 +308,57 @@ export const chainIdPrompt: IPrompt<string> = async (
     },
   })) as ChainId;
 };
+
+function walletKeysToPromptChoices(wallets: IWallet[]): {
```
barthuijgen, line 312: there is a `tableFormatPrompt` method that I've used for aligning values in a prompt, can you check if that can be re-used?

realdreamer: it's worked 👍🏽

barthuijgen (approved):

PR #2025, opened 2024-04-26
chore(kadena-cli) make mnemonic phrase more noticable

nillo:
Make logging of mnemonic more noticable so the end user knows its importance.

### DIFF
--- a/packages/tools/kadena-cli/src/wallets/commands/walletImport.ts
+++ b/packages/tools/kadena-cli/src/wallets/commands/walletImport.ts
@@ -44,16 +44,20 @@ export const createImportWalletCommand: (
       });
       wallet = await services.wallet.storeKey(wallet, key);
 
-      loading.succeed('Wallet imported successfully');
+      loading.succeed('Wallet imported successfully\n');
 
-      log.info(log.color.green('Mnemonic Phrase'));
+      log.info('====================================================');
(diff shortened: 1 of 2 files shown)


packages/tools/kadena-cli/src/wallets/commands/walletImport.ts
```diff
@@ -44,16 +44,20 @@ export const createImportWalletCommand: (
       });
```
barthuijgen, line 44: did we want to add an extra prompt? 
> first show mnemonic phrase in output
> prompt `did you save the mnemonic phrase in a safe and secure place (Y/N): `
> continue with command as usual

nillo: Sure, we might entertain the idea of doing just that, however, given the astonishingly high level of the Extreme Indicator, one would naturally assume, nay, confidently expect, that they would possess the intellectual wherewithal — indeed, an almost uncanny level of acumen — to not only grasp it, but they would intuitively, almost instinctively know that they should, without any shadow of a doubt, be smart enough to securely store it :P

What do you think?

barthuijgen: 🤷‍♂️

barthuijgen (approved):

PR #2028, opened 2024-04-26
fix(graph): update `marmalade:deploy` to deploy the updated files in marmalade repo

ggobugi27:
Updates the script to deploy the updated file naming syntax and version upgrades from https://github.com/kadena-io/marmalade/pull/197

### DIFF
--- a/packages/apps/graph/src/devnet/deployment/marmalade/deploy.ts
+++ b/packages/apps/graph/src/devnet/deployment/marmalade/deploy.ts
@@ -11,7 +11,7 @@ import { logger } from '@utils/logger';
 import { validateObjectProperties } from '@utils/validate-object';
 import type { IAccount } from '../helper';
 import { inspect, listen, signAndAssertTransaction, submit } from '../helper';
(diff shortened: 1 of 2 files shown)


packages/apps/graph/src/devnet/deployment/marmalade/deploy.ts
```diff
@@ -78,9 +81,14 @@ export async function deployMarmaladeContracts(
     file.endsWith(marmaladeRemoteConfig.templateExtension),
   );
 
+  const upgradeTemplateFiles = templateFiles.filter((f) => {
+    return f.split('.')[1] === '8' || f.split('.')[3] === 'guard-policy-v1';
+  });
+
```
nil-amrutlal-dept, line 87: Just to be sure, the files that match this criteria are to be deployed first and then upgraded?

ggobugi27: Yes, they're to be deployed with `upgrade=false` then deployed with `upgrade=true`

nil-amrutlal-dept (approved):

PR #2029, opened 2024-04-26
feat(docs): copy code blocks

sstraatemans:
https://app.asana.com/0/1205619636589593/1207146957039882/f


Ghislain89:
@sstraatemans copying works. Could we add some visual indication that the text has been copied? similar to like we did for PoU where i said ✅ Copied! if you copied the QR link

PR #2030, opened 2024-04-26
fix(docs): sidemenu visibility

sstraatemans:
bugfix. it seems that the sidemenu is not visiible between width md and lg

https://app.asana.com/0/1205619636589593/1207181240645363/f


Ghislain89:
image
What's with the grey bit not overlapping with the title bar?

Ghislain89 (approved):

PR #2035, opened 2024-04-29
Tools dApp - Module Explorer UX/UI revamp - Module list

sanderlooijenga:
This PR contains the revamped lefthand side menu of the new UX/UI of the Modules Explorer.

---
  - https://app.asana.com/0/0/1207445755216490


packages/libs/client-utils/src/built-in/describe-module.ts
```diff
@@ -1,31 +1,42 @@
 import { execution } from '@kadena/client/fp';
 import { pipe } from 'ramda';
+import { z } from 'zod';
 import { dirtyReadClient } from '../core/client-helpers';
 import type { IClientConfig } from '../core/utils/helpers';
 
 /**
  * @alpha
  */
-export interface IDescribeModuleOutput {
```
javadkh2, line 9: I think its better to keep the interface, then the type is more clear for end user.
What do you think?

sanderlooijenga: Yeah, ofcourse, no issue at all. Will refactor it to be an `interface` again 👍

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
```diff
@@ -0,0 +1,224 @@
...
+interface INodeProps<T>
+  extends Omit<ICustomAccordionProps<T>, 'data' | 'children'> {
+  data: TreeItem<T>[];
+  level: number;
+  onItemClick: (item: TreeItem<T>) => void;
+  onExpandCollapse: (item: TreeItem<T>, expanded: boolean) => void;
+}
+
+// eslint-disable-next-line react/function-component-definition
+function Node<T>({
```
r-mulder, line 131: I would split this off in separate files.

r-mulder: Same goes for the accordion, it's easier to read if they are small concentrated files.

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
```diff
@@ -0,0 +1,224 @@
...
+}: INodeProps<T>) {
+  return (
+    <CustomAccordion
+      {...rest}
+      data={data}
+      itemProps={{
+        style: {
+          borderBlockEnd:
+            level === 1 ? '1px solid rgba(0, 0, 0, 0.25)' : 'initial',
+        },
```
r-mulder, line 146: I would suggest to not use inline styles, also you are not using tokens for this color. So moving this to a css file is a win win for that.

r-mulder: You could also opt in for a optional prop on the item. It's a custom component build for the use cases in the tools app..

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
```diff
@@ -0,0 +1,224 @@
...
+            >
+              {hasChildren ? (
+                <Button
+                  variant="transparent"
+                  onClick={() => {
+                    child.toggleExpandCollapse();
+
+                    // We "toggle" the expanded state ourselves since this is still the previous state
+                    onExpandCollapse(child.data, !child.isExpanded);
+                  }}
```
r-mulder, line 181: Same as previous component, why a separate handler on the icon itself? If you don't want the icon to be clickable I would put pointerevents: none on the element and let the wrapper handle it.

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
```diff
@@ -0,0 +1,224 @@
...
+              onClick={() => {
+                if (child.data.children.length) {
+                  child.toggleExpandCollapse();
+
+                  // We "toggle" the expanded state ourselves since this is still the previous state
+                  onExpandCollapse(child.data, !child.isExpanded);
+                } else {
+                  onItemClick(child.data);
+                }
+              }}
```
r-mulder, line 165: As we encountered performance issues I would suggest to move this handler outside of the render. 

A function like handleClick has performance benefits as it's declared only once and stays in memory even during rerenders. If you have them inline they don't stay in memory.

packages/apps/tools/src/components/Global/ModuleExplorer/SidePanel/index.tsx
```diff
@@ -1,76 +1,76 @@
...
-  const { selectedNetwork } = useWalletConnectClient();
-
-  const onChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
-    setText(e.target.value);
-    startTransition(() => {
-      setSearchQuery(e.target.value);
-    });
-  };
-
+// eslint-disable-next-line react/function-component-definition
```
r-mulder, line 22: Is it possible to just disable this rule in the projects eslint?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
```diff
@@ -1,89 +1,88 @@
...
-  modules: IChainModule[];
   openedModules: IEditorProps['openedModules'];
-  onModuleClick: ISidePanelProps['onResultClick'];
-  onInterfaceClick: ISidePanelProps['onInterfaceClick'];
-  onInterfacesExpand: ISidePanelProps['onInterfacesExpand'];
-  onModuleExpand: ISidePanelProps['onModuleExpand'];
+  onModuleClick: ISidePanelProps<IncompleteModuleModel>['onModuleClick'];
   onActiveModuleChange: IEditorProps['onActiveModuleChange'];
-  onTabClose: IEditorProps['onTabClose'];
+  // onTabClose: IEditorProps['onTabClose'];
```
r-mulder, line 20: Do you still need this one?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
```diff
@@ -1,89 +1,88 @@
...
+            data: 'explorer',
+          },
+          {
+            title: 'Outline',
+            key: 'outline',
+            label: 'Outline',
+            children: outlineItems,
+            data: 'outline',
+          },
+        ]}
```
r-mulder, line 62: I see some duplication.. Cant we just use the label prop and fill in the others with a capitalize function or lowercase function for the other parts?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
```diff
@@ -101,7 +100,7 @@ const ModuleExplorer = ({
               );
             }),
           );
-          onTabClose(module);
+          // onTabClose(module);
```
r-mulder, line 103: Is this intended?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
```diff
@@ -1,89 +1,88 @@
...
+            label: 'Outline',
+            children: outlineItems,
+            data: 'outline',
+          },
+        ]}
+        onReload={onReload}
+        onModuleClick={({ data }) => {
+          if (isModuleLike(data)) {
+            const chainModule: IChainModule = moduleModelToChainModule(data);
+            setActiveModule(chainModule);
```
r-mulder, line 67: Do we really need to do these steps before checking if it's already opened?

packages/apps/tools/src/hooks/use-modules-query.ts
```diff
@@ -0,0 +1,57 @@
...
+) => {
+  const chains = chainIds ?? CHAINS;
+
+  const promises = chains.map((chainId) =>
+    listModules({
+      defaults: { networkId, meta: { chainId } },
+    }),
+  );
+
+  const results = await Promise.all(promises);
```
r-mulder, line 39: Do we need some kind of error handling?

packages/apps/tools/src/pages/modules/explorer/utils.ts
```diff
@@ -48,3 +50,93 @@ export function getCookieValue<Expected>(
...
+      const moduleList = modulesMap.get(
+        namespace,
+      )! as Array<IncompleteModuleModel>;
+
+      moduleList.push(module);
+    }
+  });
+
+  return modulesMap;
+};
```
r-mulder, line 142: Do you have tests for this file? It quite hard to read what this is about if you don't know the context. Tests might help and a few comments would be even better.

packages/apps/tools/src/components/Global/CustomAccordion/CustomAccordion.tsx
```diff
@@ -0,0 +1,79 @@
...
+          {...itemProps}
+        >
+          {children}
+        </Item>
+      ))}
+    </ul>
+  );
+}
+
+export default CustomAccordion;
```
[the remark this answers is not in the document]
sanderlooijenga: Yeah true, I started down this path and afterwards discussed it with you. Since I already had this and it worked, I didn't bother simplifying it. Still quite happy with how it works though

packages/apps/tools/src/components/Global/CustomAccordion/CustomAccordion.tsx
```diff
@@ -0,0 +1,79 @@
...
+  fillHeight,
+  style,
+  ...rest
+}: IItemProps<T>) {
+  const [isExpanded, setIsExpanded] = React.useState(_isExpanded);
+
+  return (
+    <li
+      {...rest}
+      style={{ ...style, flex: fillHeight && isExpanded ? 1 : undefined }}
```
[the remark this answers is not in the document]
sanderlooijenga: I will refactor it to use `classNames` 👍 

Regarding accessibility; I render a button (from `react-ui`) now which handles the toggling of the accordion, doesn't that button then have accessibility traits?

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
```diff
@@ -0,0 +1,224 @@
...
+          <>
+            <Stack
+              alignItems={'center'}
+              justifyContent={'space-between'}
+              onClick={() => {
+                item.toggleExpandCollapse();
+
+                // We "toggle" the expanded state ourselves since this is still the previous state
+                onExpandCollapse(item.data, !item.isExpanded);
+              }}
```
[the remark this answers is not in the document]
sanderlooijenga: I thought it might be nice to not only have the toggling handled by a "small" button but also be easy to click the entire item. I can revert this of course

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
sanderlooijenga: Will do 🫡

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
sanderlooijenga: I will refactor them to `useCallback`s 👍

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
sanderlooijenga: It will be enabled again in an upcoming PR

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
sanderlooijenga: Will be enabled again in an upcoming PR

packages/apps/tools/src/hooks/use-modules-query.ts
sanderlooijenga: It can be taken care of at the "consumer" side of things, since this function is used in Tanstack's `useQuery`, this would then  e.g. result in an error on `pages/modules/explorer/index.tsx`;

```
const {error, data}  = useModulesQuery('mainnet01');

if (error) {
   console.error(error.message)
}
```

packages/apps/tools/src/pages/modules/explorer/utils.ts
sanderlooijenga: Agreed, will do!

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
sanderlooijenga: Which steps do you mean exactly? 😁

packages/apps/tools/src/components/Global/CustomAccordion/CustomAccordion.tsx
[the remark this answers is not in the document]
r-mulder: Yes it does for the button. But the accordion has things like expanded, a different role etc.

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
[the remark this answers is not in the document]
r-mulder: I think thats the correct UX, but you don't need to put it on the "small" button as well right?

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
r-mulder: That might be even better!

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
r-mulder: Basically call these two functions:
`            const chainModule: IChainModule = moduleModelToChainModule(data);
            setActiveModule(chainModule);`
            
            If I read it this way it seems a lot of calculation is being done before deciding that you might already have done it and it's unnecessary.

packages/apps/tools/src/hooks/use-modules-query.ts
r-mulder: Awesome! 

 I missed the useModuleQuery below..  makes a lot of sense now.

packages/apps/tools/src/pages/modules/explorer/utils.ts
sanderlooijenga: Tests and some documentation added

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
sanderlooijenga: Nothing unneccessary happens. When a module is being clicked, we always need to update the active module (`setActiveModule`). Then we add the clicked module to the `openedModules` (which is an array being used in the editor part, e.g. the tabs), but only if it wasn't already opened before. 

Does that make sense?

packages/apps/tools/src/components/Global/CustomAccordion/CustomAccordion.tsx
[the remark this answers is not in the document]
sanderlooijenga: I've done an initial attempt at improving this, can you check it out?

packages/apps/tools/src/components/Global/ModuleExplorer/index.tsx
r-mulder: Ah thinking about the other part you are going to build with the tabs it makes sense indeed. Thanks!

packages/apps/tools/src/pages/modules/explorer/utils.ts
r-mulder: Very nice!

packages/apps/tools/src/components/Global/CustomAccordion/CustomAccordion.tsx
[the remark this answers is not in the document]
r-mulder: I think it looks good!

packages/apps/tools/src/components/Global/CustomTree/CustomTree.tsx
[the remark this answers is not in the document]
sanderlooijenga: I will keep it as-is for now and will do a UI/UX walkthrough together with Isa and alter it after if needed. Alright?

javadkh2 (approved):

PR #2037, opened 2024-04-29
Notification visuals update

ferreroltd:


packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -71,14 +109,39 @@ export const closeButtonClass = style([
...
 
-export const contentClass = style([
-  atoms({
-    fontSize: 'base',
-    gap: 'xs',
-    maxWidth: 'content.maxWidth',
-    marginBlockStart: 'xxs',
-  }),
-]);
+export const contentClassRecipie = recipe({
```
r-mulder, line 112:
```suggestion
export const contentClassRecipe = recipe({
```

packages/libs/react-ui/src/components/Notification/Notification.css.ts
```diff
@@ -17,27 +17,65 @@ export const notificationRecipe = recipe({
...
+      info: [
+        atoms({
+          backgroundColor: 'semantic.info.subtle',
+          borderColor: `semantic.info.default`,
+          color: `text.semantic.info.default`,
+        }),
+        {
+          vars: {
+            [iconColorVar]:
+              tokens.kda.foundation.color.icon.semantic.info.default,
```
r-mulder, line 35: I think it's best to stay consistent and use the token function. If we ever change the structure we could fix it in one place.

r-mulder (approved):

PR #2039, opened 2024-04-30
Feat(react UI)/new select component

r-mulder:
### Changes:
updated the select component with the new styling updates. These include new size variants, font variants and consistent states like the other form elements. Internally the component uses the new Field component to abstract the formfield logic and simplify the select component.

### Breaking changes:
- startIcon has been renamed to startVisual to be consistent with other rebranded components.

### DIFF
--- a/packages/libs/react-ui/src/components/Form/Form.css.ts
+++ b/packages/libs/react-ui/src/components/Form/Form.css.ts
@@ -29,7 +29,7 @@ export const outlineColor = createVar();
 export const iconFill = fallbackVar(token('color.icon.base.@init'));
 const textColor = fallbackVar(token('color.text.base.@init'));
 
-const outlineStyles = {
+export const outlineStyles = {
   outlineStyle: `solid`,
   outlineWidth: '2px',
   outlineOffset: '2px',
@@ -48,10 +48,20 @@ export const baseContainerClass = recipe({
     {
       transition: 'outline-color 0.2s ease-in-out',
       outlineColor: 'transparent',
+      minWidth: '150px',
+      '@media': {
+        [breakpoints.sm]: {
+          maxWidth: '100%',
+        },
+        [breakpoints.md]: {
+          maxWidth: '40rem',
+        },
+      },
       selectors: {
         // outline should not be shown if there is a button which is focused
         '&:focus-within:has(button:not(button:focus))': outlineStyles,
         '&:focus-within:not(&:has(button))': outlineStyles,
+        [`&:focus-within:has(button#select-button)`]: outlineStyles,
       },
     },
   ],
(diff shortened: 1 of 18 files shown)


packages/libs/react-ui/src/components/Form/Form.css.ts
```diff
@@ -48,10 +48,20 @@ export const baseContainerClass = recipe({
...
       outlineColor: 'transparent',
+      minWidth: '150px',
+      '@media': {
+        [breakpoints.sm]: {
+          maxWidth: '100%',
+        },
+        [breakpoints.md]: {
+          maxWidth: '40rem',
+        },
+      },
```
eileenmguo, line 59: I would recommend using the `responsiveStyles` utility that we export for this

eileenmguo (commented):
Should the focus state have rounded corners? A bit strange with the sharp corners of the input
image

packages/libs/react-ui/src/components/Form/Select/SelectButton.tsx
```diff
@@ -38,6 +43,7 @@ function SelectButtonBase<T extends object>(
   return (
     <button
       {...mergeProps(buttonProps, dataProps, hoverProps)}
+      id="select-button"
```
sanderlooijenga, line 46: Should this be an optional prop where this `id` is the default/fallback? If you have multiple select components on a page it would be semantically incorrect 😁

sanderlooijenga (approved):

packages/libs/react-ui/src/components/Form/Select/SelectButton.tsx
r-mulder: Yhea I use this to target a selector but indeed, what if you have multiple.. Rookie mistake.
I'll change this to a data attribute.

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
```diff
@@ -55,3 +48,29 @@ export const selectValueClass = style([
...
+  lg: { height: token('icon.size.lg') },
+});
+
+export const selectItemClass = recipe({
+  base: [atoms({ paddingInlineStart: 'md' })],
+  variants: {
+    size: {
+      sm: [],
+      md: [],
+      lg: [],
```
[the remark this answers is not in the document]
r-mulder: I need the sizes in the compound variants to toggle the font styles, but in this case I don't have any additional styles. Do you know if there is a different way?

packages/libs/react-ui/src/components/Form/Form.css.ts
r-mulder: Changed in the next commit.

packages/libs/react-ui/src/components/Form/NumberField/NumberField.tsx
```diff
@@ -137,6 +137,7 @@ export function NumberFieldBase(
         data-invalid={props.isInvalid || undefined}
         data-positive={props.isPositive || undefined}
         data-has-start-addon={!!props.startVisual || undefined}
+        data-has-end-addon
```
ferreroltd, line 140: this is always here due to the buttons right?

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
```diff
@@ -1,47 +1,40 @@
...
-        outline: `2px solid ${token('color.border.semantic.info.@focus')}`,
-      },
-      '&[data-invalid]': {
-        outline: `2px solid ${token('color.border.semantic.negative.@focus')}`,
-      },
     },
   },
 ]);
 
+globalStyle(`${selectButtonClass} > span`, {
```
ferreroltd, line 34: do we want a central place for all global changes even when its scoped to our class? Just a thought

ferreroltd (approved):

packages/libs/react-ui/src/components/Form/NumberField/NumberField.tsx
r-mulder: Yes exactly.

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
r-mulder: I would say no, this is heavily reliant on context. If we would move this to a separate file then you would never know that it's impacting your component without a proper search.

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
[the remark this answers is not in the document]
eileenmguo: hmm not sure, are you setting a default variant for this? Just to make sure we are applying the font styles at all times

eileenmguo (approved):

packages/libs/react-ui/src/components/Form/Select/Select.css.ts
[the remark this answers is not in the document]
r-mulder: I keep the default logic in the component. So the props have defaults. Is that proper usage?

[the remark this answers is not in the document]
eileenmguo: either is fine 👍 You can also add default variants to the style definitions

PR #2044, opened 2024-05-01
Test/refactor docs test

Ghislain89:
This PR has a few improvements:

- Docs test(s) now use fixtures, instead of creating new instances of objects.
- Tree validator has been refactored to fully construct the tree using locators, this should make it easier to maintain and debug. Previously it used filters, which made the code look a little bit nicer, but was horrible to debug


.github/actions/end-to-end-tests/action.yaml
```diff
@@ -10,6 +10,10 @@ runs:
         shell: bash
         run: pnpm --filter e2e-${{ inputs.app }} run install:chromium
 
+      - name: Build dependencies for e2e tests
+        shell: bash
+        run: pnpm turbo build --filter e2e-base^...
+
       - name: Run end-to-end tests
```
Ghislain89, line 17: Need to build dependencies of e2e-base package so we can actually use things like @kadena/client. I honestly have no idea how this could've ever worked before....

sstraatemans (approved):

PR #2048, opened 2024-05-01
Adjust reconcile view amount fields to decimals

nil-amrutlal-dept:
In Marmalade, non fungible token balances can be decimal. Due to the view having these fields as integer. To correct this, I am adding a migration to drop the view and recreate it with a decimal column type

---
  - https://app.asana.com/0/0/1207100487482658


MRVDH (approved):

enobayram (approved):
I've reviewed the new DB migration and it LGTM!

PR #2055, opened 2024-05-02
[@kadena/client-utils] Implement Marmalade functions - Sale functions

wooglie:
This PR adds all functions required for Marmalade sale, including conventional and dutch sale.

Every sale type has it's own test file and a test file showcasing how to use gas station to fund gas fees.

This is an extension to #1945 PR and follows the same idea with dynamic typescript types so that while writing the command, it shows you all the arguments needed for certain action.


nil-amrutlal-dept (approved):

packages/libs/client-utils/src/marmalade/config.ts
```diff
@@ -0,0 +1,335 @@
...
+    pred: BuiltInPredicate;
+  };
+  transferGuard?: {
+    keys: string[];
+    pred: BuiltInPredicate;
+  };
+}
+
+export interface ICollectionInfoInput {
+  collectionId: string;
```
ggobugi27, line 79: Do we have a way to `create-collection-id` and `create-collection` in this library? It's required for tokens configured with `collections`, so I think we should provide this feature.

packages/libs/client-utils/src/marmalade/create-token.ts
```diff
@@ -88,7 +93,7 @@ const generatePolicyTransactionData = (
   if (policyConfig?.hasRoyalty) {
     data.push(addData('fungible', props.royalty.fungible));
     data.push(addData('creator', props.royalty.creator));
-    data.push(addData('creator-guard', props.royalty.creatorGuard));
+    data.push(addData('creator-guard', props.royalty.creator.keyset));
     data.push(addData('royalty-rate', props.royalty.royaltyRate.decimal));
   }
```
ggobugi27, line 99: In #L100, when checking if the config sets upgradeableURI to false, we need an enforcement that `uri_guard` is set t a failure guard.

packages/libs/client-utils/src/marmalade/get-auction-details.ts
```diff
@@ -0,0 +1,59 @@
...
+interface IGetAuctionDetailsInput {
+  auctionConfig: IAuctionConfig;
+  saleId: string;
+  chainId: ChainId;
+  guard: {
+    account: string;
+  };
+}
+
+const getAuctionDetailsCommand = ({
```
ggobugi27, line 18: `get-auction-details` need not be a `/send` tx, but will most of the time be called as a `/local` tx, so I don't think we need to input a `guard` in a getter function.

ggobugi27: Same comment would go to all the `get-*` functions and the functions not to be sent to the blockchain: 
`create-bid-id`, `create-token-id`, `escrow-account`, `get-account-details`, `get-bid`, `get-current-price`, `get-escrow-account`, `get-token-balance`, `get-token-info`, `get-uri`

packages/libs/client-utils/src/marmalade/config.ts
```diff
@@ -0,0 +1,335 @@
...
+}
+
+export interface ICreateTokenPolicyConfig {
+  customPolicies?: boolean;
+  upgradeableURI?: boolean;
```
ggobugi27, line 84: Our function is named as `update-uri`, and we've named the interface that contains the update-uri functions as `kip.updatable-uri-policy-v1`, so I suggest using `updatableURI` in this tool, as well.

ggobugi27 (changes requested):

packages/libs/client-utils/src/marmalade/get-auction-details.ts
wooglie: It's already using a `/local` call using `dirtyReadClient` but you're right, we don't need `guard` so I removed it.

packages/libs/client-utils/src/marmalade/config.ts
wooglie: Yeah, good thinking! I've added both functions and additional test file for collection-policy.

ggobugi27 (approved):

PR #2056, opened 2024-05-03
feat(pou): Limit image size 2Mb max before NFT storing

Likita:


sstraatemans (approved):

Ghislain89:
@sstraatemans @Likita , if I understand the PR correctly. Additional validation was added to the `api/upload` endpoint that ensures the picture taken is not above 2MB.

When checking the feature, it seems to work fine as far as I can tell, but it did strike me that this validation is done fairly late: e.g. the upload only happens *after* everyone signed. If you have a large group sign only to find out after that the picture is too big...not a fun user experience I think?

PR #2061, opened 2024-05-03
feat: initial rebranding and added pagination for tabs component

r-mulder:
This PR will update the tabs component to the new brand update.

Also included:
- Pagination for the tabs list. 
- focus on active item when rendered.
- Support scroll and pagination on all devices.

### DIFF
--- a/packages/libs/react-ui/src/components/Tabs/Tabs.css.ts
+++ b/packages/libs/react-ui/src/components/Tabs/Tabs.css.ts
@@ -1,5 +1,5 @@
 import { style, styleVariants } from '@vanilla-extract/css';
-import { responsiveStyle, token } from '../../styles';
+import { token } from '../../styles';
 import { atoms } from '../../styles/atoms.css';
 import { tokens } from '../../styles/tokens/contract.css';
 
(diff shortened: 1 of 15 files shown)


eileenmguo (commented):
looks like the border doesn't go all the way across the tab container
image

Also, I don't see an example of pagination in storybook.

packages/libs/react-ui/src/components/Tabs/Tabs.css.ts
```diff
@@ -85,13 +59,89 @@ export const selectorLine = style([
...
+          color: token('color.text.base.@active'),
+        },
+        '&[data-hovered="true"]': {
+          color: token('color.text.base.@hover'),
+        },
+      },
+    },
+  ],
+  variants: {
+    inverse: {
```
ferreroltd, line 100: I wonder if this was the dark/light mode things, probably not but still curious

r-mulder: No this is to support a nested structure, A UX pattern is to have a lighter background color for the active tab so the inner tabs component can have the background color of the outer active tab.

eileenmguo:
image
 The tabs look a bit broken here. Is this the intended design? @isaozler

isaozler (approved):

eileenmguo (commented):
Code looks good! 

The storybook looks a bit broken because the close button doesn't do anything. Could you add some state management so that the close button works in the example?

sstraatemans (approved):

PR #2063, opened 2024-05-03
feat(docs): importchangelogs

sstraatemans:
https://app.asana.com/0/1205619636589593/1205619640973427/f


packages/apps/docs/src/github.td.ts
```diff
@@ -0,0 +1,172 @@
+interface IGitHubFile {
+  sha: string;
+  filename: string;
```
realdreamer, line 3: the file extension should be `github.d.ts`, isn't it?
can't we use octokit types are other related types definition from octokit?

packages/apps/docs/src/changelog.td.ts
```diff
@@ -0,0 +1,68 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type IChangelogComplete = Record<string, IChangelogPackage>;
+
+interface IRepo {
+  name: string;
+  repo: string;
```
realdreamer, line 6: the file extension should be `github.d.ts`, isn't it?

packages/apps/docs/src/scripts/importChangelogs/index.ts
```diff
@@ -0,0 +1,220 @@
...
+  await getGitHubData(content);
+  enrichContent(content);
+
+  if (!errors.length) {
+    writeContent(content);
+    success.push('Changelogs imported');
+  }
+
+  await runPrettier();
+  console.log({ errors });
```
realdreamer, line 212: is it intended?

packages/apps/docs/src/scripts/importChangelogs/utils/reduce.ts
```diff
@@ -0,0 +1,75 @@
...
+    avatar_url: user.avatar_url,
+    url: user.url,
+  };
+
+  return reduced;
+};
+
+export const getAuthor = (
+  commit: IGithubCommitDataCommit,
+): IChangelogUser | undefined => {
```
realdreamer, line 22: imo this `IChangelogUser` type could be improved because currently this makes it error prone because `id`, `url` and `avatar_url` not always available as part of the return data.

so making these fields as optional as part of return data it helps to safety check wherever we're consuming it

packages/apps/docs/src/scripts/importChangelogs/index.ts
sstraatemans: yes. it was part of the build scripts. but will keep it out. so I wanted a quick print

packages/apps/docs/src/scripts/importChangelogs/utils/reduce.ts
sstraatemans: I think I made an improvement

packages/apps/docs/src/changelog.td.ts
sstraatemans: doh.

packages/apps/docs/src/scripts/importChangelogs/utils/reduce.ts
realdreamer: you can remove the double casting from here https://github.com/kadena-community/kadena.js/pull/2063/files/be7a1835316c8e528a120d104ded95d6f060cb8b#diff-9475c0216adf8d4ed45a1637fbaf1c4aa3f19a1aa6df32ee145ae961f972242eR25-R29

packages/apps/docs/src/scripts/importChangelogs/index.ts
```diff
@@ -0,0 +1,220 @@
...
+};
+
+const createContent = async (repo: IRepo) => {
+  const md: Root = remark.parse(getChangelog(repo));
+
+  const repoCrawler = await crawl(repo);
+  return await repoCrawler(md);
+};
+
+const getContent = async (repos: IRepo[]): Promise<IChangelogComplete> => {
```
realdreamer, line 164:
```suggestion
const getReposContent = async (repos: IRepo[]): Promise<IChangelogComplete> => {
```

packages/apps/docs/src/scripts/importChangelogs/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+    return content;
+  };
+
+  innerCrawl(tree);
+
+  return content;
+};
+
+const createRecord = (content: Node): IChangelogVersionRecord => {
```
realdreamer, line 70: maybe adding some more context on the method name, it would be better to understand what it does

packages/apps/docs/src/scripts/importChangelogs/utils/__tests__/prs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      const pkg = changelog['Pact 4'] as unknown as IChangelogPackage;
+      const result = getPrs(pkg);
+
+      expect(result[0].id).toEqual(1326);
+    });
+  });
+
+  describe('getVersionPRs', () => {
+    it('should return an array with all commits', async () => {
```
realdreamer, line 78:
```suggestion
    it('should return an array with all commits', async () => {
```
this could be better wording to describe what it does return because even the above test cases `getPrs` also have the same test describes but the return data is different, so it makes sense to add that context. 

even for below test scenarios

packages/apps/docs/src/scripts/importChangelogs/utils/__tests__/prs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      const pr = {
+        id: 1337,
+        tries: 3,
+        commits: [],
+      };
+
+      expect(filterPRsWithoutData(pr)).toEqual(false);
+    });
+
+    it('should return false if the commit has tried 2 times already, but there is data data', () => {
```
realdreamer, line 125:
```suggestion
    it('should return false if the commit has tried 2 times already, but there is data', () => {
```

packages/apps/docs/src/scripts/importChangelogs/utils/__tests__/prs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+        id: 1337,
+        tries: 2,
+        commits: [],
+        data: {} as IChangelogPRData,
+      };
+
+      expect(filterPRsWithoutData(pr)).toEqual(false);
+    });
+
+    it('should return true if the commit has tried 1 times and there is no data ', () => {
```
realdreamer, line 136:
```suggestion
    it('should return true if the commit has tried 1 time and there is no data ', () => {
```

packages/apps/docs/src/scripts/importChangelogs/utils/enrichContent.ts
```diff
@@ -0,0 +1,97 @@
...
+      if ((acc && lastModifiedDate && isAfter(lastModifiedDate, acc)) || !acc) {
+        return lastModifiedDate;
+      }
+
+      return acc;
+    },
+    undefined,
+  );
+
+  const prssDate = getVersionPRs(version).reduce(
```
realdreamer, line 64:
```suggestion
  const prsDate = getVersionPRs(version).reduce(
```

packages/apps/docs/src/scripts/importChangelogs/utils/enrichContent.ts
```diff
@@ -0,0 +1,97 @@
...
+    undefined,
+  );
+
+  if (!prssDate) return commitsDate;
+  if (!commitsDate) return prssDate;
+
+  return prssDate > commitsDate ? prssDate : commitsDate;
+};
+
+export const enrichContent = (content: IChangelogComplete) => {
```
realdreamer, line 85:
```suggestion
export const enrichPackageContent = (content: IChangelogComplete) => {
```
or maybe `enrchChangelogContent`

packages/apps/docs/src/scripts/importChangelogs/utils/github.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    for (let i = 0; i < prs.length; i++) {
+      const pr = prs[i];
+
+      await getPRData(pkg, pr);
+      writeContent(content);
+    }
+
+    //check how many commits and prs do not have data
```
realdreamer, line 33: adding context of when this scenario occurs, might be useful

packages/apps/docs/package.json
```diff
@@ -24,10 +24,11 @@
     "lint:fmt": "prettier . --cache --check",
     "lint:pkg": "lint-package",
     "lint:src": "eslint src --ext .js,.ts,.jsx,.tsx,.mjs",
+    "prettier": "prettier",
     "start": "next start",
     "test": "vitest run",
     "test:coverage": "vitest run --coverage",
-    "test:watch": "vitest",
+    "test:watch": "vitest  --coverage",
```
realdreamer, line 31:
```suggestion
    "test:watch": "vitest  --coverage",
```
we already have coverage separately then why we need with coverage again here

realdreamer (approved):

packages/apps/docs/package.json
sstraatemans: i like to see the coverage during watch and development

PR #2067, opened 2024-05-06
bugfix(hd-wallet): chainweaver sign did not use the right hash

barthuijgen:
* Chainweaver sign expects base64 decoded hash as a "message".

### DIFF
--- a/packages/libs/hd-wallet/src/chainweaver/tests/chainweaver.test.ts
+++ b/packages/libs/hd-wallet/src/chainweaver/tests/chainweaver.test.ts
@@ -6,6 +6,7 @@ import {
   kadenaGenKeypair,
   kadenaGenMnemonic,
   kadenaMnemonicToRootKeypair,
+  kadenaSign,
   kadenaSignFromRootKey,
 } from '../index.js';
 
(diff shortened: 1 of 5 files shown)


nillo (approved):

nillo (approved):

realdreamer (approved):

packages/libs/hd-wallet/src/chainweaver/tests/chainweaver.test.ts
```diff
@@ -135,7 +136,17 @@ describe('kadenaGenKeypair', () => {
...
     expect(signature).toBeTruthy();
     expect(Buffer.from(signature).toString('hex')).toBe(
-      '4a9a35634ca8b1efd0d84053bf73d78dda4e59223d230869ea7abbee8f923b723538f9a3ed13de31c8d629c97205a3becb15f5fdc71dc1dc4e42596b9a10d906',
+      'e3d3ffe85877a181487ef66acd4e2b5968fa4d055d81efcce812cfcee855d42547a962dc6a28a445f9da696ad2270b046b27a462d1f1ced6c7408b65a1a7fa0c',
+    );
+  });
+  it('should generate a valid signature with keypair', async () => {
+    const rootKey = await kadenaMnemonicToRootKeypair(PASSWORD, MNEMONIC);
+    const keypair = await kadenaGenKeypair(PASSWORD, rootKey, 1);
+    const signature = await kadenaSign(PASSWORD, 'hello', keypair.secretKey);
```
javadkh2, line 145: though `hello` is a valid base64 string but it still confusing here, 
it would be more readable like this
```
const base64hash = `aaaa`
const signature = await kadenaSign(PASSWORD, base64hash, keypair.secretKey);
```

barthuijgen: For the best clarity you'd actually just want to do a real base64 encode I think. I indeed missed this making the PR sorry, with real values it's a base64 decoded blake2b hash, where the hash was never encoded as base64 (afaik) and just happens to be able to be decodable. The whole thing seems odd to me, but apparently this is how it should be.

PR #2068, opened 2024-05-07
E2E test improvements and fixes

Ghislain89:
This PR has a few changes:

1. Reduce the amount of components used in E2E tests, specifically for the Tools dApp.
2. Removes I18n translations in our page objects to improve readability. Also there is nothing on the roadmap (as far as I know) about multilingual support.
3. Remove path aliases for the e2e-base package
4. Precompiles applications instead of using `next dev` to start them. This makes starting the webserver for the tests a lot faster, removing the need for extremely long timeouts on our tests.

**Reduced the amount of components**
Previously I invested quite some time in setting up component page objects for specific UI components that where present on multiple pages. From previous experience, I know that heavily relying on composition can make the test suite easier to maintain as well as easier to understand. As I sort of expected us to have way more dApps, I used the same structure here.

As it turns out, we don't have that many dApps and we can use accessibility locators so the applying the composite pattern does not give us as much value as it would in very large projects. To increase readability and usability, i've removed them for now. Should the need arise in the future, we can always move certain page interactions to components again.

**Remove path aliases for the e2e-base package**
I initially added path aliases to the e2e-base package in order to make imports look nicer. Unfortunately this had unforeseen side effects. We would lose types in every third child and would remove the possibility to navigate to the function or page definition. While I personally know where to find the definition, this is likely  harder for someone that is less familiar with these set of packages.  In order to restore this, I removed path aliases from the base package.
image
image.

Took a while to figure this one out...meh.

**pre-compile apps**
Previously, apps where started in ci using `next dev`. Primarily because we wanted to set some environment variables specifically for E2E tests and NextJS wants it's environment variables/config on **build time** instead of on deployment of the app.  While this worked from a functional perspective, it was painfully slow and required a long timeout on starting the webserver as well as the tests. To improve this I wanted to build the apps first and then use `next start` to start them up.  According to the [NextJS docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#test-environment-variables), you need to set NODE_ENV to test.  This was specifically tricky in CI as when using Turbo to build say the Docs app, setting NODE_ENV to test will also do this for all of its dependencies which has a ton of side-effects.

I ended up adding a turbo task to build apps with a specific script `build:e2e` that sets NODE_ENV to test _just for the app_ while depending on the **build** script for dependencies.


packages/e2e/e2e-base/src/page-objects/react-ui/card.component.ts
```diff
@@ -1,43 +1,42 @@
-import type { Locator, Page } from '@playwright/test';
-import { ListBoxComponent } from './listBox.component';
+// import type { Locator, Page } from '@playwright/test';
```
sstraatemans, line 1: is this all needed?

packages/e2e/e2e-base/playwright.config.ts
```diff
@@ -18,9 +18,7 @@ export const baseConfig: PlaywrightTestConfig = {
     trace: 'retain-on-failure',
     viewport: { width: 1920, height: 1080 },
   },
-  timeout: 100000,
-
   expect: {
-    timeout: 2 * 30000,
+    timeout: 15000,
```
sstraatemans, line 22: 🔥

packages/e2e/e2e-base/src/page-objects/react-ui/card.component.ts
Ghislain89: This file has been made obsolete and should be removed. Commenting it was my way of saying: I need to re-implement this logic elsewhere.  Good find!

packages/e2e/e2e-base/playwright.config.ts
Ghislain89: Now that we pre-compile apps, this is finally possible. This is particularly useful when a test fails, we don't have to wait for this long for feedback.

PR #2072, opened 2024-05-07
feat(kadena-cli) fixing test of transactions

nillo:
- txTest now gets network id and chain Id from transaction
- txSend/txTest now share generic functionality
- output is printed more structured now


packages/tools/kadena-cli/src/tx/utils/txDisplayHelper.ts
```diff
@@ -34,53 +36,113 @@ export async function printTx(
...
+    displayCustomTransactionInfo(
+      {
+        fileName: files[i],
+        transactionHash: item.transaction.hash,
+      },
+      baseIndent,
+    );
+    log.info('\n');
+    if (item.response) {
+      log.info(log.color.black(`${' '.repeat(baseIndent)}Response:`));
```
realdreamer, line 69:
```suggestion
      log.info(log.color.black(`${' '.repeat(baseIndent)}Response:`));
```
can't we use padStart instead of using `black` color with indentation

PR #2073, opened 2024-05-07
feat(kadena-cli): change how config init works and add search for kad…

barthuijgen:
…nea dir


packages/tools/kadena-cli/src/services/config/config.service.ts
```diff
@@ -62,11 +71,11 @@ export interface IConfigService {
...
   private directory: string | null = null;
-  public constructor(
-    // eslint-disable-next-line @typescript-eslint/naming-convention, @typescript-eslint/parameter-properties
-    private services: Services,
-    directory?: string,
-  ) {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  private services: Services;
+
+  public constructor(services: Services, directory?: string) {
```
nillo, line 77: just call i IServices. so u dont have to disable lint. ( yeah i don't like it either but, better hthen the lint error)

nillo (approved):

realdreamer (approved):

packages/tools/kadena-cli/src/services/config/config.service.ts
barthuijgen: the lint is about the variable name, it should be `_services` because its private.

PR #2075, opened 2024-05-07
feat(account): remove on chain validation for add from wallet option and make conditional validation for manual

realdreamer:
---
  - https://app.asana.com/0/0/1207178351013591

### DIFF
--- a/packages/tools/kadena-cli/src/account/accountOptions.ts
+++ b/packages/tools/kadena-cli/src/account/accountOptions.ts
@@ -20,6 +20,7 @@ import { isEmpty } from './utils/addHelpers.js';
 export const accountOptions = {
   accountTypeSelection: createOption({
     key: 'type' as const,
+    defaultIsOptional: false,
     prompt: account.accountTypeSelectionPrompt,
     validation: z.string(),
     option: new Option(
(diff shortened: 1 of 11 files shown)


packages/tools/kadena-cli/src/account/accountOptions.ts
```diff
@@ -213,9 +214,9 @@ export const accountOptions = {
     defaultIsOptional: false,
     prompt: account.publicKeysForAccountAddPrompt,
     expand: async (publicKeys: string): Promise<string[]> => {
-      const keys = publicKeys.split(',');
+      const keys = publicKeys?.split(',');
       return keys
-        .map((key: string) => key.trim())
+        ?.map((key: string) => key.trim())
         .filter((key: string) => !isEmpty(key));
```
nillo, line 220: how would a empty key be still available here? shouldnt it be restricted at validation?

packages/tools/kadena-cli/src/account/accountOptions.ts
```diff
@@ -224,4 +225,13 @@ export const accountOptions = {
       'Public keys to add to account',
     ),
   }),
+  confirmAccountVerification: createOption({
+    key: 'verify' as const,
+    validation: z.string(),
+    prompt: account.confirmAccountVerificationPrompt,
+    option: new Option(
+      '-v, --verify <verify>',
```
barthuijgen, line 233: is this supposed to be a boolean flag? or is there an actual value you're expecting to get. I think it should be `'-v, --verify'`

realdreamer: I'm expecting the value `true` or `false` explicitly from the user

barthuijgen: That's not consistent then though, i've used the boolean-flag on other commands where you would just have to pass `--verify`.

packages/tools/kadena-cli/src/account/accountOptions.ts
realdreamer: when you pass --quiet flag and it's an optional field. it could be empty.

packages/tools/kadena-cli/src/account/accountOptions.ts
realdreamer: okay! makes sense! will update it

nillo (approved):

barthuijgen (approved):

PR #2076, opened 2024-05-07
Feat/dialog sizes

ferreroltd:
this pr adds the sizes for the dialog

### DIFF
--- a/packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
+++ b/packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
@@ -49,7 +49,7 @@ A Dialog is a type of modal that is used to display information or prompt the us
       control: {
         type: 'radio',
       },
-      options: ['compact', 'base', 'large'],
+      options: ['sm', 'md', 'lg'],
       defaultValue: 'base',
     },
   },
--- a/packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
+++ b/packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
@@ -3,9 +3,10 @@ import React, { useState } from 'react';
 import { Button } from '../Button';
 import type { IDialogProps } from '../Dialog';
 import { Dialog, DialogContent, DialogFooter, DialogHeader } from '../Dialog';
+import { DialogHeaderSubtitle } from './DialogHeaderSubtitle';
 import { ModalContent } from './StoryComponents';
 
-type DialogStoryProps = IDialogProps & { title: string };
+type DialogStoryProps = IDialogProps & { title: string; subtitle: string };
 
 const meta: Meta<DialogStoryProps> = {
   title: 'Overlays/Dialog',
@@ -62,8 +63,9 @@ export const DialogStory: Story = {
   name: 'Dialog',
   args: {
     title: 'Dialog Title',
+    subtitle: 'Dialog subtitle',
   },
-  render: ({ title, size }) => {
+  render: ({ title, subtitle, size }) => {
     const [isOpen, setIsOpen] = useState(false);
 
     return (
@@ -76,6 +78,7 @@ export const DialogStory: Story = {
         >
           {(state) => (
             <>
+              <DialogHeaderSubtitle>{subtitle}</DialogHeaderSubtitle>
               <DialogHeader>{title}</DialogHeader>
               <DialogContent>
                 <ModalContent />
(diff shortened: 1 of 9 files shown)


packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -45,6 +45,13 @@ A Dialog is a type of modal that is used to display information or prompt the us
         type: { summary: 'string' },
       },
     },
+    size: {
+      control: {
+        type: 'radio',
+      },
+      options: ['compact', 'base', 'large'],
```
r-mulder, line 52: I just put it here because it's readable but I don't like the naming for the 3 variants. What if we would add another size to this array which is smaller than compact? I think that we should be consistent and only use the compact naming if it's a more condensed look and basically the only alternative to the regular size. (so only when we have the default styles and have a smaller version)

For this situation I would suggest to use t shirt sizing. But perhaps check the other components how we named them to be consistent.

packages/libs/react-ui/src/components/Dialog/Dialog.tsx
```diff
@@ -79,6 +96,7 @@ export const Dialog: FC<IDialogProps> = ({
     defaultOpen,
     onOpenChange,
   });
+  console.log(props);
```
r-mulder, line 99: Another log!

packages/libs/react-ui/src/components/Modal/Modal.css.ts
```diff
@@ -10,7 +10,7 @@ export const underlayClass = style([
     inset: 0,
   }),
   {
-    // TODO: Update to use token
-    backgroundColor: 'rgba(26, 26, 26, 0.8)',
+    backdropFilter: 'blur(12px)',
+    backgroundColor: 'background.base.default',
```
r-mulder, line 14: You will have to move this line into the atoms function.

packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
ferreroltd: I used the naming from the figma documentation from isa, I also thought we should use the normal sizes (or standard) but if they go and read the docs would be confusing. Ill ask isa to update and make a separate pr for that given he is 1 week off

packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -45,6 +46,13 @@ A Dialog is a type of modal that is used to display information or prompt the us
         type: { summary: 'string' },
       },
     },
+    size: {
+      control: {
+        type: 'radio',
+      },
+      options: ['sm', 'md', 'lg'],
+      defaultValue: 'base',
```
r-mulder, line 54: You forgot to change the default value after changing the options.

packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -55,19 +63,22 @@ export const DialogStory: Story = {
...
       <>
         <Button onClick={() => setIsOpen(true)}>Modal Trigger</Button>
         <Dialog
           isOpen={isOpen}
+          size={size}
           onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
         >
           {(state) => (
             <>
+              <DialogHeaderSubtitle>{subtitle}</DialogHeaderSubtitle>
```
r-mulder, line 81: why is the subtitle above the title?

packages/libs/react-ui/src/components/Dialog/Dialog.tsx
```diff
@@ -10,18 +10,37 @@ import { useOverlayTriggerState } from 'react-stately';
...
+  sm: smClass,
+  md: mdClass,
+  lg: lgClass,
+};
 interface IBaseDialogProps
   extends Omit<IModalProps, 'children'>,
     AriaDialogProps {
   children?: ((state: OverlayTriggerState) => ReactNode) | ReactNode;
   className?: string;
+  size?: 'sm' | 'md' | 'lg';
```
r-mulder, line 31: I suggest to keep it a bit more dynamic, for example call the classMap sizeMap and use the keyof typeof sizeMap here.

packages/libs/react-ui/src/components/Dialog/Dialog.stories.tsx
ferreroltd: its meant like that in the design

packages/libs/react-ui/src/components/Dialog/Dialog.tsx
ferreroltd: i like it but since i didnt export it i thought this would make it faster to spot the values, but im okay with doing this

r-mulder (approved):

PR #2085, opened 2024-05-08
feat: basic app setup and create layouts with design

lkononenko:
- created layoutMini for all pages before customer is logged in the wallet as it has a different design from pages after the login
- added color theme switcher, favicon


packages/apps/dev-wallet/src/App/layout-mini.css.ts
```diff
@@ -0,0 +1,32 @@
+import { atoms, tokens, vars } from '@kadena/react-ui/styles';
+import { style } from '@vanilla-extract/css';
+import { globalStyle } from '@vanilla-extract/css';
+
+globalStyle('body', {
```
ferreroltd, line 5: I would move this to a globals overwrite or something

lkononenko: sounds better indeed, just moved it

packages/apps/dev-wallet/src/App/layout.tsx
```diff
@@ -31,14 +42,26 @@ export const Layout: FC = () => {
         }
       >
         <NavHeaderLinkList>
-          <Link to="/">
-            <Text bold>DX-Wallet</Text>
-          </Link>
-          <Link to="/networks">
-            <Text bold>Network</Text>
-          </Link>
+          <NavHeaderLink href="/">
```
ferreroltd, line 45: we need to use react-router links instead

ferreroltd (approved):
left some remarks but looks good!

packages/apps/dev-wallet/src/App/layout.tsx
lkononenko: updated!

packages/apps/dev-wallet/src/App/routes.tsx
```diff
@@ -41,33 +42,37 @@ const ProtectedRoute: FC<
...
+            path="/backup-recovery-phrase/:keySourceId/write-down"
+            element={<WriteDownRecoveryPhrase />}
+          />
+          <Route
+            path="/account-discovery/:keySourceId"
+            element={<AccountDiscovery />}
+          />
+          ,
+          <Route path="/accounts/:account" element={<p>Account</p>} />,
+          <Route path="/sig-builder" element={<SignatureBuilder />} />,
```
javadkh2, line 67: I think these two routes should go to the normal layout

lkononenko: updated

javadkh2 (approved):

PR #2088, opened 2024-05-08
Block and Transaction Queries, Subscriptions and Fragments

nil-amrutlal-dept:
---
  - https://app.asana.com/0/0/1207265113893282


packages/apps/explorer/src/graphql/queries/transaction.graph.ts
```diff
@@ -0,0 +1,39 @@
+import { gql } from '@apollo/client';
+import type { DocumentNode } from 'graphql';
+
+export const transaction: DocumentNode = gql`
+  query transaction(
```
MRVDH, line 5:
```suggestion
export const transactions: DocumentNode = gql`
  query transactions(
```

packages/apps/explorer/src/graphql/subscriptions/newBlocks.graph.ts
```diff
@@ -0,0 +1,17 @@
+import { gql } from '@apollo/client';
+import type { DocumentNode } from 'graphql';
+
+export const newBlocksSubscription: DocumentNode = gql`
```
MRVDH, line 4:
```suggestion
export const newBlocks: DocumentNode = gql`
```

The generated code will already postfix these with Query and Subscription

MRVDH (changes requested):

MRVDH (approved):

PR #2091, opened 2024-05-08
Restore generated chainweb openapi, consolidate chainweb-node and cha…

lsgunnlsgunn:
…in-stream into two topics instead of three, update config.yaml accordingly.

Accidentally deleted the generated API doc. Restore the imported version until I can take a different stab at providing this information.


lsgunnlsgunn:
@LindaOrtega Can you take a look and approve? I didn't realize that the generated Chainweb node API got left out of the party right as we were merging the new structure PR. This PR restores the missing file and some related changes to the config file and other topics in the section.

sstraatemans (approved):

PR #2105, opened 2024-05-10
fix(kadena-cli): update readme + fix network command

realdreamer:
- make `--network-delete` as required flag when user use `--quiet` flag
- network list provide better error message for empty state
- fix/update readme

### DIFF
--- a/packages/tools/kadena-cli/src/account/commands/accountFund.ts
+++ b/packages/tools/kadena-cli/src/account/commands/accountFund.ts
@@ -96,7 +96,10 @@ export const createAccountFundCommand = createCommand(
 
         const { deployFaucet } = await option.deployFaucet();
 
-        if (deployFaucet === false) {
+        if (!deployFaucet) {
+          log.info(
+            'Skipping faucet deployment... If you would like to deploy the faucet, please re-run the command with the "--deploy-faucet" flag.',
+          );
           return;
         }
 
--- a/packages/tools/kadena-cli/src/networks/utils/networkDisplay.ts
+++ b/packages/tools/kadena-cli/src/networks/utils/networkDisplay.ts
@@ -1,4 +1,7 @@
-import { networkDefaults } from '../../constants/networks.js';
+import {
+  NO_NETWORKS_FOUND_ERROR_MESSAGE,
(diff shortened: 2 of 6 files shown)


nillo (approved):

nillo:
we should decide if we still gonna merge the technical changes before monday @barthuijgen @realdreamer

realdreamer:
> we should decide if we still gonna merge the technical changes before monday @barthuijgen @realdreamer

imo we can merge it i dont see any major change but if we're concern about i can split this PR into two PRs one for updating readme and copy change

packages/tools/kadena-cli/src/networks/utils/networkDisplay.ts
```diff
@@ -37,6 +40,10 @@ export async function displayNetworksConfig(): Promise<void> {
   const defaultNetworkName = await getDefaultNetworkName();
   const existingNetworks: ICustomNetworkChoice[] = await getExistingNetworks();
 
+  if (existingNetworks.length === 0) {
+    return log.error(NO_NETWORKS_FOUND_ERROR_MESSAGE);
```
barthuijgen, line 44: I don't think no networks is an error.

packages/tools/kadena-cli/src/account/commands/accountFund.ts
```diff
@@ -96,7 +96,10 @@ export const createAccountFundCommand = createCommand(
 
         const { deployFaucet } = await option.deployFaucet();
 
-        if (deployFaucet === false) {
+        if (!deployFaucet) {
+          log.info(
+            'Skipping faucet deployment... If you would like to deploy the faucet, please re-run the command with the "--deploy-faucet" flag.',
```
barthuijgen, line 101: The wording seems a bit odd. it says `skipping faucet` as if it will continue to do something else.
but if faucet needs to be deployed and you choose not to, nothing else can happen right?

realdreamer: yes, nothing else can happen, it's just exit.

realdreamer:
closing this PR in favour of not including functional changes

PR #2121, opened 2024-05-13
feat: add variant options with new colors

r-mulder:
In this PR I've updated the divider to support the new colors and have the options to choose which one to use. 

Different colors are being used when using the new variant prop.

### DIFF
--- a/.changeset/shiny-crews-sleep.md
+++ b/.changeset/shiny-crews-sleep.md
@@ -2,4 +2,4 @@
 "@kadena/react-ui": minor
 ---
 
-Add variant option for different colors
+Added a new variant to apply different colors to the Divider component
--- a/.changeset/shiny-crews-sleep.md
+++ b/.changeset/shiny-crews-sleep.md
@@ -1,0 +1,5 @@
+---
+"@kadena/react-ui": minor
+---
+
+Add variant option for different colors
--- a/packages/libs/react-ui/src/components/Divider/Divider.css.ts
+++ b/packages/libs/react-ui/src/components/Divider/Divider.css.ts
@@ -1,6 +1,6 @@
 import { recipe } from '@vanilla-extract/recipes';
+import { token } from '../../styles';
 import { atoms } from '../../styles/atoms.css';
-import { tokens } from '../../styles/index';
 
 export const dividerClass = recipe({
   base: [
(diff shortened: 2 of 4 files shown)


ferreroltd (approved):

packages/libs/react-ui/src/components/Divider/Divider.css.ts
```diff
@@ -1,15 +1,34 @@
...
-    height: tokens.kda.foundation.border.width.hairline,
+export const dividerClass = recipe({
+  base: [
+    atoms({
+      width: '100%',
+      marginBlock: 'lg',
+      border: 'none',
+    }),
+    {
+      height: tokens.kda.foundation.border.width.hairline,
```
ferreroltd, line 13: do we prefer tokens over token() then

.changeset/shiny-crews-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kadena/react-ui": minor
+---
+
+Add variant option for different colors
```
eileenmguo, line 5: I don't think this is descriptive enough for someone reading release docs. Maybe something like "Added a new variant to apply different colors to the Divider component"

packages/libs/react-ui/src/components/Divider/Divider.css.ts
```diff
@@ -1,15 +1,35 @@
...
+        backgroundColor: token('color.border.base.default'),
+      },
+      bold: {
+        backgroundColor: token('color.border.base.bold'),
+        boldest: {
+          backgroundColor: token('color.border.base.boldest'),
+        },
+        highContrast: {
+          backgroundColor: token('color.border.base.high-contrast'),
+        },
```
eileenmguo, line 31: I feel like these are at the wrong level?

r-mulder: You mean that they should be in the atoms?

.changeset/shiny-crews-sleep.md
r-mulder: Agreed, changed.

packages/libs/react-ui/src/components/Divider/Divider.css.ts
r-mulder: nvm, I just noticed that I make a mistake while resolving Sams comment.

PR #2126, opened 2024-05-13
Split off cli-config as a separate PR

lsgunnlsgunn:


packages/apps/docs/src/config.yaml
barthuijgen, line 1: I don't think you meant to delete this entire file?

PR #2130, opened 2024-05-14
Added `endHeight` to `blocksFromHeight` query

nil-amrutlal-dept:
- Added `endHeight` to `blocksFromHeight` query
- Added additional validation on this field

---
  - https://app.asana.com/0/0/1207314477196992


packages/apps/graph/src/graph/query/blocks-from-height.ts
```diff
@@ -27,6 +33,12 @@ builder.queryField('blocksFromHeight', (t) =>
         },
       }),
     },
+    validate(args) {
+      if (args.endHeight && args.startHeight > args.endHeight) {
+        return false;
+      }
+      return true;
+    },
```
MRVDH, line 41: Do we return an proper error message in the response when the value is incorrect?

Also this could be simplified:

```suggestion
    validate: args => !(args.endHeight && args.startHeight > args.endHeight)
```

MRVDH (changes requested):

packages/apps/graph/src/graph/query/blocks-from-height.ts
nil-amrutlal-dept: So instead of validating it here, and throwing an error if it fails, I just passed this validation to the resolver and I threw an input error with the correct custom message

nil-amrutlal-dept: Let me know if it looks good

MRVDH (approved):

PR #2133, opened 2024-05-15
Feat / Dev wallet/ onboarding

Likita:
- Styles for onboarding first step


packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -1,34 +1,58 @@
...
+    confirmation: string;
   }>();
   const { createProfile, isUnlocked, createKey, createKAccount } = useWallet();
   const { activeNetwork } = useNetwork();
   const [createdKeySource, setCreatedKeySource] = useState<IKeySource>();
   const { createHDWallet } = useHDWallet();
+  const [isSubmitDisabled, setIsSubmitDisabled] = useState<boolean>(true);
+  const passwordRef = useRef<HTMLInputElement>(null);
+  const passwordConfirmationRef = useRef<HTMLInputElement>(null);
+  const validate = (value: string) => {
```
javadkh2, line 32: you don't need these refs and the validate function.
you can use `react-hook-form` features
https://react-hook-form.com/docs/useform#mode

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -54,19 +79,61 @@ export function CreateProfile() {
   }
   return (
     <>
-      <Box margin="md">
-        <Heading variant="h5">Create wallet</Heading>
-        <Text>Enter a password to encrypt the wallet data with that</Text>
+      <Box className={authCard}>
```
lkononenko, line 82: I think it's good to separate this card together with a back button to a component as the same card is used also on other pages so it can be reused

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -54,19 +79,61 @@ export function CreateProfile() {
...
+              onValueChange={validate}
+              {...register('password')}
+              className={inputClass}
+              isRequired
+              ref={passwordRef}
+            />
+            <TextField
+              id="confirmation"
+              type="password"
+              label="Confirmation password"
```
lkononenko, line 119: The label is 'Confirm password'

packages/apps/dev-wallet/src/pages/select-profile/select-profile.css.ts
```diff
@@ -23,19 +46,19 @@ export const imgClass = style([
...
     borderRadius: 'xxl',
-    marginBlockEnd: 'md',
   }),
   {
-    width: '100px',
-    height: '100px',
-    border: `2px solid ${borderColor}`,
+    width: '40px',
+    height: '40px',
+    borderRadius: '2px',
```
lkononenko, line 53: this should be removed as there is already `borderRadius: 'xxl',` specified above and based on the design it's should be rounded

packages/apps/dev-wallet/src/pages/select-profile/select-profile.tsx
```diff
@@ -12,50 +20,60 @@ export function SelectProfile() {
   }
   return (
     <main>
-      <Box margin="md">
-        <Heading variant="h5">Select a profile</Heading>
+      <Box>
+        <Heading variant="h1" as="h1" className={titleClass}>
```
lkononenko, line 24: `variant="h1"` can be removed as already specified `as="h1"`

packages/apps/dev-wallet/src/pages/create-profile/create-profile.css.ts
```diff
@@ -0,0 +1,52 @@
...
+  }),
+  {
+    borderRadius: '1px',
+    textAlign: 'left',
+    backgroundColor: 'rgba(255, 255, 255, 0.03)',
+    border: '1px solid rgba(255, 255, 255, 0.1)',
+  },
+]);
+
+export const inputClass = style([
```
lkononenko, line 16: not sure if we should override styling of the input as if it will be re-designed/re-branded on the react-ui it will not be updated here

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -54,19 +79,61 @@ export function CreateProfile() {
...
-        <Heading variant="h5">Create wallet</Heading>
-        <Text>Enter a password to encrypt the wallet data with that</Text>
+      <Box className={authCard}>
+        <Link to="/select-profile" className={backBtnClass}>
+          <Stack alignItems="center" gap="xs" paddingBlockEnd="md">
+            <MonoArrowBackIosNew className={iconStyle} /> Back
+          </Stack>
+        </Link>
+
+        <Heading variant="h4">Create wallet</Heading>
```
lkononenko, line 89: In the design it's "Choose a password"

packages/apps/dev-wallet/src/pages/select-profile/select-profile.tsx
```diff
@@ -12,50 +20,60 @@ export function SelectProfile() {
   }
   return (
     <main>
```
lkononenko, line 22: this `main` tag was removed in my PR as there is already `main` in the layout itself

packages/apps/dev-wallet/src/App/layout-mini.css.ts
```diff
@@ -11,17 +11,28 @@ export const headerStyle = style({
...
   },
 ]);
 
 export const containerStyle = style({
-  maxWidth: '500px',
+  width: '474px',
+  maxWidth: '100%',
   margin: '0 auto',
   textAlign: 'center',
+  padding: '10px',
```
lkononenko, line 32: probably better to use value of padding from ui tokens spacing

packages/apps/dev-wallet/src/App/layout-mini.css.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const containerStyle = style({
+  width: '474px',
+  maxWidth: '100%',
+  margin: '0 auto',
+  textAlign: 'center',
+  padding: '10px',
+  '@media': {
+    'screen and (min-height: 48rem)': {
+      margin: '-100px auto 0',
```
lkononenko, line 35: Could you please explain why this negative top margin is needed? 😊

packages/apps/dev-wallet/src/pages/select-profile/select-profile.css.ts
```diff
@@ -23,19 +46,19 @@ export const imgClass = style([
...
+    backgroundColor: cardColor,
   },
 ]);
 export const aliasClass = style([
   atoms({
-    fontSize: 'lg',
-    color: 'text.brand.primary.default',
+    fontSize: 'sm',
   }),
+  { color: '#ffffff' },
```
lkononenko, line 61: the color should be used from the react-ui tokens

packages/apps/dev-wallet/src/pages/select-profile/select-profile.tsx
```diff
@@ -12,50 +20,60 @@ export function SelectProfile() {
...
-            <div className={cardClass}>
+          <Link
+            to="/create-profile"
+            style={{ textDecoration: 'none' }}
+            className={cardClass}
+          >
+            <Stack alignItems="center" gap="md">
               <div className={imgClass}>
-                <MonoAdd />
+                <MonoAdd color="#ffffff" />
```
lkononenko, line 60: the color should be used from the react-ui tokens

packages/apps/dev-wallet/src/App/layout-mini.css.ts
Likita: It better looks with this

javadkh2 (changes requested):
The `confirmation` input is not used at all! and user can submit the form without it.
it should work like this:
the submit should be `disabled` if input is empty or confirmation does not matched.
You can show an error message under the confirmation field when there is a mismatch.

javadkh2 (approved):

lkononenko (commented):
found the submit 😅 anyway I'll pick this change in my pr :)

PR #2134, opened 2024-05-15
feat: add arrows towards the trigger and add isCompact prop for a sma…

r-mulder:
In this PR we have added an arrow which points to the trigger element and add 2 size variants.
You can now provide the isCompact prop on the element to have a smaller version of the tooltip.

### DIFF
--- a/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
+++ b/packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
@@ -1,14 +1,12 @@
 import { style, styleVariants } from '@vanilla-extract/css';
 import { atoms } from '../../styles/atoms.css';
-import { tokens } from '../../styles/index';
+import { token } from '../../styles/index';
 
 export const base = style([
   atoms({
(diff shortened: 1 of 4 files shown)


packages/libs/react-ui/src/components/Tooltip/Tooltip.css.ts
```diff
@@ -41,12 +45,20 @@ export const tooltipPositionVariants = styleVariants({
...
         transform: 'translate(-50%, -100%)',
       },
+      '::after': {
+        borderBottom: `6px solid ${token('color.background.base.inverse.default')}`,
+        borderLeft: '6px solid transparent',
+        borderRight: '6px solid transparent',
+        transform: 'translateX(50%)',
+        top: '-6px',
+        right: '50%',
+      },
```
eileenmguo, line 55: if you look closely, there is actually already an indicator, it just doesn't have the right color. I would recommend updating the color of that one rather than adding an additional pseudo element 
image

eileenmguo (approved):

PR #2137, opened 2024-05-15
fix(docs): remove the theme provider

sstraatemans:
https://app.asana.com/0/1205619636589593/1207307716269071/f


packages/apps/docs/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    d.classList.remove('light');
+    d.classList.remove(darkThemeClass);
+    d.classList.add(name === 'dark' ? darkThemeClass : 'light');
+  }, []);
+
+  const setTheme = (value: ITheme): void => {
+    const resolved = lockedTheme ? lockedTheme : value;
+    setThemeState(resolved);
+    applyTheme(resolved);
```
realdreamer, line 77: i think this logic could be handled in one place, currently the same logic also part of storage listener

packages/apps/docs/src/components/Markdown/Heading/styles.css.ts
```diff
@@ -32,6 +32,7 @@ export const headerClass = style([
     marginBlockStart: 'xl',
   }),
   {
+    color: tokens.kda.foundation.color.text.base.default,
```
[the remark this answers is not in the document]
sstraatemans: because of the wonderfull way some things in docs are working we do need to overwrite.
but you are right.. could be an atom. will do

packages/libs/react-ui/src/utils/useTheme/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    // If theme is system, resolve it before setting theme
+    if (innerTheme === 'system') {
+      resolved = getSystemTheme();
+    }
+    if (lockedTheme) {
+      resolved = lockedTheme;
+    }
+
+    const name = resolved;
+    const d = document.documentElement;
```
r-mulder, line 59: Can we name this one documentElement? I like what you did the line above to rename a variable to match the context but having a single letter variable name the line below kind of undo's the good thing.

packages/libs/react-ui/src/utils/useTheme/index.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    d.classList.remove('light');
+    d.classList.remove(darkThemeClass);
+    d.classList.add(name === 'dark' ? darkThemeClass : 'light');
```
r-mulder, line 63: I was talking about this part (in slack) to use constants for the dark and light theme values. Will make it a bit more future proof if we need to change something.

ferreroltd (approved):
this is really good for dev wallet as well so cheers for this!

r-mulder (approved):

PR #2138, opened 2024-05-15
Test/account command

realdreamer:


packages/tools/kadena-cli/src/account/tests/accountDetails.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      ['0']: accountDetailsSuccessData.result.data,
+    });
+  });
+
+  it('should return error message when api fails', async () => {
+    server.use(
+      http.post(
+        'https://api.testnet.chainweb.com/chainweb/0.0/testnet04/chain/2/pact/api/v1/local',
+        () => {
+          return HttpResponse.json('shit hit the fan', { status: 500 });
```
nillo, line 173: is this a response we actually get back from chainweb and u mocked :S lol

nillo (approved):

packages/tools/kadena-cli/src/account/tests/accountDetails.test.ts
realdreamer: lol, no :D it could be any error message, so added this

barthuijgen (approved):

PR #2144, opened 2024-05-16
fix: vanilla extract issue with vite in dev mode

lkononenko:
Themes didn't work from react-ui in the dev mode built with Vite. Turned out that there is an issue with libraries built with Vanilla Extract in development mode with Vite https://github.com/vanilla-extract-css/vanilla-extract/discussions/1051

### DIFF
--- a/packages/apps/dev-wallet/package.json
+++ b/packages/apps/dev-wallet/package.json
@@ -49,6 +49,7 @@
     "@types/react-dom": "^18.2.25",
     "@typescript-eslint/eslint-plugin": "^6.0.0",
     "@typescript-eslint/parser": "^6.0.0",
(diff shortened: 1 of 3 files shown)


ferreroltd (approved):

javadkh2 (approved):

javadkh2:
if this is hard to fix; we can have two vite.config files
one for development that uses new plugin and one for production with the old config

ferreroltd:
Vite has the [conditional config](https://vitejs.dev/config/#conditional-config) but im not sure it will work. I even get the error in the editor

lkononenko:
Thanks to Sam @ferreroltd the issue with built is solved 🥳 it required upgrade of vite to 5th version

PR #2150, opened 2024-05-17
chore: added hook and removed theme provider

ferreroltd:
following pattern from docs

### DIFF
--- a/packages/apps/dev-wallet/src/hooks/useTheme.ts
+++ b/packages/apps/dev-wallet/src/hooks/useTheme.ts
@@ -1,4 +1,3 @@
-// Duplicated from docs until this gets exported from react-ui
 import { darkThemeClass } from '@kadena/react-ui/styles';
 import { useCallback, useEffect, useState } from 'react';
 const MEDIA = '(prefers-color-scheme: dark)';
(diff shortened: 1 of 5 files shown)


packages/apps/dev-wallet/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,143 @@
...
+   * It will overwrite the localstorage and even the switching of themes in your system will not update the theme
+   * IMPORTANT: using the useTheme twice in a application can lead to unexpected race conditions
+   * You would need to lock the theme on all the uses of this hook with the same ITheme
+   */
+  lockedTheme?: ITheme;
+  /**
+   * overwriteTheme can be used when you want 1 particular component to have a fixed theme
+   * IMPORTANT: this will also overwrite the lockedTheme for that component
+   */
+  overwriteTheme?: ITheme;
```
javadkh2, line 21: As far as I can see, the theme is applied at the application level. How can you set it at the component level?

ferreroltd:
This will be re-done using the hook coming from react-ui lib https://github.com/kadena-community/kadena.js/pull/2137

PR #2157, opened 2024-05-21
feat: index sales and bids

wooglie:
Updating the indexer to index sales and bids alongside events


packages/apps/marmalade-marketplace/.env.example
```diff
@@ -7,7 +7,17 @@ NEXT_PUBLIC_CHAIN_IDS="0,1"
 # Optional parameter to specify the start block for the event indexer.
 # NEXT_PUBLIC_START_BLOCK=800
 
-NEXT_PUBLIC_EVENTS="marmalade-v2.ledger.OFFER,marmalade-sale.conventional-auction.AUCTION_CREATED,marmalade-sale.dutch-auction.AUCTION_CREATED,marmalade-v2.ledger.SALE,marmalade-v2.ledger.BUY,marmalade-sale.conventional-auction.BID_PLACED,marmalade-dutch-auction.PRICE_ACCEPTED"
+NEXT_PUBLIC_EVENTS="
```
jermaine150, line 10: Do we need these events in a configuration file? It seems to me a constant list in the code would also suffice since they are hard coded anyway. We can't add a new event without writing custom code for it right?

wooglie: > We can't add a new event without writing custom code for it right?

Technically we can index a new event by just adding it to this env var, but if we want the data parsed in some way (like sale data) then we would need to update the code.

jermaine150: What would the need be for indexing and storing raw events? 

I missed the part that we're also still storing the events but since we're now storing the parsed data why do we need to store the raw values? It seems like just doubling the amount of data we store

wooglie: Yeah, you're right! We do not need to store events anymore. I removed it from the code now.

PR #2161, opened 2024-05-21
Build dev with esbuild plugin

ferreroltd:


ferreroltd:
https://github.com/kadena-community/kadena.js/pull/2144

PR #2162, opened 2024-05-21
refactor(dev-wallet): split keyset and account model

javadkh2:


packages/apps/dev-wallet/src/modules/db/db.service.tsx
```diff
@@ -49,7 +49,7 @@ const createConnectionPool = (
 };
 
 const DB_NAME = 'dev-wallet';
-const DB_VERSION = 28;
+const DB_VERSION = 31;
```
ferreroltd, line 52: in general, as long as this number is different it will rebuild right? regardless of the number

ferreroltd (approved):

packages/apps/dev-wallet/src/modules/db/db.service.tsx
javadkh2: if the version is less than the current db version then it throws an error but if its bigger than that it runs the migrate code which for now it just wipe the database ad create a fresh one (later we must address this).

PR #2182, opened 2024-05-23
Feat/kadena cli/txlocal

barthuijgen:
- add tx local command that can be used to quickly test pact code


packages/tools/kadena-cli/src/tx/commands/templates/templates.ts
```diff
@@ -50,6 +50,24 @@ networkId: "{{network:networkId}}"
...
+  code: string,
+  chainId: string,
+  network: string,
+): string => `
+code: |-
+  ${code}
+data:
+meta:
+  chainId: "${chainId}"
+  gasLimit: 2300
```
nillo, line 63: I would make the gaslimit optional also. When sending large bits of code you might have to increase it.

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+    async (option, { values }) => {
+      const { network, networkConfig } = await option.network();
+      const { chainId } = await option.chainId();
+      log.debug('tx:local', { network, networkConfig, chainId, values });
+
+      const code = values[0];
+      const templateChainId = (chainId ?? '0') as ChainId;
+      const templateNetworkId = networkConfig?.networkId ?? 'testnet04';
+      const templateNetworkHost =
+        networkConfig?.networkHost ?? 'https://api.testnet.chainweb.com';
```
nillo, line 30: Why is this here, its been a while, but if I recall correctly when none is give it automatically would fall back to this. Well at least it used to.

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+          chainId: templateChainId,
+          network: templateNetwork,
+          networkId: templateNetworkId,
+          networkHost: templateNetworkHost,
+          networkExplorerUrl: '',
+        }),
+      );
+
+      const response = await client
+        .local(transaction, {
```
nillo, line 54: I forgot the function but kadena client exposes another function, that has this automatically disabled and does the call , saves u padding 2 options.

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -99,6 +99,29 @@ export const globalOptions = {
...
+    ),
+    expand: async (network: string) => {
+      if (network === undefined) return null;
+      try {
+        return await loadNetworkConfig(network);
+      } catch (e) {
+        log.info(
+          `\nNo configuration for network "${network}" found. Please configure the network.\n`,
+        );
+        await program.parseAsync(['', '', 'networks', 'create']);
```
nillo, line 120: This should not be in here anymore

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
ash-vd: That's dirtyRead: https://github.com/kadena-community/kadena.js/blob/main/packages/libs/client/src/client/client.ts#L392

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
barthuijgen: Yeah I wasn't sure how to do this. The problem is I want the network flag to be optional and not prompted. So unless you pass `-n mainnet` as example, you won't have an `networkConfig` object. 

I can change it so it fetches one of the existing networks instead of hardcoding default values here. But then we need to discuss the logic of picking a network

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
barthuijgen: Personally I like local better, having the option there makes it clear I'm doing a local call with those settings. dirtyRead doesn't mean anything to me. But if it's against Kadena's best practices I can change it...

packages/tools/kadena-cli/src/tx/commands/templates/templates.ts
barthuijgen: is it even possible to submit a transaction without gasLimit? shouldn't I just hardcode it at a very high value then?

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
nillo: We can go with local dirtyRead in naming doesn’t say anything idd we should actually rename that function to something more obvious in the client. Let’s keep local for now for us

packages/tools/kadena-cli/src/utils/globalOptions.ts
```diff
@@ -139,6 +162,27 @@ export const globalOptions = {
       }
     },
   }),
+  chainIdOptional: createOption({
```
realdreamer, line 165: instead of creating this option completely - we can do like this right `globalOptions.chainId({ isOptional: true })`  and then adding a change in transform what you've added now `if (!chainId) return null;` it works, do we really need separate optional one?

packages/tools/kadena-cli/src/tx/commands/templates/templates.ts
nillo: optional wasn't the right word, should have been configurable. we can keep this as a default but to option to increase the value should be there

packages/tools/kadena-cli/src/tx/commands/txLocal.ts
nillo: we can go for this. it should fall back to something, wtihout going through all those hassles

packages/tools/kadena-cli/src/utils/globalOptions.ts
barthuijgen: Yes because when you add `if (!chainId) return null;` it also affects the return type. I will need to go and add guards and errors everywhere that network is used.

nillo (approved):

PR #2191, opened 2024-05-24
Feat/UI checkbox component

r-mulder:
This PR will add the new checkbox component to the library. 

The checkbox is fully based on react aria with our custom styles.


.changeset/shaggy-files-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@kadena/react-ui": patch
+---
+
+feat: new checkbox component added to the component library
```
eileenmguo, line 5: Would recommend reading the CONTRIBUTING.md file in the root of kadena.js 

Here's a snippet about changesets: 
```#### Authoring Changelogs

To keep everything clear for ourselves and our end users, we have a "changelog
etiquette". Only include changes that affect the consumer of your
package/app/product. It's not a commit log. Write descriptions that are
understandable from the consumers' perspective:

- Start descriptions with action verbs like "Add", "Remove", "Deprecate", "Fix",
  "Improve", "Update".
- Avoid the term "bug", use "issue" instead.
- Add GitHub issue numbers when fixing those.
- When referring to public API changes, use backticks and parentheses for code
  like function names and classes (e.g. `createClient()`,
  `new TransactionBuilder()`, `hash`).
- Upgrades should be documented with old and new version numbers.
- Avoid trailing periods (unless in a description below the commit title).

Examples:

- Deprecate the `doSomething()` API function.
- Use `doSomethingBetter()` instead.
- Fix issue where `ExampleWidget` API did not handle dates correctly (fix #81)
- Improve the diagnostic logging when running in advanced mode.
- Upgrade from `react@15` to `react@16-beta` release of the flexible panels
  feature

(Credits: [rushjs.io/pages/best_practices/change_logs][8])

If you don't see a need for authoring changelogs for your package (e.g. a PoC or
packages without consumers), add it to [.changeset/config.json#ignore][9].
```

packages/libs/react-ui/src/components/Form/Checkbox/Checkbox.css.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export const labelClass = style([
+  atoms({
+    display: 'flex',
+    color: 'text.base.default',
+  }),
+  {
+    alignItems: 'flex-start',
+    lineHeight: token('size.n4'),
+    cursor: 'pointer',
```
eileenmguo, line 12: can alignItems and cursor be in atoms?

packages/libs/react-ui/src/components/Form/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+  args: {
+    children: 'Check this box',
+    isReadOnly: true,
+  },
+  render: (props: ICheckboxProps) => {
+    return <Checkbox {...props}>{props.children}</Checkbox>;
+  },
+};
+
+export const _ReadOnlyChecked: CheckboxStoryType = {
```
eileenmguo, line 97: curious: what's the reason for the underscore?

packages/libs/react-ui/src/components/Link/Link.stories.tsx
```diff
@@ -68,7 +68,6 @@ type LinkStory = StoryObj<ILinkProps>;
 export const _Button: LinkStory = {
   args: {
     children: 'Hello world',
-    variant: 'primary',
```
eileenmguo, line 71: i believe the variant arg type defined in meta is incorrect. I think there should be color and style variant?

r-mulder: Link and Button have the same API, except for the default value of the variant prop. So by removing this prop you get the "transparent" variant.

packages/libs/react-ui/src/components/Form/Checkbox/Checkbox.stories.tsx
r-mulder: One of the very few things I copy pasted.. I'll see if I can remove them

packages/libs/react-ui/src/components/Form/Checkbox/Checkbox.css.ts
r-mulder: Just discussed to remove atoms from the react-ui library. So this will change but  I'll be moving the styles out the atoms function.

.changeset/shaggy-files-juggle.md
r-mulder: Clear, I'll edit this message. Thanks!

packages/libs/react-ui/src/components/Link/Link.stories.tsx
eileenmguo: ahh I see the variant includes styles and colors. thanks!

eileenmguo (approved):

PR #2192, opened 2024-05-27
fix(docs): fixing some redirects

sstraatemans:


packages/apps/docs/src/redirects/paths/blogchain.mjs
```diff
@@ -54,7 +54,7 @@ export const blogchainPathUrlRedirects = [
   },
   {
     source: '/blogchain',
-    destination: 'https://www.kadena.io/bloddg',
```
Ghislain89, line 57: Hah :D

Ghislain89 (approved):

packages/apps/docs/src/redirects/paths/blogchain.mjs
sstraatemans: 🤦🏻

PR #2193, opened 2024-05-27
Add Integration test for transaction subscriptions

Ghislain89:
### Add Integration test for transaction subscriptions

Related Issue/Asana ticket: [Asana Ticket](https://app.asana.com/0/1206316065856327/1207378878709844/f)

Short description: 
Adds an integration test for subscribing to transactions using GraphQL. This ensures that the various states a transaction can be in are correctly returned when using subscriptions.

### Test scenarios
- Create a transaction, execute this up until `preflight`  using client utils `executeTo()` function.  This yields the `requestKey`
- Set up the subscription using `graphql-ws` (tried natively with Playwright, unfortunately only HTTP requests are supported)
- Assert that the first and second response contain the expected values

### Reminders (if applicable)

- [X] I ran `pnpm install` and `pnpm test` in the root of the monorepo
      projects)
- [/] I ran `pnpm changeset` in the root of the monorepo
- [X] Test coverage has not decreased

### Questions
In order to do this test, I needed full control of the transaction and it's state. The `executeTo()` function provides this. This function results in a `task` that you can then proceed with (at a later stage) I wanted to move this logic to a separate helperfile, but I got stuck on the typings of the task. It might be a nice improvement to look at this together with one of you.


packages/e2e/e2e-graph/tests/transaction.test.ts
```diff
@@ -462,3 +482,67 @@ test.describe('Query: getTransactions', () => {
...
+      const secondEvent = (await subscription.next()).value.data;
+      expect(secondEvent).toEqual({
+        transaction: {
+          result: {
+            __typename: 'TransactionResult',
+            badResult: null,
+            goodResult: JSON.stringify('Write succeeded'),
+          },
+        },
+      });
```
nil-amrutlal-dept, line 545: Are we not playing with race conditions here? So is it guaranteed that the 2nd event will for sure be the transaction result? 
This will only work if the from the `const subscription = wsClient.iterate(query);` statement to the next `txTask.executeTo()` the time difference of when they are executed is less that the time it takes for the subscription to sent out a second event. Correct me if I am wrong @Ghislain89 , just doing a check 😄

Ghislain89: Theoretically, I think you're right. Practically speaking I believe the risk to be negligible in this case. The way I understand the subscription, is that an event is only emitted when a response (== change in data) is returned.  

By not awaiting the `txTask.executeTo()` the test just continues on, the first event that is emitted should _always_ be a Pending result, as the subscription was already active before we actually sent the tx to the chain. The second change in data should then always be a finished transaction given that we're using the instant mining sandbox container. 

If I understood correctly how a subscription works, if there is no change in response compared to the first event, no new event is emitted: meaning the second response should always be the succes message.

nil-amrutlal-dept: Hmm... I see, and I think you might be right and what we do currently is actually not correct. At the moment we return undefined on every iteration. I could change it so it only returns something if its different from the previous event

packages/e2e/e2e-graph/tests/transaction.test.ts
```diff
@@ -462,3 +479,102 @@ test.describe('Query: getTransactions', () => {
...
+          badResult: null,
+          goodResult: JSON.stringify('Write succeeded'),
+        },
+      },
+    });
+  });
+  test.afterEach(async () => {
+    wsClient.terminate();
+  });
+});
```
nil-amrutlal-dept, line 580: Niceeee

nil-amrutlal-dept (approved):
LGTM 🚀

PR #2198, opened 2024-05-28
feat(dev-wallet): Onboarding long version

Likita:
- Long version for onboarding
- Choose color and profile name


packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -107,9 +105,23 @@ export function CreateProfile() {
             />
           </Stack>
           <Stack flexDirection="column">
-            <Button type="submit" isDisabled={!isValid}>
-              Continue
-            </Button>
+            {profileList.length ? (
+              <Button isDisabled={!isValid}>
+                <Link
+                  to={'/personalize-profile'}
```
lkononenko, line 111: Shouldn't it be part of the `create` of handlesubmit function instead of the link in the button? 
Also currently the `Continue` button doesn't work on the first click while inputs of the password and confirm password are valid.

packages/apps/dev-wallet/src/pages/personalize-profile/personalize-profile.tsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const handleChooseColor = (color: string) => {
+    setActiveColor(color);
+    setLayoutContext({ accentColor: color });
+  };
+
+  return (
+    <>
+      <AuthCard>
```
lkononenko, line 81: you need to specify the url `backButtonLink="/create-profile"` as currently back button gives an error

packages/apps/dev-wallet/src/App/routes.tsx
```diff
@@ -47,6 +48,7 @@ export const Routes: FC = () => {
         <Route path="/" element={<HomePage />} />
         <Route path="/select-profile" element={<SelectProfile />} />
         <Route path="/create-profile" element={<CreateProfile />} />
+        <Route path="/personalize-profile" element={<PersonalizeProfile />} />
```
lkononenko, line 51: Should it be a separate page/url or part of the `create-profile` @javadkh2?

javadkh2: its better to to add it to the create-profile page. avoids code duplication

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
Likita: If we skip the '/personalize-profile' page, then yes, it is better to do it in the script. Will change.

I put a link before because there is a good practice to use links, not buttons, when we change the URL.

packages/apps/dev-wallet/src/pages/create-profile/create-profile.css.ts
```diff
@@ -0,0 +1,13 @@
+import { style } from '@vanilla-extract/css';
+
+export const buttonLinkClass = style([
+  {
+    textDecoration: 'none',
```
lkononenko, line 5: will be not needed in case of handling submit button as logic and not as link inside a button

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
lkononenko: Using a button as a link can cause issues for users in terms of the accessibility (by navigating with keyboard or who use a screen reader)

packages/apps/dev-wallet/src/pages/personalize-profile/personalize-profile.css.ts
```diff
@@ -0,0 +1,31 @@
...
+    borderRadius: 'xs',
+    cursor: 'pointer',
+  }),
+  {
+    width: '68px',
+    maxWidth: '100%',
+    aspectRatio: '1 / 1',
+    selectors: {
+      '&.active': {
+        border: `2px solid ${tokens.kda.foundation.color.palette.red.n50}`,
```
lkononenko, line 27: as discussed at the refinement, it should be checked with UX (Julian) as red border can look like an validation error

packages/apps/dev-wallet/src/pages/home/home-page.tsx
```diff
@@ -14,9 +16,12 @@ export function HomePage() {
     profile,
   } = useWallet();
   const { activeNetwork } = useNetwork();
+  const { setLayoutContext } = useContext(LayoutContext);
+
   if (!isUnlocked) {
     return <Navigate to="/select-profile" replace />;
   }
+  profile && setLayoutContext({ accentColor: profile?.accentColor });
```
javadkh2, line 24: this causes re-rendering! this should be wrapped within a useEffect, also its better to handled it in the layoutProvider itself then it will be applied to all pages automatically. 
So basically you need to update LayoutProvider like this
```TypeScript
const { profile } = useWallet();
const [accentColor, setAccentColor] = useState(profile?.accentColor ?? defaultAccentColor);
  
useEffect(()=>{
   setAccentColor(profile?.accentColor ?? defaultAccentColor,)
[profile])

return <LayoutContext.Provider value={accentColor}>
      {children} 
</LayoutContext.Provider>

```

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
javadkh2: This does not make the whole button a link just the children is a link; its why the button does not work as 
expected 
also we don't need a separate page, instead you can use the Wizard (stepper) component to handel this.
which lets us to navigate between steps easier since the state is all in one place
check the signature-builder page.

packages/apps/dev-wallet/src/pages/personalize-profile/personalize-profile.css.ts
Likita: Julian has updated the design. I will implement the new one

packages/apps/dev-wallet/src/modules/layout/layout.provider.tsx
```diff
@@ -0,0 +1,37 @@
+import {
+  createContext,
+  Dispatch,
+  FC,
+  PropsWithChildren,
+  SetStateAction,
+  useState,
+} from 'react';
+
+export const defaultAccentColor: string = '#42CEA4';
```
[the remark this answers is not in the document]
Likita: I didn't find this color in tokens

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
ferreroltd: @javadkh2 the stepper is not meant to allow click to navigate in the top, only next and back, but indeed each step has the continue and back

packages/apps/dev-wallet/src/pages/personalize-profile/personalize-profile.tsx
```diff
@@ -0,0 +1,127 @@
...
+    register,
+    handleSubmit,
+    formState: { isValid, errors },
+  } = useForm<{
+    profileName: string;
+  }>({ mode: 'all' });
+  const location = useLocation();
+  const { password } = location.state;
+  const { setLayoutContext } = useContext(LayoutContext);
+  const colorList = ['#CEA742', '#42CEA4', '#42BDCE', '#4269CE', '#B242CE'];
```
ferreroltd, line 32: those should be token names, i can help u figure them out

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
Likita: I changed the code.
No stepper for Personalize because it's not a step.

packages/apps/dev-wallet/src/pages/home/home-page.tsx
Likita: Updated!
I still left the function to update the color from the app because we want it to see what color we prefer before saving it to the profile.

packages/apps/dev-wallet/src/pages/personalize-profile/personalize-profile.tsx
ferreroltd: as discussed we will go like this and see if designers provide the tokens

packages/apps/dev-wallet/src/Components/PersonalizeProfile/PersonalizeProfile.tsx
```diff
@@ -0,0 +1,82 @@
+import {
```
javadkh2, line 1: lets put this file next to the page in the create-profile folder

Likita: Done

packages/apps/dev-wallet/src/pages/create-profile/create-profile.tsx
```diff
@@ -61,57 +70,64 @@ export function CreateProfile() {
       />
     );
   }
+
   return (
     <>
       <AuthCard backButtonLink="/select-profile">
```
javadkh2, line 76: I think this is too general,
when the user is in personalise profile step `back button` should go to the enter password step not the select-profile page.
I think best place for handeling this is the stepper component, so I will approve this for now

javadkh2 (approved):

PR #2205, opened 2024-05-29
Feat/UI radiogroup component

r-mulder:
Add radio and radiogroup components

Radio group has an optional prop to toggle the direction.


eileenmguo (commented):
Saw a small mistake in storybook
image

I think it would be good to add a little more detail to storybook component descriptions. Also adding links to the react-aria docs when those are applicable

eileenmguo (approved):

PR #2210, opened 2024-05-30
test: improve msw setup and create a helper utilities to create handlers

realdreamer:
- Removed global msw server handlers setup
- Created a helper util to use server handler for both simple and dynamic response use cases
- Removed unused helpers from account / account utils
- Added tests for increasing account commands and account utils test coverage 

---
  - https://app.asana.com/0/0/1207431604117120


nillo (approved):

packages/tools/kadena-cli/src/commands/account/utils/fundHelpers.ts
```diff
@@ -56,9 +56,10 @@ async function fetchTransactionDetails(
     }
     return transactionResult;
   } catch (e) {
-    throw new Error(
-      `ChainID: "${chainId}" - requestKey: ${requestKey} - ${e.message}`,
-    );
+    if (e instanceof Error) {
+      throw new Error(e.message);
```
barthuijgen, line 60: can just throw `throw e;` if it's already an error.

PR #2211, opened 2024-05-30
refactor(dev-wallet): use nested routers for steps

javadkh2:
Used nested routes for two steps in create-profile page


packages/apps/dev-wallet/src/App/layout-mini.tsx
```diff
@@ -25,7 +31,12 @@ export const LayoutMini: FC = () => {
           </Link>
         </Text>
       </Stack>
-      <main className={mainStyle}>
+      <main
+        className={mainStyle}
+        style={{
+          backgroundImage: `radial-gradient(circle farthest-side at 50% 170%, ${accentColor}, transparent 75%)`,
```
ferreroltd, line 37: this could be done this way or with VE variants, we can keep it this way for now

packages/apps/dev-wallet/src/Components/AuthCard/AuthCard.tsx
```diff
@@ -1,22 +1,28 @@
...
     <Box className={authCard}>
-      <Link to={backButtonLink} className={backButtonClass}>
-        <Stack alignItems="center" gap="xs" paddingBlockEnd="md">
-          <MonoArrowBackIosNew className={iconStyle} /> Back
-        </Stack>
-      </Link>
+      <Stack paddingBlockEnd="md">
+        <Button
+          variant="transparent"
+          onClick={() => navigate(-1)}
```
ferreroltd, line 18: this is so we navigate to the prev route? i could configure the stepper to read the route and this way we would not need to use next, it could be "smart" by checking the config provided to it and compare to the route

packages/apps/dev-wallet/src/modules/wallet/wallet.service.ts
```diff
@@ -101,7 +101,7 @@ export async function createProfile(
     name: profileName,
     networks,
     secretId,
-    accentColor: idToColor(uuid),
+    accentColor: accentColor || defaultAccentColor,
```
ferreroltd, line 104: I was thinking the other day, due to a conversation with linda, that we could always assign the color for the user

ferreroltd (approved):
looks good but the create profile is looking biGG

PR #2238, opened 2024-06-05
feat(explorer): account page

sstraatemans:
Title: Add Search Functionality to User Dashboard.
Related Issue:[ #1234](https://app.asana.com/0/1205801361945248/1207432676007835/f)
Short Description:
Create Account Detail Page


packages/apps/explorer/src/graphql/queries/account.graph.ts
alber70g, line 1: We should probably check what fields we select, and what we don't need, exclude.

I'd rather have a query per page, so we load specifically the data that's needed for that page and the features on that page.

Let's discuss in a refinement.

For now, let's do it like it is

alber70g (approved):

PR #2249, opened 2024-06-07
Migrate @kadena/docs to use Vercel CLI to deploy to preview environments.

Ghislain89:
### Deploy to Vercel using Vercel CLI instead of Vercel/Github Integration
Related Issue/Asana ticket:  https://app.asana.com/0/1206316065856327/1207502646350161/f

Short description: 
This PR adds preview deployments for @kadena/docs by using the Vercel CLI instead of relying on the Vercel/GitHub integration. This allows us to build our apps on GitHub runners and then upload prebuilt sources for deployment. For Docs this reduces deployment time from around 6-7 minutes to around 40-60 seconds. This should greatly reduce the time developers need to spent waiting for their deployment to be ready, especially when all other apps are migrated too

The deployment workflow uses `turbo-ignore` to check for differences between the current branch and the base branch to determine if a deployment for an app is required. If a deployment is required, the Vercel CLI wil pull environment variables, build the app, create a tgz archive and upload to Vercel after which a deployment will be triggered.

The deployment URL is then added to environments like so:
image

 In short, this results in the following benefits:

- Full control over Turbo Ignore and which branch we use to compare to when determining if a deployment is needed.
- Faster App deployments as we don't need to wait on Vercel to fully rebuild our apps.
- Many apps can be built in parallel, in Vercel only 2 (up to 12 if we pay more)
- Developers don't need access to Vercel to deploy, they need to be able to run Workflows. 
- Deployment failures are echo'd in GitHub Actions instead of having to go to Vercel
- 

Unfortunately, it's not all sunshine and rainbows there are also a few disadvantages. We lose branch-names for preview URLs. So instead of a branch name like `https://alpha-docs-git-feat-kadena-clipactjs-kadena-js.vercel.app/`, the generated preview url is more like `https://alpha-docs-3dn06fmjg-kadena-js.vercel.app/` 

### Tasks

- [X] Use Turbo Ignore to detect changes to base branch 
- [X] Pull environment variables from preview
- [X] Deploy prebuilt artifact to Vercel
- [X] Add Preview URL to PR
- [X] Add Metadata to deployment to easily identify all deployments associated with a PR. Use `<repository>-<pr_number>` 
- [X] Delete all preview environments associated with a PR when merging a PR
- [ ] Refactor workflow to support Matrix runs for the individual apps

### Test scenarios

* Should deploy if changes have been detected to the package
* Should deploy if dependencies have changed
* Should Add deployment URL to PR
* Should use the correct environment variables for the deployment

### Reminders (if applicable)

- [X] I ran `pnpm install` and `pnpm test` in the root of the monorepo
      projects)
- [X] I ran `pnpm changeset` in the root of the monorepo
- [X] Test coverage has not decreased
- [X] Docs have been updated to reflect changes in PR (don't forget
      docs.kadena.io)


Randynamic (approved):

Randynamic:
@alber70g your call

Ghislain89:
Based on recent features in Vercel this PR shouldn't be needed anymore. It's unlikely that the added benefit outweighs the maintenance and complexity.

PR #2256, opened 2024-06-10
Feat/kadena cli/pactjs

nillo:
Port from pactjsCLI to kadenaCLI


packages/tools/kadena-cli/src/services/pactjs/pactjs.service.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    const targetDirectory = join(
+      dirname(targetPackageJson),
+      'node_modules',
+      '.kadena/pactjs-generated',
+    );
+
+    if (args.clean === true) {
+      log.info(`Cleaning ${targetDirectory}`);
+      rimraf.sync(targetDirectory);
```
barthuijgen, line 81: should be replaced with `filesystem` service. that will also save us using another dependency :) It's just a `deleteDirectory` which it already has.

packages/tools/kadena-cli/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "outDir": "./lib",
     "lib": ["ES2022"],
     "module": "NodeNext",
-    "types": [".kadena/pactjs-generated", "./src/utils/ttys.d.ts"]
+    "types": [".kadena/pactjs-generated", "./src/utils/ttys.d.ts"],
+    "resolveJsonModule": true
```
nillo, line 8: (node:21486) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

zie import package.json in the pactjservice.

any suggestions @barthuijgen

barthuijgen: Not sure what the implications are for compatibility. To be safe I think just I'd use `fs.readFile()`.

barthuijgen (approved):

PR #2258, opened 2024-06-10
feat(explorer): search page

sstraatemans:
### Add search page results

Related Issue/Asana ticket: https://app.asana.com/0/1205801361945248/1207432694751780/f

See https://explorer.chainweb.com/mainnet/eventsearch?q=coin.TRANSFER for inspiration.

Should not navigate to a new page but replace the chain data with the event search results.


nil-amrutlal-dept:
- Noticed we only display 20 chains. This is because graph only returns 20 chains and paginates the remaining (if there are any). In case there's orphans, they wont be displayed. 
- When searching for block height there is no links in the hashes to go to the block detail page

nil-amrutlal-dept:
When searching for an account, for eg, `k:db776793be0fcf8e76c75bdb35a36e67f298111dc6145c66693b0133192e2616`, I see this:
image

packages/apps/explorer/src/hooks/search/index.ts
```diff
@@ -0,0 +1,136 @@
...
+        SearchOptionEnum.REQUESTKEY,
+      ),
+    },
+    skip:
+      !searchQuery ||
+      !isSearchRequested(searchOption, SearchOptionEnum.REQUESTKEY),
+  });
+
+  useEffect(() => {
+    console.log(1);
```
nil-amrutlal-dept, line 73: can be removed

alber70g (approved):

nil-amrutlal-dept:
I still do not see a table when searching for a transaction, eg. `http://localhost:3000/?q=WjrMBxu0n6tGHfcKZ50BI6As8_n4d8guYwwO4gbPgLw` 
One additional question is: would it not be valuable to indicate in the URL what type of search it is? So for example, if you navigate to `http://localhost:3000/?q=WjrMBxu0n6tGHfcKZ50BI6As8_n4d8guYwwO4gbPgLw` I can see that we are sending 5 requests, one for each "type" (block, block hash, block height, request key and event). Since we always need to choose an option from the dropdown we could use the URL to infer which type it is so that when we navigate to it we'd reduce the amount of requests. 
Let me know what you think :)

nil-amrutlal-dept (commented):
I still do not see a table when searching for a transaction, eg. `http://localhost:3000/?q=WjrMBxu0n6tGHfcKZ50BI6As8_n4d8guYwwO4gbPgLw`

One additional question is: would it not be valuable to indicate in the URL what type of search it is? So for example, if you navigate to `http://localhost:3000/?q=WjrMBxu0n6tGHfcKZ50BI6As8_n4d8guYwwO4gbPgLw` I can see that we are sending 5 requests, one for each "type" (block, block hash, block height, request key and event). Since we always need to choose an option from the dropdown we could use the URL to infer which type it is so that when we navigate to it we'd reduce the amount of requests.

Let me know what you think :)

nil-amrutlal-dept (approved):
Looks awesome!!  🙌🏽

PR #2277, opened 2024-06-14
chore: rename updatable to non updatable uri policy

wooglie:
rename updatable to non updatable uri policy

### DIFF
--- a/packages/libs/client-utils/src/marmalade/create-token.ts
+++ b/packages/libs/client-utils/src/marmalade/create-token.ts
@@ -130,26 +130,12 @@ const generatePolicyTransactionData = (
   }
 
   if (!policyConfig?.guarded && policyConfig?.nonUpdatableURI === false) {
-    if (props.guards.uriGuard)
-      data.push(
-        addData(
-          'uri_guard',
-          props.guards.uriGuard as unknown as ValidDataTypes,
-        ),
-      );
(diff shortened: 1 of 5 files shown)


packages/libs/client-utils/src/marmalade/create-token.ts
```diff
@@ -139,7 +139,7 @@ const generatePolicyTransactionData = (
       );
   }
 
-  if (policyConfig?.guarded && !policyConfig?.updatableURI) {
+  if (policyConfig?.guarded && policyConfig?.nonUpdatableURI) {
```
ggobugi27, line 142: With addition of nonUpdatableURI, we now don't have to enforce usage of guard policy with the registration of a failure guard token uri is non-updatable, because the existence of `non-updatable-uri-policy` will do so. 

One suggestion I'd make is enforcing uri guard registration for updatable uri's , This way we'll restrict the option of not using guard policy to handle the update-uri logic, but we be conservative and prevent an updatable uri token being created without an uri guard. Open for hearing different opinions

ggobugi27 (changes requested):

ggobugi27 (approved):

packages/libs/client-utils/src/marmalade/create-token.ts
wooglie: You're right! I have updated the code to reflect this

PR #2279, opened 2024-06-14
Fix/layers

r-mulder:
This PR will resolve the issue where styles cannot be overridden from consuming applications.

Main functionality is that we wrap the vanilla extract functions to style a component with our custom functions. These wrap the styling into a layer.
This layer will make sure that all styles applied without layer will be applied to the component itself.

### DIFF
--- a/packages/libs/react-ui/src/components/Button/Button.tsx
+++ b/packages/libs/react-ui/src/components/Button/Button.tsx
@@ -78,15 +78,16 @@ const Button = forwardRef(
     const iconOnly = Boolean(
       // check if children is a ReactElement
       (typeof children !== 'string' && typeof children !== 'number') ||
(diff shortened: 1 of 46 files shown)


packages/libs/react-ui/src/components/Button/Button.tsx
```diff
@@ -21,11 +21,12 @@ import { disableLoadingProps } from './utils';
 type Variants = NonNullable<RecipeVariants<typeof button>>;
 
 export type IButtonProps = Omit<AriaFocusRingProps, 'isTextInput'> &
-  Variants &
+  Omit<Variants, 'icCompact'> &
```
sanderlooijenga, line 24: Typo? 😁

sanderlooijenga (approved):

PR #2286, opened 2024-06-17
fix: make the icon optional for the textlink component

r-mulder:
Ticket: https://app.asana.com/0/1206752931509995/1207525414079731

The icon should be optional, this was the intended behaviour by design but implemented incorrect.
This PR will fix that, you can show the icon if you provide the withIcon prop.

### DIFF
--- a/packages/libs/react-ui/src/components/TextLink/TextLink.tsx
+++ b/packages/libs/react-ui/src/components/TextLink/TextLink.tsx
@@ -24,6 +24,7 @@ export type ITextLinkProps = Omit<AriaFocusRingProps, 'isTextInput'> &
     title?: string;
     style?: React.CSSProperties;
     children?: string | number;
+    withIcon?: boolean;
   };
 
 /**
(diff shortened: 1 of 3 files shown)


Ghislain89 (approved):

packages/libs/react-ui/src/components/TextLink/TextLink.tsx
```diff
@@ -34,11 +35,18 @@ export type ITextLinkProps = Omit<AriaFocusRingProps, 'isTextInput'> &
...
  */
 
 const TextLink = forwardRef(
   (
-    { isCompact = false, children, className, ...props }: ITextLinkProps,
+    {
+      isCompact = false,
+      children,
+      className,
+      withIcon,
```
sanderlooijenga, line 47: Should this, for backwards compatibility reasons, default to true? 🤷‍♂️

sanderlooijenga (approved):

packages/libs/react-ui/src/components/TextLink/TextLink.tsx
r-mulder: Technically yes, but as the component is still very new we were bold and changed it to how UX initially wanted it to work.

sanderlooijenga: So bold 💪 😬

Ghislain89 (approved):

PR #2332, opened 2024-06-27
ag/add graphql info

alber70g:
- chore(explorer): create titles for every page. Improve truncateValues function and added tests
- chore(explorer): add more info to the graphql modal

---
  - https://app.asana.com/0/0/1207667342191156


packages/apps/explorer/src/components/graphql-query-dialog/graphql-query-dialog.tsx
```diff
@@ -28,8 +28,47 @@ export const GraphQLQueryDialog = (): JSX.Element => {
       >
         {() => (
           <DialogContent>
-            <p>Amount of queries on this page: {queries.length}</p>
+            <h2>
```
sstraatemans, line 31: we should use the reactui components
Heading, Text TextLink etc

sstraatemans (approved):

PR #2339, opened 2024-06-28
chore: marmalade marketplace sales/auctions

wooglie:
Marmalade Sale/Auction WIP


packages/apps/marmalade-marketplace/src/components/Sale/DutchAuction.tsx
```diff
@@ -0,0 +1,286 @@
...
+  const handleBuyNow = async () => {
+    if (!auctionDetails) return;
+
+    if (!account?.alias) {
+      alert("Please connect your wallet first to buy.");
+      return;
+    }
+
+
+    const FUNDING_ACCOUNT = "sender00";
```
jermaine150, line 146: Do we need / want this private / public key info in the code?

packages/apps/marmalade-marketplace/src/components/Sale/RegularSale.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export function RegularSale({ tokenImageUrl, sale }: RegularSaleProps) {
+
+  const router = useRouter() as AppRouterInstance;
+  const searchParams = useSearchParams();
+  const { account } = useAccount();
+
+  useEffect(() => {
+
+    const transaction = searchParams.get("transaction");
```
jermaine150, line 28: In what case do we get here with a transaction in the queryparams?

ggobugi27: When the signed transaction from the spirekey is returned the page is directed to the `/transaction`page.

packages/apps/marmalade-marketplace/src/components/Sale/DutchAuction.tsx
ggobugi27: Sender00 is a keyset in a public repo (https://github.com/kadena-io/chainweb-node/blob/f258a705e77ce28c08cf934980c7d12cb4f5be30/pact/genesis/devnet/keys.yaml#L4) and is for the devnet admin key so I think this shouldn't be a problem. We'll need to change that to the buyer key in the future.

packages/apps/marmalade-marketplace/src/components/Sale/ConventionalAuction.tsx
```diff
@@ -0,0 +1,250 @@
...
+        alt="Token Image"
+        className={styles.tokenImageClass}
+      />
+      <div className={styles.tokenInfoClass}>
+        {!isActive && <>
+          <span>Auction has ended</span>
+          <hr />
+        </>}
+
+        {mustCreateAuction && <>
```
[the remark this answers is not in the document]
ggobugi27: I think this is specifically for sales that are made with `conventional-auction` as sale-type but `create-auction` is yet to be completed. We can change the message to , `auction must be initiated`?

packages/apps/marmalade-marketplace/src/components/Sale/RegularSale.tsx
jermaine150: As discussed it would be better if spirekey was redirected to the transaction page directly, which would remove the need for this extra redirect.
However since this call is being made from client-utils we need to update that first with the ability to specify the return url.

packages/apps/marmalade-marketplace/src/components/Sale/ConventionalAuction.tsx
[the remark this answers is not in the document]
ggobugi27: Good call. I think this component assumes sale creation with the conventional auction sale-type but that's true this page can be rendered with different sale-id's.

jermaine150 (approved):

PR #2356, opened 2024-07-02
ag/explorer add custom network

alber70g:
- chore(explorer): minor fixes and styling improvements
- WIP: custom networks


packages/apps/explorer/src/components/select-network/new-network.tsx
```diff
@@ -0,0 +1,154 @@
...
+            networkInfo {
+              totalDifficulty
+            }
+          }`,
+          variables: {},
+          operationName: 'networkInfo',
+          extensions: {},
+        }),
+      });
+      setCheckStatus(result.status);
```
alber70g, line 81: Maybe check also that networkId is same as the one that's passed by the user

packages/apps/explorer/src/components/select-network/new-network.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+      console.log(result);
```
alber70g, line 83:
```suggestion
```

packages/apps/explorer/src/components/select-network/new-network.tsx
```diff
@@ -0,0 +1,154 @@
...
+      await result.json();
+
+      if (result.status === 200) {
+        addNetwork(newNetwork);
+        createNetwork(e);
+      }
+    } catch (e) {
+      console.log(222, e);
```
alber70g, line 91:
```suggestion
```

packages/apps/explorer/src/components/select-network/styles.css.ts
```diff
@@ -0,0 +1,7 @@
+import { globalStyle, style } from '@vanilla-extract/css';
+
+export const selectWrapperClass = style({});
+
+globalStyle(`${selectWrapperClass} ~ ul`, {
+  background: 'green!important',
```
alber70g, line 6: 😳

packages/apps/explorer/src/components/select-network/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+export const validateNewNetwork = (
+  networks: INetwork[],
+  newNetwork: INetwork,
+): string[] => {
+  const errors = [];
+
+  if (networks.some((network) => network.networkId === newNetwork.networkId)) {
+    errors.push('network already exists');
+  }
+
```
alber70g, line 12: I think I made this. Now that I think about it, we should probably check uniqueness on the combination of networkId and label. NetworkId can be used more than once, when you have multiple devnnets it'll be `development` for both.

packages/apps/explorer/src/hooks/network/redirect.ts
```diff
@@ -1,20 +0,0 @@
...
-      if (mainnetUrl) {
-        // window.location.href = mainnetUrl;
-      }
-    } else if (selectedNetwork === networkConstants.testnet04.key) {
-      if (testnetUrl) {
-        // window.location.href = testnetUrl;
-      }
-    }
-  }, [selectedNetwork]);
-};
```
alber70g, line 20: For switching between mainnet and testnet, we might want to still have this, as we don't have a `<host>/testnet` yet.

packages/apps/explorer/src/hooks/search/index.ts
```diff
@@ -77,14 +75,15 @@ export const useSearch = () => {
...
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        router.replace(`${router.asPath}`);
       }
       return;
     }
 
-    const q = params.get('q');
-    const so = params.get('so');
+    const q = router.query.q;
+    const so = router.query.so;
```
alber70g, line 86:
```suggestion
    const query = router.query.q;
    const searchOption = router.query.so;
```

packages/apps/explorer/src/components/select-network/utils.ts
sstraatemans: makes sense

packages/apps/explorer/src/components/select-network/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+export const getFormValues = <T extends Record<string, unknown>>(
+  data: FormData,
+): T => {
+  const values: Record<string, unknown> = {};
+  data.forEach((value, key) => {
+    values[key] = value;
+  });
+  return values as T;
+};
```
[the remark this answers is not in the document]
sstraatemans: should be fine

sstraatemans (approved):

PR #2371, opened 2024-07-08
fix(explorer): sticky header needs more space

alber70g:


packages/apps/explorer/src/components/layout/styles.css.ts
```diff
@@ -16,5 +16,6 @@ export const layoutWrapperClass = style({
 });
 
 export const contentClass = style({
+  height: '64px',
```
sstraatemans, line 19: need to use a token for this

packages/apps/explorer/src/components/layout/layout.tsx
```diff
@@ -14,7 +14,6 @@ export const Layout: FC<IProps> = ({ children }: IProps) => {
   return (
     <div className={documentStyle}>
       <Header />
-      <Stack marginBlock="xl" />
       <Stack
```
sstraatemans, line 17: you can add the marginblockstart to the below stack

sstraatemans (approved):

packages/apps/explorer/src/components/layout/layout.tsx
alber70g: Isa mentioned that the header height isn't tokenized...

packages/apps/explorer/src/components/layout/styles.css.ts
alber70g: Isa mentioned that the header height isn't tokenized...

PR #2373, opened 2024-07-09
feat(proof): connect to spirekey sdk

Ghislain89:
https://app.asana.com/0/1206452987729603/1207666615570583/f


Randynamic:
@EnoF Can we merge this?

packages/e2e/e2e-proof-of-us/tests/test-1.spec.ts
```diff
@@ -1,33 +1,46 @@
...
 
+const TESTURL = 'http://localhost:3000/';
+
 test('1 Initiator, 1 signers. all participants sign -> Should be able to mint the connection token @xs', async ({
   initiator,
   signer1,
 }) => {
+  const webAuthNHelper = new WebAuthNHelper();
+  await webAuthNHelper.enableWebAuthN(initiator);
+  await webAuthNHelper.enableWebAuthN(signer1);
```
Ghislain89, line 20: The Virtual Authenticators are enabled for a specific session, in this specific test you're essentially creating multiple  virtual authenticators for various domains. On line 19 and 20, you're actually creating the virtual authenticators, but have not yet opened the proof-of-us app. 

In the functions `createSpireKeyAccountFor` and `signTransaction` your **recreating** the virtual authenticator, which will likely result in being unable to sign, it would be a considerably security problem if you could just randomly create a virtual authenticator and sign with it.

What I would do to streamline this:

* Navigate to the `TESTURL` with both the initiator and signer.
* Enable the virtual authenticator by calling `enableWebAuthN` for both the initiator and signer
* With both the Initiator and the Signer, create a Spirekey Account
* With the Initiator, create the proof
* With the signer, open the shareURL
* Start the signing process with the initiator
* Sign with the signer
* Sign with initiator
* Epic win

Note: all the promise.all's are only needed if you want to scale up the amount of particpants, for the purpose of these tests (functional validation) I would skip that for now. If you get really stuck, I can create a fork on monday and help a friend out.

sstraatemans: the issue is that the popups for the new inline SDK need auhtenticators as well.

Randynamic (approved):

PR #2376, opened 2024-07-09
chore(graph): add sentry to @kadena/graph

alber70g:
This PR adds Sentry to Graph.

**Testnet configuration:**
SENTRY_DSN: Can be found [here](https://kadena-test.sentry.io/settings/projects/kadena-graph/keys/)
ENVIRONMENT: `testnet`

**Mainnet configuration:**
SENTRY_DSN: Can be found [here](https://kadena-test.sentry.io/settings/projects/kadena-graph/keys/)
ENVIRONMENT: `mainnet`

The default value for ENVIRONMENT is `mainnet`.  There is no default value for `SENTRY_DSN`.  If no value is set, Sentry will not be enabled and the environment will not be used.


sstraatemans (approved):

packages/apps/graph/src/index.ts
```diff
@@ -30,12 +33,43 @@ if (dotenv.NODE_ENV === 'development') {
...
+};
+
+if (dotenv.SENTRY_DSN) {
+  console.log(
+    ` ✔ starting with sentry ${dotenv.NODE_ENV}-${dotenv.ENVIRONMENT}`,
+  );
+  Sentry.init({
+    dsn: dotenv.SENTRY_DSN,
+    tracesSampleRate: 1,
+    environment: dotenv.ENVIRONMENT,
```
alber70g, line 68: Should this be `${dotenv.NODE_ENV}-${dotenv.ENVIRONMENT}`?

```suggestion
    environment: `${dotenv.NODE_ENV}-${dotenv.ENVIRONMENT}`,
```

I think just mainnet/testnet isn't enough to differentiate. Especially when we run it in development too.

Would be nice to query the configured chainwebapi (https://api.chainweb.com/info) and use the `nodeversion`

Ghislain89: I thought about this. The default value for NODE_ENV is `production`. If a developer incorrectly sets environment variables (or just forgets them), you could end up with `production-local` which sounds confusing.

Having said that, the default for ENVIRONMENT is `mainnet`. If a developer sets the SENTRY_SDN but forgets the environment, they could pollute the logging...

PR #2391, opened 2024-07-11
feat(marmalade-marketplace): Add styled pages, navbar, create-token, create-sale

ggobugi27:


packages/apps/marmalade-marketplace/public/connect_icon.svg
jermaine150, line 1: We can use the icon from kode-icons instead of adding the image ourselves. But we can do that in the next iteration when working on the theming.

jermaine150 (approved):

isaozler (approved):

PR #2401, opened 2024-07-12
feat(explorer): add funcitonality to remove networks

sstraatemans:
https://app.asana.com/0/1205801361945248/1207765395158110/f


Randynamic (approved):
I like this a lot I only want to add a bit of margin on the right side to make the text more readable 
Screenshot 2024-07-22 at 12 00 59

isaozler:
> I like this a lot I only want to add a bit of margin on the right side to make the text more readable Screenshot 2024-07-22 at 12 00 59

I will share the new design for this

packages/apps/explorer/src/components/select-network/ConfigNetwork.tsx
```diff
@@ -0,0 +1,279 @@
...
+                    }}
+                    variant="outlined"
+                  >
+                    Cancel
+                  </Button>
+                  <Button type="submit">
+                    {!network.isNew ? 'Update network' : 'Create Network'}
+                  </Button>
+                </>
+              )}
```
eileenmguo, line 272: Are you trying to have buttons with space between? If so, there should be a prop on the CardFooter that allows for this (separated)

PR #2411, opened 2024-07-16
chore(explorer): replace networkid with slug for routing

sstraatemans:
https://app.asana.com/0/1205801361945248/1207795059899493/f


packages/apps/explorer/src/constants/network.ts
```diff
@@ -1,15 +1,17 @@
...
     explorerUrl: 'https://explorer.kadena.io/',
   },
-  testnet04: {
+  testnet: {
     networkId: 'testnet04',
     label: 'Testnet',
+    slug: 'testnet',
     chainwebUrl: 'api.testnet.chainweb.com',
     graphUrl: 'https://graph.testnet.kadena.network/graphql',
     wsGraphUrl: 'https://graph.testnet.kadena.network/graphql',
```
alber70g, line 17: We don't do anything specific that we need to use a Record<string,string> so you can also convert it to an array with `type Array<INetwork>`

packages/apps/explorer/src/context/networks-context.tsx
```diff
@@ -109,34 +119,39 @@ const NetworkContextProvider = (props: {
...
-    const network = networks.find((x) => x.networkId === networkId)!;
-    setActiveNetwork(network);
+  const setActiveNetworkByKey = (networkSlug: string): void => {
+    const network = networks.find((x) => x.slug === networkSlug);
+
+    if (!network) {
+      addToast({
+        type: 'negative',
+        label: 'Network not found',
+        body: `network ${networkSlug} is depricated`,
```
alber70g, line 129: Should we even add a toast when the network isn't found? Wouldn't it be better to just forward to `/mainnet` and ignore the wrong network.

Also, what does it mean tha ta network is deprecated, does it work now, and later not anymore?


Typo
```suggestion
        body: `network ${networkSlug} is deprecated`,
```

alber70g (approved):

packages/apps/explorer/src/context/networks-context.tsx
sstraatemans: my english is not so good :)

PR #2435, opened 2024-07-23
fix(explorer): editting networks

sstraatemans:


packages/apps/explorer/src/context/networksContext.tsx
```diff
@@ -244,6 +244,13 @@ const NetworkContextProvider = (props: {
       Cookies.set(selectedNetworkKey, newNetwork.slug);
 
       window.location.href = `/${newNetwork.slug}`;
+    } else {
+      const newStorage = storage.map((network) => {
+        if (network.slug === newNetwork.slug) return newNetwork;
+        return network;
+      });
+      localStorage.setItem(storageKey, JSON.stringify(newStorage));
+      window.dispatchEvent(new Event(storageKey));
```
Randynamic, line 253: What is the reason for the dispatch?

Randynamic (approved):

packages/apps/explorer/src/context/networksContext.tsx
sstraatemans: when you change something in localstorage. you can listen to the event in other tabs. so they can change their state accordingly.
this works for other tabs. but not for the CURRENT tab.
if you want to take action in the current tab when storage changes, you need to dispatch a custom event yourself

maybe not needed at the moment. but I want the app to be concience that localstorage is changed in ALL tabs

PR #2446, opened 2024-07-29
feat(explorer): search page results per type

sstraatemans:
---
  - https://app.asana.com/0/0/1207919421324419


packages/apps/explorer/src/context/searchContext/utils/utils.ts
```diff
@@ -0,0 +1,24 @@
+export enum SearchOptionEnum {
+  ACCOUNT,
+  REQUESTKEY,
+  BLOCKHASH,
+  BLOCKHEIGHT,
+  EVENT,
+}
```
alber70g, line 7: Use object `as const` instead

alber70g (approved):

packages/apps/explorer/src/context/searchContext/utils/utils.ts
sstraatemans: done

PR #2447, opened 2024-07-30
Improve sale pages

ggobugi27:
- Add token info page that supports sale/buy.


packages/apps/marmalade-marketplace/src/components/CreateToken/index.tsx
```diff
@@ -96,7 +96,7 @@ function CreateTokenComponent() {
 
   const onTransactionSigned = (transaction: IUnsignedCommand | ICommand) => {
     setTransaction(transaction);
-    router.push(`/transaction?returnUrl=/mint?tokenId=${tokenId}`);
+    router.push(`/transaction`);
```
jermaine150, line 99: Was this removed by accident?

ggobugi27: Yes, just pushed fix

packages/apps/marmalade-marketplace/src/pages/tokens/[tokenId]/styles.css.ts
```diff
@@ -1,6 +1,6 @@
 import { style } from '@vanilla-extract/css';
 import { tokens } from '@kadena/kode-ui/styles';
-import { deviceColors } from './tokens.css';
+import { deviceColors } from '../../../styles/tokens.css';
```
Comment (jermaine150) [inline packages/apps/marmalade-marketplace/src/pages/tokens/[tokenId]/styles.css.ts:3]: We can get the borderColor from tokens as well, that way they will change with the theming

Comment (jermaine150) [reply packages/apps/marmalade-marketplace/src/pages/tokens/[tokenId]/styles.css.ts]: I updated this in another branch.

jermaine150 (approved):

PR #2448, opened 2024-07-31
Update with r:account, add listing page

ggobugi27:
- Update `keyset` inside client-util/marmalade functions to `guard`
- add support for `keyset-ref-guard` guard type in `client-utils/marmalade`'s `create`, `mint`, `offer`, `buy`. 
- Add listing page styles


packages/apps/marmalade-marketplace/src/components/Badge/style.css.ts
```diff
@@ -0,0 +1,11 @@
+import { token } from '@kadena/kode-ui/styles';
+import { style } from '@vanilla-extract/css';
+
+export const tokenBadgeWrapperClass = style({
+  position: 'absolute',
+  top: '8px',
```
EnoF, line 6: I think these should be styled with `token('spacing.n2')` not sure if that's the correct token though

packages/apps/marmalade-marketplace/src/components/Badge/style.css.ts
```diff
@@ -0,0 +1,11 @@
...
+  right: '8px',
+  backgroundColor: token('color.background.accent.primary.@hover'),
+  padding: '4px 10px',
+  borderRadius: '16px',
```
EnoF, line 10: the border radius has a specific token too I believe

packages/apps/marmalade-marketplace/src/components/CreateToken/index.tsx
```diff
@@ -59,40 +58,39 @@ function CreateTokenComponent() {
   });
 
   const [guardInput, setGuardInput] = useState({
```
EnoF, line 60: wouldn't it be better to have this state remain as `undefined` till the `useEffect` take place?

packages/apps/marmalade-marketplace/src/components/CreateToken/index.tsx
```diff
@@ -59,40 +58,39 @@ function CreateTokenComponent() {
...
+    royaltyCreator: account?.accountName ?? '',
+    royaltyGuard: account?.guard,
     royaltyRate: '0.05',
   });
 
   useEffect(() => {
-    if (webauthnAccount) {
-      const key = webauthnAccount.guard.keys[0];
-
+    if (account) {
```
EnoF, line 76: I usually invert the if statements for early returns, so in this case,
```
if (!account) return

const.....
```

allows for less context that needs to be kept in mind while reading the code

packages/apps/marmalade-marketplace/src/components/GuardForm/index.tsx
```diff
@@ -2,9 +2,8 @@ import React, { FC } from 'react';
 import { TextField, Checkbox } from '@kadena/kode-ui';
 import CrudCard from '@/components/CrudCard';
 
-
 interface GuardFormProps {
-  guardInput: { [key: string]: string };
+  guardInput: { [key: string]: any };
```
EnoF, line 6: The sdk has exposed a `Guard` type you could use here.

EnoF: nvm, I thought it was a guard type

packages/apps/marmalade-marketplace/src/components/MintToken/index.tsx
```diff
@@ -20,7 +21,8 @@ import CrudCard from '@/components/CrudCard';
 
 function MintTokenComponent() {
   const router = useRouter();
-  const { account, webauthnAccount } = useAccount();
+  const { account } = useAccount();
+  console.log("Account from provider", account)
```
EnoF, line 25: left over console log

packages/apps/marmalade-marketplace/src/components/MintToken/index.tsx
```diff
@@ -34,7 +36,7 @@ function MintTokenComponent() {
   
   const onTransactionSigned = (transaction: IUnsignedCommand | ICommand) => {
     setTransaction(transaction);
-    router.push(`/transaction?returnUrl=/tokens/${tokenId}`);
+    router.push(`/transaction?returnUrl=/tokens/${tokenId}?chainId=${`8`}`);
```
EnoF, line 39: is the `8` a preparation for it to be parameterized?

packages/apps/marmalade-marketplace/src/components/GuardForm/index.tsx
ggobugi27: I think `Guard` is declared locally and not exported

packages/apps/marmalade-marketplace/src/components/MintToken/index.tsx
ggobugi27: Yes, user will be selecting the chainId, but we're setting it to 8 for now

packages/apps/marmalade-marketplace/src/components/Sale/ConventionalAuction.tsx
```diff
@@ -142,7 +142,7 @@ export function ConventionalAuction({ tokenImageUrl, sale }: ConventionalAuction
 
     try {
       await buyToken({
-        signer: "",
+        signerPublicKey: "",
```
EnoF, line 145: I'm a bit surprised by the empty string here, why is the signerPubKey empty?

packages/apps/marmalade-marketplace/src/pages/mytokens/index.tsx
```diff
@@ -7,24 +7,34 @@ import { Grid, GridItem, Stack, Heading } from "@kadena/kode-ui";
...
   useEffect(() => {
-    fetchTokens(webauthnAccount?.account);
-  }, [webauthnAccount?.account]);
+    fetchTokens(account?.accountName);    
+  }, [account?.accountName]);
 
   return (
     <Stack flex={1} flexDirection="column">
-      <Heading as="h1">My Tokens</Heading>
+      <div style={{ marginTop: '-60px', marginLeft: '-6%', marginRight: '-6%'}}>
```
EnoF, line 26: At somepoint you'd probably want to refactor this to use `Stack` and `Image`. Specially Image is very useful as it provides optimizations for faster initial loads.

packages/apps/marmalade-marketplace/src/components/CreateToken/index.tsx
ggobugi27: I'd vote to iterate that on the next round, since we'd like to deploy the new changes internally.

packages/apps/marmalade-marketplace/src/providers/AccountProvider/AccountProvider.tsx
```diff
@@ -20,35 +18,34 @@ export interface WebauthnAccountDetails {
...
   const chainId = '8';
 
+  useEffect(() => {
+    initSpireKey({ hostUrl: env.WALLET_URL, useRAccount: true })
+  }, []);
+
   const login = useCallback(async () => {
     const account = await connect(env.NETWORKID, chainId);
     console.log(account)
+    console.log(account)
```
EnoF, line 48: more logs :stuck_out_tongue_closed_eyes:

packages/apps/marmalade-marketplace/src/styles/create-token.css.ts
```diff
@@ -194,7 +193,7 @@ export const container = style({
   margin: '160px 15% 0',
   backgroundColor: tokens.kda.foundation.color.background.layer.default,
   borderRadius: tokens.kda.foundation.radius.sm,
-  border: `1px solid ${deviceColors.borderColor}`,
+  border: `1px solid ${tokens.kda.foundation.color.border.base.subtle}`,
```
EnoF, line 196: I think you should be able to replace the lengthy path with:

```
token('color.border.base.subtle')
```

packages/apps/marmalade-marketplace/src/utils/helper.ts
```diff
@@ -120,5 +113,5 @@ export const getPolicies = (policyConfig: ICreateTokenPolicyConfig) => {
 
 export const isPrecise = (n:number, precision:number):boolean => {
   const length = String(n).split(".")[1] ? String(n).split(".")[1].length : 0;
```
EnoF, line 115: You could opt to use a regex:

```
new RegExp(`\.[\d]{0,${precision + 1}}`).test(n.toString())
```

EnoF: or maybe even do something like:

```
n.toString() === n.toPrecision(precision)
```

packages/libs/client-utils/src/marmalade/helpers.ts
```diff
@@ -89,3 +95,16 @@ export const formatWebAuthnSigner = (
 
   return formatSingleSigner(signer);
 };
+
+export const isKeysetGuard = (guard: Guard): guard is Keyset => {
```
EnoF, line 99: For the sake of it, I'd be more specific with this check, just to avoid any unwanted behaviors.

```
Object.prototype.keys = 'hello'
isKeysetGuard({}) // true
```
So either add `hasOwnProperty` or `Array.isArray(guard.keys)` would prevent this from occuring

EnoF (approved):

PR #2465, opened 2024-08-19
feat(explorer): accountpage

sstraatemans:
---
  - https://app.asana.com/0/0/1207841752442712


alber70g (changes requested):
I've added some comments to the individual functions, but they can be ignored if this can be done cleaner from the start. 

I think the collection of functions divideChains, fillChains, chainBalancePercentage is a bit cumbersome.

My proposal would be to have a `processChainAccounts` function that will do everything in one place:
- convert the chainAccounts to objects
- fill it up to the amount of chains
- set '-' for chains that have no account
- set the percentage of the individual chain compared to the highest account balance with a minimal threshold

It becomes one array of something like `ProcessedChainAccount[]` or `ViewChainAccount` where 

```
type ProcessedChainAccount: { 
  chainId: string,
  percentage: number, 
  balance: number | '-',
}
```

What do you think?

packages/apps/explorer/src/components/AccountBalanceDistribution/AccountBalanceDistribution.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type { IChainAccounts } from './components/ChainList';
+import { ChainList } from './components/ChainList';
+
+interface IProps extends PropsWithChildren {
+  accountName: string;
+  chains: IChainAccounts;
+}
+export const AccountBalanceDistribution: FC<IProps> = ({ chains = [] }) => {
+  const { chains1, chains2, maxValue } = useMemo(() => {
+    return devideChains(fillChains(chains, CHAINCOUNT));
```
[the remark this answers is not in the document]
sstraatemans: doh

packages/apps/explorer/src/components/AccountBalanceDistribution/components/ChainBalance.tsx
```diff
@@ -0,0 +1,60 @@
...
+  chainBalanceWrapperClass,
+  chainTextBaseClass,
+  chainTextDisabledClass,
+  chainTextLargeClass,
+  chainTextSubtleClass,
+  percentageValueVar,
+} from './style.css';
+
+interface IProps extends PropsWithChildren {
+  chain: IViewChain;
```
alber70g, line 17: chainAccount?

packages/apps/explorer/src/components/LoadingSkeleton/loadingData/loadingDataAccountquery.ts
```diff
@@ -4,7 +4,7 @@ export const loadingData: AccountQuery = {
   fungibleAccount: {
     accountName: '0',
     totalBalance: 0,
-    chainAccounts: Array(5).fill({
+    chainAccounts: Array(20).fill({
```
alber70g, line 7: Use the const for chain count?

packages/apps/explorer/src/utils/processChainAccounts.ts
```diff
@@ -0,0 +1,78 @@
...
+  const maxChainBalance = calculateMaxChainBalance(chains);
+
+  return Array.from(Array(chainCount).keys()).map((idx) => {
+    const chain = chains.find((c) => {
+      return typeof c !== 'string' && c?.chainId === `${idx}`;
+    });
+
+    if (!chain || typeof chain === 'string') {
+      return {
+        chainId: `${idx}`,
```
alber70g, line 45: Would you want to set balance to - and percentage to 0 if there's no account?
Now this has to be determined downstream when you can apply the logic here in one place

alber70g (approved):
Minor comments. LGTM

packages/apps/explorer/src/utils/processChainAccounts.ts
sstraatemans: you want to have the difference between a chain where the user is active and has a balance of 0. and a chain where he is not active: balance: undefined

packages/apps/explorer/src/components/AccountBalanceDistribution/components/ChainBalance.tsx
sstraatemans: ?? what do you mean?

alber70g: Instead of `chain` rename the prop name to `chainAccount`

PR #2469, opened 2024-08-21
feat(explorer): able to finish crosschain transfer

sstraatemans:
---
  - https://app.asana.com/0/0/1205842872749419


alber70g (approved):
LGTM (only code review, not tested on preview)

PR #2472, opened 2024-08-22
ag/graph performance transaction pactid

alber70g:
- **fix(graph): get counterparts only when sender or receiver is empty**
- **feat(graph): add cw-d extra migration for index on transactions.sender transfers.to_account and transfers.from_account using height DESC**

---
  - https://app.asana.com/0/0/1207813394370745

### DIFF
--- a/packages/apps/graph/.env.example
+++ b/packages/apps/graph/.env.example
@@ -4,15 +4,24 @@ COMPLEXITY_ENABLED=false
 COMPLEXITY_EXPOSED=false
 
 # Database
-DATABASE_URL="postgresql://devnet@localhost:5432/devnet?pool_timeout=0"
 PRISMA_LOGGING_ENABLED=false
 PRISMA_LOG_TO_FILE=false
 PRISMA_LOG_FILENAME="prisma.log"
 
+# Database local
+DATABASE_URL="postgresql://devnet@localhost:5432/devnet?pool_timeout=0"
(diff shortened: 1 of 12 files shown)


packages/apps/graph/.env.example
```diff
@@ -4,16 +4,25 @@ COMPLEXITY_ENABLED=false
 COMPLEXITY_EXPOSED=false
 
 # Database
-DATABASE_URL="postgresql://devnet@localhost:5432/devnet?pool_timeout=0"
 PRISMA_LOGGING_ENABLED=false
 PRISMA_LOG_TO_FILE=false
 PRISMA_LOG_FILENAME="prisma.log"
 
+# Database local
+DATABASE_URL="postgresql://devnet@localhost:5432/devnet?pool_timeout=0"
```
sstraatemans, line 12: 👍🏻

packages/apps/graph/src/graph/objects/fungible-account.ts
```diff
@@ -163,62 +163,37 @@ export default builder.node(
...
+                moduleName: parent.fungibleName,
+              },
+            });
           } catch (error) {
             throw normalizeError(error);
           }
         },
 
-        async resolve(condition, parent) {
+        async resolve(condition, parent, args, context, info) {
```
sstraatemans, line 182: these are not used?

sstraatemans (approved):

PR #2498, opened 2024-08-30
feat(tools): send message on slack

sstraatemans:


packages/apps/tools/src/scripts/faucetCron/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      extensions: {},
+    }),
+  });
+
+  const data = (await result.json()) as IAccount;
+  return data;
+};
+
+export const runJob = async () => {
+  const accountResult = await getFaucetAccount();
```
alber70g, line 57: Can we wrap this in a try-catch and send an error message to slack when something fails? Maybe the graph doesn't reply, or some other unforseen error occurs

alber70g (approved):
Minor comment. LGTM

PR #2506, opened 2024-09-05
feat(graph): expose other types of guard aside from KeysetGuard only

alber70g:
---
  - https://app.asana.com/0/0/1208225160131340
  - https://app.asana.com/0/0/1208188219888969


packages/apps/graph/src/graph/objects/block.ts
```diff
@@ -8,7 +8,7 @@ import {
 import { normalizeError } from '@utils/errors';
 import { networkData } from '@utils/network';
 import { builder } from '../builder';
-import type { IGuard } from '../types/graphql-types';
+import type { Guard } from '../types/graphql-types';
```
sstraatemans, line 11: the only type that is not prefixed with I. Shouldnt this be IGuard, for consistency

packages/apps/graph/src/graph/objects/non-fungible-token-balance.ts
```diff
@@ -44,9 +44,7 @@ export default builder.node(
       version: t.exposeString('version'),
       guard: t.field({
         type: 'Guard',
-        resolve(parent) {
-          return parent.guard;
-        },
+        resolve: (parent) => parent.guard,
```
sstraatemans, line 47: awesome

sstraatemans (changes requested):
small comment about the IGuard

sstraatemans (approved):

PR #2511, opened 2024-09-10
fix(graph): do not throw when no crosschain transfer is found

alber70g:


packages/apps/graph/src/graph/objects/transfer.ts
```diff
@@ -86,9 +86,12 @@ export default builder.prismaNode(Prisma.ModelName.Transfer, {
           }
 
           if (parent.senderAccount === '') {
+            console.log('parent.transaction', parent.transaction);
```
sstraatemans, line 89: console.log

sstraatemans (approved):

PR #2522, opened 2024-09-16
fix(graph): use empty fields for keys: [] and predicate: '' instead of null

alber70g:


alber70g:
This is the first commit of wallet-sdk. Not sure why it ended up in the graph branch

PR #2527, opened 2024-09-18
fix(ui): fix padding of SelectBox Component

sstraatemans:


packages/libs/kode-ui/src/components/Form/Select/Select.css.ts
```diff
@@ -13,24 +14,28 @@ export const selectButtonValue = style({
...
+    paddingBlock: 'no',
+    flex: 1,
+    overflow: 'hidden',
+    position: 'relative',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+  }),
+  {
+    display: 'flex',
+    backgroundColor: token('color.background.input.default'),
```
EnoF, line 28: if the background color would be defined in `atoms` the `token` wouldn't be necessary right?

EnoF (approved):

Randynamic (approved):

packages/libs/kode-ui/src/components/Form/Select/Select.css.ts
sstraatemans: created a story to cleanup the CSS
https://app.asana.com/0/1206752931509995/1208343913799309/f

PR #2542, opened 2024-09-24
feat(tools): adds testnet05 as network

alber70g:


sstraatemans (approved):

Randynamic (approved):
The module explorer doesn't work yet correct

alber70g:
> The module explorer doesn't work yet correct

@Randynamic testnet05 doesn't have `(list-modules)` yet... It won't work until we have that

PR #2570, opened 2024-10-10
feat(dw): networks page + create client based on networks in db

javadkh2:
### Modify this title

Short description: \_

### Test scenarios

### Reminders (if applicable)

      projects)
      docs.kadena.io)


packages/apps/dev-wallet/src/modules/account/account.repository.ts
```diff
@@ -94,10 +95,10 @@ const createAccountRepository = ({
       );
       return accounts;
     },
-    async getAccountsByProfileId(profileId: string, networkId: string) {
+    async getAccountsByProfileId(profileId: string, networkUUID: UUID) {
```
alber70g, line 98: Should the name change to `getAccountsByProfileAndNetwork`? or `getAccounts`

packages/apps/dev-wallet/src/modules/account/account.service.ts
```diff
@@ -112,7 +119,7 @@ export const accountDiscovery = (
         await emit('key-retrieved')(key);
         const chainResult = (await discoverAccount(
           `k:${key.publicKey}`,
-          networkId,
+          network.networkId,
           undefined,
           contract,
         )
```
alber70g, line 125: Should the variable name change to `networkUUID` for `discoverAccount`s 2nd argument?

packages/apps/dev-wallet/src/modules/account/account.service.ts
```diff
@@ -184,11 +191,12 @@ export const accountDiscovery = (
 
 export const syncAccount = async (account: IAccount) => {
   console.log('syncing account', account.address);
+  const network = await networkRepository.getNetwork(account.networkUUID);
   const updatedAccount = { ...account };
 
   const chainResult = (await discoverAccount(
     updatedAccount.address,
-    updatedAccount.networkId,
+    network.networkId,
```
alber70g, line 199: Idem

packages/apps/dev-wallet/src/modules/account/account.service.ts
```diff
@@ -235,18 +243,38 @@ export async function fundAccount({
   address,
   keyset,
   profileId,
-  networkId = 'testnet04',
+  network,
```
alber70g, line 246: Do we need the full network here or is `networkUUID` sufficient?

packages/apps/dev-wallet/src/modules/network/network.service.ts
```diff
@@ -0,0 +1,58 @@
...
+          ...network,
+          hosts: hosts.filter((host) => host.isHealthy),
+        };
+      }),
+    );
+  };
+  checkNetworks();
+  return ({ networkId, chainId }: INetworkOptions) => {
+    if (Date.now() - lastCheckTime > 30000) {
+      checkNetworks();
```
alber70g, line 46: do we need to await `checkNetworks` at some time? Or return the promise

packages/apps/dev-wallet/src/modules/wallet/wallet.provider.tsx
```diff
@@ -52,11 +56,14 @@ export const syncAllAccounts = throttle(AccountService.syncAllAccounts, 10000);
 export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
   const [contextValue, setContextValue] = useState<ExtWalletContextType>({
     networks: [],
+    client: createClient(),
```
alber70g, line 59: Shouldn't we pass the hostGenerator based on the active network? As without hosteGenerator we can only work with mainnet and testnet

alber70g (commented):
Minor questions

alber70g (approved):
Minor questions

packages/apps/dev-wallet/src/modules/account/account.service.ts
javadkh2: `discoverAccount` is imported form `client-utils` it only needs networkId,
the client is injected via the wallet so it will work with correct network incase if there are two networks with same networkId

packages/apps/dev-wallet/src/modules/account/account.service.ts
javadkh2: I need both uuid and networkId there. 
I can read networkId from db inside the function again but since the network object was already in the context to passed it down

packages/apps/dev-wallet/src/modules/network/network.service.ts
javadkh2: not ready. it just checks the network health, I don't want to stop the whole process for this check
but I will add catch handler to it

javadkh2: basically it filters the healthy networks for the next call not the current one

PR #2604, opened 2024-10-22
Feat/contact page

javadkh2:
### Modify this title

Short description: \_

### Test scenarios

### Reminders (if applicable)

      projects)
      docs.kadena.io)


packages/apps/dev-wallet/src/modules/account/account.repository.ts
```diff
@@ -87,15 +98,23 @@ const createAccountRepository = ({
...
     updateAccount: async (account: IAccount): Promise<void> => {
       return update('account', deleteKey(account, 'keyset'));
     },
+    deleteAccount: async (uuid: string): Promise<void> => {
+      return remove('account', uuid);
+    },
     getAccount: async (id: string) => {
       const account: IAccount = await getOne('account', id);
-      return appendKeyset(account);
+      return appendKeyset(account as IAccount);
```
alber70g, line 110: Why is this `as IAccount` necessary? The `const account` is already typed as `IAccount`. 
I saw the `as IAccount` in more places.

alber70g (approved):

packages/apps/dev-wallet/src/modules/account/account.repository.ts
javadkh2: nice catch! I was doing something that I removed eventually but forgot to clean up this line

PR #2620, opened 2024-10-25
feat(ui): add sidebar component

sstraatemans:
---
  - https://app.asana.com/0/0/1208619870274433


Randynamic (approved):

packages/apps/dev-wallet/src/App/Layout/SideBar.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const handleLogout = () => {
+    lockProfile();
+  };
+  return (
+    <SideBarUI
+      appContext={
+        <SideBarItem
+          visual={<MonoNetworkCheck />}
+          label="Select network"
+          onPress={() => {}}
```
Randynamic, line 53: Are we going to add another function later?

packages/apps/dev-wallet/src/App/Layout/SideBar.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    isCompact={!isExpanded}
+                    variant="outlined"
+                    endVisual={<MonoContacts />}
+                  >
+                    {isExpanded ? 'Profile' : undefined}
+                  </Button>
+                }
+              >
+                <ContextMenuItem
+                  onClick={() => navigate('/profile')}
```
Randynamic, line 132: Don’t we need onPress on all?

packages/apps/dev-wallet/src/App/Layout/SideBar.tsx
isaozler: image

This is indeed a dynamic area where we will have different callbacks later on

isaozler (approved):

PR #2631, opened 2024-11-04
chore(devwallet): set new network in aside

sstraatemans:


packages/apps/dev-wallet/src/Components/Aside/Aside.tsx
```diff
@@ -3,6 +3,10 @@ import { useLocation } from 'react-router-dom';
 
 const importView = async (key: string) => {
   switch (key) {
+    case 'AddNetwork':
```
javadkh2, line 6: This approach wont work with hashRouter which we might need to use for extension, 
Also this way makes the form disconnected form the the page itself. 
We can use portal view instead. 
Also we don't need to do lazy loading in this level, we will improve this later in the page level. for example network page will have its own bundle including the add form.
We will make this kind of decisions later when we make the PWA ready.

PR #2669, opened 2024-11-19
ag/graph/dockerfile

alber70g:
- **fix(graph): .env.example for remote graph db**
- **feat(graph): adds Dockerfile and docker-compose.yaml**


lbpage (approved):
LGTM

PR #2723, opened 2024-12-11
fix(rwa): init the contract with the correct admin keyset

sstraatemans:


.changeset/rich-horses-vanish.md
alber70g, line 1: You don't need changesets for this. They're not published, so you can omit them

packages/apps/rwa-demo/src/services/createContract.ts
```diff
@@ -26,3 +43,15 @@ export const createContract = async (
...
+//     (read-keyset 'keyset)
+//     (read-keyset 'keyset)
+//   )
+//   (namespace (ns.create-principal-namespace (read-keyset 'keyset)))
+
+//       (let ((keyset-name:string (format "{}.{}" [(ns.create-principal-namespace (read-keyset 'keyset)) 'admin-keyset]) ))
+//     (define-keyset keyset-name (read-keyset 'keyset))
+//     (enforce-keyset keyset-name)
+//     keyset-name
+//   )
```
alber70g, line 57: Can be removed?

alber70g (approved):

packages/apps/rwa-demo/src/services/createContract.ts
sstraatemans: a yes..will do

PR #2726, opened 2024-12-11
fix(graph): transfers query when using pagination

alber70g:


sstraatemans (approved):

packages/apps/graph/generated-schema.graphql
```diff
@@ -225,6 +224,9 @@ type GasLimitEstimation {
 type GraphConfiguration {
   """The lowest block-height that is indexed in this endpoint."""
   minimumBlockHeight: BigInt
+
+  """The version of the @kadena/graph package."""
+  version: String!
```
Randynamic, line 229: 👍

Randynamic (approved):

packages/apps/graph/src/graph/objects/fungible-account.ts
```diff
@@ -48,343 +49,220 @@ interface ITransferQueryConditions {
...
+        ORDER BY "height" DESC, "chainid" DESC, "requestkey" DESC, "idx" DESC
+        OFFSET ${condition.skip}
+        LIMIT ${condition.take};
+      `;
+
+      return result;
+    } else {
+      // Backward pagination with cursor (take < 0)
+      const positiveTake = -condition.take;
+      log('condition is negative, page back, take:', positiveTake);
```
Randynamic, line 223: Do we want to keep the logs?

alber70g: Yes, they only output when in dev mode.

PR #2741, opened 2024-12-16
ag/explorer/remove accountname from transactions

alber70g:
- **chore(graph): improve transactions query. Don't load signers when not requested**
- **chore(explorer): remove column accountname when querying for transactions for an account**


alber70g:
Combined in #2740

PR #2744, opened 2024-12-17
chore(rwa): add extra check on deleting an investor. there should be no balance

sstraatemans:


packages/apps/rwa-demo/src/components/InvestorList/InvestorList.tsx
```diff
@@ -99,17 +95,7 @@ export const InvestorList: FC = () => {
...
-                      <Button
-                        isCompact
-                        isDisabled={!isDeleteInvestorAllowed}
-                        variant="outlined"
-                        startVisual={<MonoDelete />}
-                        onPress={handleDelete}
-                      />
-                    ),
-                  }),
+                  render: FormatDeleteInvestor(),
```
EnoF, line 98: is this a `class` or a `Component`? This seems odd

alber70g (approved):

EnoF (approved):

packages/apps/rwa-demo/src/components/InvestorList/InvestorList.tsx
sstraatemans: TLDR: this is the way the component that will be rendered in this data field is created

its a function that returns the actual component that will be rendered in the datacell.
this way you can add all kind of properties to the component.

and the real component will be created inside the data-cell with the value that can be rendered (this is done by the compactDataTable component).
and we can use the extra props from the first function inside the component.
this one does not need any extra information. so no props are given here.

kind of hard to explain without showing you.

PR #2781, opened 2025-01-09
feat(client-example): deploy contract example using @kadena/client

javadkh2:
### Modify this title

Short description: \_

### Test scenarios

### Reminders (if applicable)

      projects)
      docs.kadena.io)


trendzetter:
The example could be improved: it uses hard-coded chain id. It deploys only on one chain, probably more realistic is if it deploys to all chains in parallel. The example uses outdated practices, eg "free" namespace instead of principal namespaces.

lsgunnlsgunn (approved):

PR #2783, opened 2025-01-10
update schema to represent correct structure in preparation of transition to kadindexer.io/graphql

alber70g:
- **chore(explorer): adds hackachain configs**
- **feat(graph): update schema to represent correct structure in preparation of transition to kadindexer.io/graphql**


packages/apps/rwa-demo/codegen-sdk.yml
```diff
@@ -1,6 +1,6 @@
 overwrite: true
-# schema: ./node_modules/@kadena/graph/generated-schema.graphql
-schema: https://graph.kadena.network/graphql
+schema: ./node_modules/@kadena/graph/generated-schema.graphql
+# schema: https://graph.kadena.network/graphql
```
sstraatemans, line 3: doesnt this need the kdaindexer schema as well?

alber70g: We'll first deploy explorer with kadindexer. Lateron we can modify this. The current changes are based on the diff between the two endpoints.

```sh
bunx @graphql-inspector/cli diff https://graph.kadena.network/graphql https://www.kadindexer.io/graphql > diff-kadena-hackachain-schema.log
```

packages/apps/graph/src/graph/objects/transfer.ts
```diff
@@ -25,6 +22,25 @@ export default builder.prismaNode(Prisma.ModelName.Transfer, {
     moduleName: t.exposeString('moduleName'),
     requestKey: t.exposeString('requestKey'),
     receiverAccount: t.exposeString('receiverAccount'),
+    height: t.expose('height', { type: 'BigInt' }),
+    blockHash: t.exposeString('blockHash'),
+    chainId: t.expose('chainId', { type: 'BigInt' }),
+
+    // blockHash: t.string({
+    //   deprecationReason: 'Use `block.hash` field instead.',
```
sstraatemans, line 30: do we still need this?

sstraatemans (approved):

packages/apps/graph/src/graph/objects/transfer.ts
alber70g: Good one. But not for the reason you think. I need to add the deprecation message to the normal fields. Then we'll upgrade to the ones that are commented. Then we'll remove them

PR #2814, opened 2025-01-21
graph cron: use totalCutHeight from Graph by adding up all heights

alber70g:
- use totalCutHeight from Graph by adding up all heights. 
- Add lastBlockHeight to message. 
- Adds types

---
  - https://app.asana.com/0/0/1209191585829839

### DIFF
--- a/packages/apps/tools/src/scripts/graphCron/index.ts
+++ b/packages/apps/tools/src/scripts/graphCron/index.ts
@@ -5,23 +5,51 @@ interface IEnvProps {
   env: 'testnet04' | 'mainnet01';
   chainweb: string;
   graphql: string;
+  graphqlRef: string;
 }
 
-const countHeightOnChainweb = async (props: IEnvProps): Promise<number> => {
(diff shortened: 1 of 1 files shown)


packages/apps/tools/src/scripts/graphCron/index.ts
```diff
@@ -67,32 +120,47 @@ export const runJobPerEnvironment = async (props: IEnvProps) => {
...
+[GraphiQL lastBlockHeight](${props.graphqlRef}): ${lastBlockHeightGraph}
+Total height difference: ${lastBlockHeightChainweb - lastBlockHeightGraph}
+`,
       });
     }
   } catch (e) {
     await sendErrorMessage({
       title: `There was a general issue with the ${props.env} graph cron job`,
-      msg: ``,
+      msg: e,
```
sstraatemans, line 141: are we sure this is a string? or an object? if object this will fail. maybe stringify to be sure?

sstraatemans (approved):

Issue #2832, opened 2025-01-27
Template not passed correctly with --template flag


V-Staykov:
### Please confirm the following:

- [x] I have read the [Code of Conduct](https://github.com/kadena-community/kadena.js/blob/main/CODE_OF_CONDUCT.md)
- [x] The bug was not reported before
- [x] I am using the **latest version** of the relevant package(s)

### What operating system are you using? And what version?

Ubuntu 22.x

### What environment or browser are you using? And what version?

Node.js

### What package is the bug report about?

@kadena/kadena-cli

### Describe the bug

When I use the `--template` flag, I get 
```
Failed to create transaction from template
Failed to parse template: payload: 
meta:
```

I debugged it and found out that the template is actually found and parsed correctly, but then a string is passed to `createAndWriteTransaction`. It seems to expect a template string, but instead just the name is passed, which is then split into parts and holes in `getPartsAndHoles` and it returns gibberish.

### Steps to reproduce the behavior

use `kadena tx add --template="transfer.ktpl"

### What did you expect to happen?

To find and parse the correct template.

### Participation

- [x] I am willing to submit a pull request for this issue.

Randynamic:
@V-Staykov TY for the PR we will review.

PR #2833, opened 2025-01-27
fix(kadena-cli): pass correct template string when --template is used

V-Staykov:
### Modify this title

Related Issue/Asana ticket: https://github.com/kadena-community/kadena.js/issues/2832

Short description: \_
Title: Fix `--template` flag but
Related Issue:[ #2832](https://github.com/kadena-community/kadena.js/issues/2832)
Short Description: When the --template flag is used, the template file name string is passed into `createAndWriteTransaction` instead of the template string which is then used. This is leading to error that the template can't be parsed.

### Test scenarios
Execute a `tx add` command with added `-template=transfer.ktpl` flag.

### DIFF
--- a/.changeset/breezy-zoos-heal.md
+++ b/.changeset/breezy-zoos-heal.md
@@ -1,0 +1,6 @@
+---
+'@kadena/kadena-cli': patch
+---
+
+fix a bug where a template filename is passed to createAndWriteTransaction
+instead of the actual template string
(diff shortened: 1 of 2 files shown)


barthuijgen (approved):
Tested this fix and it works fine. Seems there is an additional issue in the debug output "executed: ..." line using all sorts of wrong flags for template data. But this change is correct.

barthuijgen:
Merged via different PR so our CI pipelines would run https://github.com/kadena-community/kadena.js/pull/2916

PR #2839, opened 2025-01-28
feat(wallet-sdk): added support for finding accounts by publickey

nillo:
[x] added support for finding accounts by publickey


alber70g (approved):

packages/libs/wallet-sdk/etc/wallet-sdk.api.md
```diff
@@ -22,8 +22,12 @@ export type ChainwebHostGenerator = (options: {
 // @public (undocumented)
 export type GraphqlHostGenerator = (options: {
     networkId: string;
+    graphType?: GraphType;
 }) => string;
 
+// @public (undocumented)
+export type GraphType = 'kadena' | 'hackachain';
+
```
alber70g, line 30: Why do we distinguish between the two graphs?
It should be that everything works on both endpoints, EXCEPT the accountsByPublicKey. All the others should be backwards compatible (potentially with deprecation message, but not failing)

packages/libs/wallet-sdk/src/gql/gql.ts
alber70g, line 1: This one's generated, correct? If so, can we move it to a directory that shows that? Something like __generated__ or equivalent.

Same for graphql.ts

packages/libs/wallet-sdk/src/sdk/host.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * @public
+ */
+export type GraphType = 'kadena' | 'hackachain';
+
```
alber70g, line 5: See comment above

packages/libs/wallet-sdk/src/sdk/host.ts
```diff
@@ -9,32 +14,64 @@ export type ChainwebHostGenerator = (options: {
...
+const kadenaGraphqlHostMap: Record<string, string> = {
+  mainnet01: 'https://graph.kadena.network/graphql',
+  testnet04: 'https://graph.testnet.kadena.network/graphql',
+};
+
+// Hackachain GraphQL Host Map
+const hackachainGraphqlHostMap: Record<string, string> = {
+  mainnet01: 'https://www.kadindexer.io/graphql',
+  testnet04: 'https://www.kadindexer.io/graphql',
+};
```
alber70g, line 32: Is this determined by the consumer or in the library itself?
Can the consumer override it?

alber70g (approved):

alber70g (changes requested):
Comment regarding two graphql types. Should only be one, and have a sensible error when the `accountsByPublicKey` is used on the one that doesn't support it.

nillo:
On hold until hackachain provides testnet

PR #2858, opened 2025-02-07
fix: get pact error code function

javadkh2:
### Modify this title

Short description: \_

### Test scenarios

### Reminders (if applicable)

      projects)
      docs.kadena.io)

### DIFF
--- a/packages/libs/client/src/utils/getPactErrorCode.ts
+++ b/packages/libs/client/src/utils/getPactErrorCode.ts
@@ -15,7 +15,7 @@ export type PactErrorCode =
  * This function is compatible with both Pact 4 and Pact 5 error formats.
  *
  * @param  error - The error returned by Pact.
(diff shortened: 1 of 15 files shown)


packages/libs/client/src/utils/getPactErrorCode.ts
```diff
@@ -0,0 +1,39 @@
+export type PactErrorCode =
+  | 'RECORD_NOT_FOUND'
+  | 'DEFPACT_COMPLETED'
+  | 'CANNOT_RESOLVE_MODULE'
+  | 'EMPTY_CODE'
+  | 'ERROR';
+
+export function getPactErrorCode(
```
alber70g, line 8: We should add some docs above this one with a description that tells the user that it works for pact4 and pact5

alber70g (approved):

alber70g (approved):

PR #2863, opened 2025-02-13
chore(cw3): update plugin system to be more generic

alber70g:


packages/apps/chainweaver-pact-console-plugin/vite.config.ts
```diff
@@ -57,11 +59,11 @@ export const config: UserConfig = {
   build: {
     ssr: false,
     minify: false,
-    sourcemap: 'hidden',
+    sourcemap: 'inline',
```
javadkh2, line 62: isn't better if we create separate source map file?

packages/apps/chainweaver-pact-console-plugin/package.json
```diff
@@ -75,5 +70,12 @@
     "vite-plugin-static-copy": "^1.0.0",
     "vite-tsconfig-paths": "^4.2.1",
     "vitest": "^1.6.0"
+  },
+  "chainweaver-plugin-manifest": {
+    "shortName": "Pact Console",
+    "description": "Remote Pact Console",
```
javadkh2, line 76: I think we can remove shortName and have `name` and `description`. then for this plugin would be 
```
"name": "Pact Remote Console",
"description": "A console for interacting remotely with Pact on different networks (read-only)",
```

packages/apps/dev-wallet/scripts/symlink-plugins.ts
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ./node_modules/.bin/tsx
+import path from 'path';
+import { $ } from 'zx';
+
+if (path.basename(process.cwd()) !== 'dev-wallet') {
+  throw new Error('This script should be run from the dev-wallet directory');
+}
+
+const { dependencies } = await $`cat package.json`.json();
```
javadkh2, line 9: You can import a json file just like a module

packages/apps/chainweaver-wallet-connect-plugin/src/component/WalletConnect.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useEffect, useState } from 'react';
+
+const core = new Core({
+  projectId: '1d8fe1e80daaaf10e5e793df2bffdaab',
+});
+
+const metadata = {
+  name: 'Kadena Chainweaver v3',
+  description: 'AppKit Example',
+  url: 'https://reown.com/appkit', // origin must match your domain & subdomain
```
javadkh2, line 12: we should check if actually this work since we load iframe with srcDoc that would be just local file

javadkh2 (approved):

packages/apps/chainweaver-pact-console-plugin/vite.config.ts
alber70g: Yes, we can do it, but for now it's nice to have it on hand in the browser

PR #2899, opened 2025-03-06
fix(wallet): not possible anymore to choose an existing profilename during recovery

sstraatemans:


packages/apps/dev-wallet/src/pages/wallet-recovery/recover-from-mnemonic/recover-from-mnemonic.tsx
```diff
@@ -57,12 +57,10 @@ export function RecoverFromMnemonic() {
     control,
     formState: { isValid, errors },
   } = useForm<Inputs>({
+    mode: 'onChange',
     defaultValues: {
       mnemonic: '',
-      profileName:
-        profileList.length === 0
-          ? 'default'
-          : `profile-${profileList.length + 1}`,
```
alber70g, line 65: We don't want to prefill with a default value so the user can press next immediately?

alber70g (approved):

PR #2902, opened 2025-03-10
feat(wallet): maintenance mode

sstraatemans:


packages/apps/dev-wallet/src/utils/env.ts
alber70g, line 1: I suppose we can't use dotenv like `packages/apps/graph/src/utils/dotenv.ts` because it's not a nodejs app but client only, right?

packages/apps/dev-wallet/src/Components/MaintenanceProvider/MaintenanceProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  useEffect(() => {
+    const date = Date.now();
+
+    // there should be at least an enddate if you want to maintenance message to show
+    setIsInMaintenance(false);
+    if (!env.MAINTENANCE_STARTDATE && !env.MAINTENANCE_ENDDATE) return;
+    if (date > env.MAINTENANCE_STARTDATE && date < env.MAINTENANCE_ENDDATE) {
+      setIsInMaintenance(true);
+    }
```
alber70g, line 40: Shouldn't the minimum be only a startdate? That way it'll be in maintenance until the error is fixed. I also like to have a way to override this, something like `?ignoreMaintenance=true` that's stored in localStorage until the enddate has passed

alber70g (approved):

packages/apps/dev-wallet/src/App/app.tsx
```diff
@@ -23,8 +24,10 @@ function Providers({ children }: { children: React.ReactNode }) {
             <PromptProvider>
               <DatabaseProvider>
                 <WalletProvider>
-                  {/* TODO: fixed the issue with prompt and remove this one in favor of the one above */}
-                  <PromptProvider>{children}</PromptProvider>
+                  <MaintenanceProvider>
```
javadkh2, line 27: I would say if you don't use any wallet api in the provider then move it to the root level so the wallet wont load at all on maintenance mode.

javadkh2 (approved):

packages/apps/dev-wallet/src/Components/MaintenanceProvider/MaintenanceProvider.tsx
sstraatemans: good one

packages/apps/dev-wallet/src/App/app.tsx
sstraatemans: 👍🏻

alber70g (approved):

PR #2920, opened 2025-03-21
port to pact 5 error message

abuvanth:
exception occurred due to mismatched error message


alber70g (approved):

alber70g:
@abuvanth I'm fixing this one in #2932

alber70g:
Hi @abuvanth, did this work on your side? This actually fixed the issue? I get breaking tests and the error that's been thrown in the code above your change isn't modified to throw that specific error

PR #2937, opened 2025-03-31
feat(wallet): new transfer flow

sstraatemans:


packages/apps/dev-wallet/src/Components/Assets/AssetCards.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => {
+  const [showAll, setShowAll] = useState(false);
+
+  const { filteredAssets, hasShowAllButton } = useMemo(() => {
+    if (showAll || assets.length <= MAXASSETCOUNT) {
+      return { filteredAssets: assets, hasShowAllButton: false };
+    }
+
+    return {
+      filteredAssets: assets.slice(0, MAXASSETCOUNT - 1),
```
EnoF, line 31: I'm not quite sure of the intend of this

packages/apps/dev-wallet/src/pages/transaction/components/Signer.tsx
```diff
@@ -81,7 +81,7 @@ export const RenderSigner = ({
         {signer.clist &&
           signer.clist.map((cap) => (
             <Stack
-              key={cap.name}
+              key={`${cap.name}${cap.args.toString()}`}
```
EnoF, line 84: don't think this does what you intend for it to do, I think you'd rather want `cap.args.join(',')` as this result in `[object Object]`, right?

EnoF (approved):

javadkh2:
Selecting chain is mandatory in some scenarios like when you select the same account as receiver then the user need to select target chain explicitly.

packages/apps/dev-wallet/src/pages/transfer/Components/AccountSearchBox.tsx
```diff
@@ -209,18 +210,18 @@ export function AccountSearchBox({
   ) {
     if (accounts.length === 1) {
       const account = accounts[0];
-      onSelectHandel(account);
```
isaozler, line 212: 😅

PR #2940, opened 2025-04-09
chore(wallet): missing dependency for chainweaver-wallet-connect-plugin

sstraatemans:


javadkh2 (approved):
We need to investigate this since the ws is not a direct dependency

PR #2965, opened 2025-05-06
Chainweaver Wallet Adapter

nillo:
ChainWeaver Wallet adapter

Add new adapter for chainweaver wallet connect


packages.json
```diff
@@ -203,6 +203,12 @@
     "private": false,
     "path": "packages/libs/wallet-adapter-ecko"
   },
+  {
+    "name": "@kadena/wallet-adapter-chainweaver-legacy",
+    "version": "0.0.1-beta.0",
+    "private": false,
```
barthuijgen, line 209: This one needs to be true for now, I need to update the rest as well. Until we do non-beta releases.

barthuijgen (approved):

packages/libs/wallet-adapter-chainweaver-legacy/src/provider.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  if (silent !== undefined && typeof silent !== 'boolean') {
+    throw new Error('Expected option "silent" to be a boolean.');
+  }
+
+  if (timeout !== undefined && typeof timeout !== 'number') {
+    throw new Error('Expected option "timeout" to be a number.');
+  }
+
+  const provider: IChainweaverProviderLegacy = {
```
alber70g, line 35:
```suggestion
  const provider: IChainweaverLegacyProvider = {
```
Idem. ChainweaverLegacy is the subject, Provider is the type

alber70g (approved):

packages/libs/wallet-adapter-chainweaver-legacy/package.json
```diff
@@ -0,0 +1,39 @@
+{
+    "name": "@kadena/wallet-adapter-chainweaver-legacy",
```
alber70g, line 2: Should we name it with `v2` as it applies to that one. Otherwise it might be confusing when there's v3 and maybe another v4 where v2 and v3 are legacy.

```suggestion
    "name": "@kadena/wallet-adapter-chainweaver-v2",
```

alber70g (approved):

packages/libs/wallet-adapter-chainweaver-legacy/package.json
nillo: denied 🗡️

packages/libs/wallet-adapter-chainweaver-legacy/src/provider.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param options - Options for detection (e.g. timeout, silent).
+ * @returns A promise resolving to the Chainweaver Legacy provider or null.
+ * @public
+ */
+export async function detectChainweaverLegacyProvider(options?: {
+  silent?: boolean;
+  timeout?: number;
+  // eslint-disable-next-line @rushstack/no-new-null
+}): Promise<IChainweaverLegacyProvider | null> {
```
alber70g, line 24: Why not `undefined`?

alber70g (approved):

packages/libs/wallet-adapter-chainweaver-legacy/src/ChainweaverLegacyAdapter.ts
```diff
@@ -0,0 +1,157 @@
...
+        return {
+          id: this.nonce,
+          jsonrpc: '2.0',
+          result: data,
+        } as JsonRpcResponse<unknown>;
+      }
+
+      case 'kadena_sign_v1': {
+        const signingCmd = parsedParams as ISigningRequestPartial;
+        const resp = await fetch('http://127.0.0.1:9467/v1/sign', {
```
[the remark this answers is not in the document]
nillo: This is the path to the local api, that is superseeded by modern wallets anyways, yet chainweaver legacy uses it, and this is its hardcoded path. so idd a nice to have. but i think is best to leave it as is

packages/libs/wallet-adapter-chainweaver-legacy/src/provider.ts
nillo: I want to be explicit... in our case null means “I attempted to get a provider, and none exists,” which is different from “not set yet” or “omitted” in our other adapters. By using null, we stay consistent with the rest of the codebase and avoid breaking the core detect function’s signature, which expects IProvider | null. (The local call is a bit of a edge case, but we are consitent across all adapters using null)

PR #2971, opened 2025-05-07
feat(client): add gas estimation + change default TTL to 15 min

javadkh2:
**client**:
* Changed default TTL to 900 (15 min)

**cleint-utils**:
* `estimateGasPrice`:  estimate gas price on a chain
```TS
const currentGasPriceEstimation: number = await estimateGasPrice({ chainId: "0" });
```
* `getBlocksGasInformation`: returns useful information about gas 
```TS
const gasInformation :IBlockGasInformation[] = await getBlocksGasInformation({ chainId: "0" });

type IBlockGasInformation = Array<{
  blockHeight: number;
  totalGasConsumed: number; // sum of all gas consumed in the block transactions extracted from gas field of tx result
  totalGasLimit: number; // sum of all gas limits in the block transactions extracted from gasLimit field of tx command
  totalGasPaid: number; // sum of all gasPrice * gasLimit in the block transactions extracted from meta field of tx command
  txCount: number; // number of transactions in the block
  blockGasUsedPercent: number; // percentage of gas used in the block (totalGasConsumed / MAX_BLOCK_CAPACITY)
  gasPriceStats: {
    min: number; // minimum gas price in the block transactions
    max: number; // maximum gas price in the block transactions
    avg: number; // average gas price in the block transactions (does not consider usage)
    median: number; // median gas price in the block transactions
  }>;
}
```

### DIFF
--- a/packages/libs/client-utils/src/core/estimate-gas.ts
+++ b/packages/libs/client-utils/src/core/estimate-gas.ts
@@ -1,6 +1,7 @@
 import type { IPartialPactCommand } from '@kadena/client';
 import { createTransaction } from '@kadena/client';
 
+import { estimateGasPrice } from './estimateGasPrice';
 import { asyncPipe } from './utils/asyncPipe';
 import type { IClientConfig } from './utils/helpers';
 import { composeWithDefaults, getClient, throwIfFails } from './utils/helpers';
@@ -35,10 +36,9 @@ export const calculateGasConsumption = (
 };
 
 /**
- * estimate gas for a command;
- * this function always returns a gasPrice of 1.0e-8, if you want to estimate the gas price use `estimateGasPrice` which is more accurate
+ * estimate gasLimit and gasPrice for a command;
+ * This function might be slow since it needs to estimate the gasPrice on each call, use it with caution
  * @alpha
- * @deprecated for gasLimit use `calculateGasConsumption` and for gasPrice use `estimateGasPrice`
  * @see {@link estimateGasPrice}
  * @see {@link calculateGasConsumption}
  */
@@ -52,8 +52,16 @@ export const estimateGas = async (
   client = getClient(host),
 ) => {
   const gasLimit = await calculateGasConsumption(command, host, client);
+  const cmd = typeof command === 'function' ? command() : command;
+  if (!cmd.meta?.chainId) {
+    throw new Error('chainId is required in meta');
+  }
   return {
     gasLimit,
-    gasPrice: 1.0e-8,
+    gasPrice: await estimateGasPrice({
+      host: typeof host === 'string' ? host : undefined,
+      networkId: cmd.networkId,
+      chainId: cmd.meta.chainId,
+    }),
   };
 };
--- a/packages/libs/client-utils/src/core/estimate-gas.ts
+++ b/packages/libs/client-utils/src/core/estimate-gas.ts
@@ -62,6 +62,7 @@ export const estimateGas = async (
       host: typeof host === 'string' ? host : undefined,
       networkId: cmd.networkId,
       chainId: cmd.meta.chainId,
-    }),
+      // fallback to minimum gas price if estimation fails
+    }).catch(() => 1e-8),
   };
 };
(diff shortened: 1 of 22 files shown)


packages/libs/client-utils/src/core/estimate-gas.ts
```diff
@@ -22,14 +22,38 @@ export const estimateGas = (
...
     throwIfFails,
-    (response) => ({ gasLimit: response.gas, gasPrice: 1.0e-8 }),
+    (response) => response.gas,
   );
   return pipeLine(command);
 };
+
+/**
+ * estimate gas for a command;
+ * this function always returns a gasPrice of 1.0e-8, if you want to estimate the gas price use `estimateGasPrice` which is more accurate
```
edmundnoble, line 39: Would it be unreasonable to make `estimateGas` use the new `estimateGasPrice`? It seems like that would get people using it more quickly. The more common use case should be using a gas price estimate and not a hard-coded price.

javadkh2: I can add it as well. I thought we could deprecate this function, but it’s actually useful since it provides both gasLimit and gasPrice. However, it calls the block endpoint on every call, which would be slower

javadkh2: updated the function

edmundnoble (approved):
This PR implements the algorithm we've discussed offline as well as the TTL change, so I approve it.

packages/libs/client-utils/src/core/estimateGasPrice/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  }>,
+) {
+  const mins = gasData
+    .filter((b) => b.txCount > 0)
+    .map((data) => data.gasPriceStats.min)
+    .sort((a, b) => a - b);
+
+  const estimate = median(mins);
+
+  return estimate || MINIMUM_GAS_PRICE;
```
barthuijgen, line 154: If it's a minimum instead of default, shouldn't this be `Math.max(estimate || 0, MINIMUM_GAS_PRICE)`

packages/libs/client-utils/src/core/estimateGasPrice/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+    median: number; // median gas price in the block transactions
+  };
+}
+
+/**
+ * Calculates gas information for a given set of blocks.
+ * @param blocks - An object containing block information.
+ * @returns An array of gas information for each block.
+ */
+export function calculateGasInformation(
```
barthuijgen, line 71: I'd add a try catch to catch the various potential errors from base64decoding and json parsing and provide more helpful errors.

alber70g (approved):
Should we use `MINIMUM_GAS_PRICE` throughout the app where we use the literal value `1e-8`?

packages/libs/client-utils/src/core/estimateGasPrice/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+        } catch (e) {
+          console.error('Error parsing transaction data:', e);
+          return {
+            hash: 'N/A',
+            gasPrice: 0,
+            gasLimit: 0,
+            consumedGas: 0,
+            gasPaid: 0,
+          };
+        }
```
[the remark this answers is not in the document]
javadkh2: I thought about throwing and exception but I think thats too much by the end this is just estimation

packages/libs/client-utils/src/core/estimateGasPrice/utils.ts
```diff
@@ -0,0 +1,166 @@
...
+
+export const average = (arr: number[]): number => {
+  if (arr.length === 0) {
+    return 0;
+  }
+  return arr
+    .reduce((acc, gasFee) => acc.plus(gasFee), new PactNumber(0))
+    .dividedBy(arr.length)
+    .toNumber();
+};
```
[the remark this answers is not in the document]
javadkh2: Yes I think thats a good idea. Also I should rename the `gasFee` here to something more general

PR #3041, opened 2025-07-01
fix(wallet): only use the copyTransactionAs legacy, which works everywhere

sstraatemans:
---
  - https://app.asana.com/0/0/1210641578236005


javadkh2 (approved):
looks file

PR #3042, opened 2025-07-01
fix(wallet): can only unsign own account

sstraatemans:
---
  - https://app.asana.com/0/0/1210641578236001


packages/apps/dev-wallet/src/modules/wallet/wallet.hook.tsx
```diff
@@ -167,6 +167,19 @@ export const useWallet = () => {
     [getPublicKeyData, context.profile?.uuid],
   );
 
+  const isOwnedPubkey = useCallback(
```
javadkh2, line 170: `getPublicKeyData` from `useWallet` is exactly for this purpose, and covers all keys not only the keys used in accounts. 
because users can create keys without assigning them to any account

javadkh2 (changes requested):

javadkh2 (approved):

PR #3051, opened 2025-07-08
Fix/kadena cli format error

javadkh2:
Fixed showing error messages for nested properties in template files


alber70g (approved):

barthuijgen (approved):
Thanks, LGTM!

javadkh2:
I don’t think additional (regression) tests are needed since all related apps and libraries are in the monorepo, and the CI already runs all unit, integration, and end-to-end tests to ensure nothing is broken. Also, this change only modifies the error message formatting and does not involve any logic changes.

PR #3069, opened 2025-08-05
Feat: MM Snap Wallet Adapter

chicodias:
### Add Metamask Snap Adapter and `changeNetwork_v1` to Base Adapter

Short description:
Adds the Snap adapter with support for several Wallet API methods. Also introduces the `changeNetwork_v1` method to the Base Adapter. The React example is updated to reflect usage of `BaseAdapter` v0.0.2 with account type validation.

The Snap adapter has the following methods:

* `connect`
* `getAccount_v1`

  > Works similarly to Zelcore Adapter — returns only the first account
* `getAccounts_v2`
* `getNetwork_v1`
* `getNetworks_v1`
* `quicksign_v1`
* `changeNetwork_v1`
* `checkStatus`

**Note:** `sign_v1` is not implemented, as the wallet natively only supports `quicksign`.

---

## Base Adapter

* Add `changeNetwork_v1` method

---

## React Example

* Updated to use `BaseAdapter` v0.0.2 with account type validation


packages/apps/wallet-adapter-react-example/package.json
```diff
@@ -29,6 +30,7 @@
   },
   "devDependencies": {
     "@kadena-dev/eslint-config": "workspace:*",
+    "@metamask/providers": "^22.1.0",
```
barthuijgen, line 33: I don't see `@metamask/providers` being used in `wallet-adapter-react-example` and also wouldn't expect to need another dependency, if something from there is needed then `@kadena/wallet-adapter-snap` should re-export it.

.changeset/green-stingrays-live.md
```diff
@@ -0,0 +1,9 @@
+---
+'@kadena/wallet-adapter-snap': major
```
barthuijgen, line 2: I've been avoiding `major` to keep versions below 1.0. We did stop using `beta` tag but I'd prefer our packages to be seen as an early project still adapting to feedback where needed. I'll leave this one up to you though, it's your package. If you feel like `1.0.0` is appropriate then that's all good.

chicodias: Thanks for the context, let's keep it minor then. I think it makes sense to use the same versioning as the other adapter packages

packages/apps/wallet-adapter-react-example/package.json
chicodias: You're right, we probably missed that — this package shouldn't be used directly at all.

barthuijgen (approved):

PR #3081, opened 2025-08-19
fix(wallet): hotfix the migrate account

sstraatemans:
---
  - https://app.asana.com/0/0/1211090210461999

### DIFF
--- a/packages/libs/kode-ui/src/patterns/FocussedLayout/components/LayoutProvider/FocussedLayoutProvider.tsx
+++ b/packages/libs/kode-ui/src/patterns/FocussedLayout/components/LayoutProvider/FocussedLayoutProvider.tsx
@@ -17,7 +17,7 @@ export const useLayout = (): ILayoutContext => {
   const context = useContext(LayoutContext);
   if (!context) {
     throw new Error(
-      'useOrganisation must be used within a LayoutContextProvider',
+      'useLayout must be used within a FocussedLayoutContextProvider',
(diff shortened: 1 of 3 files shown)


packages/libs/kode-ui/src/patterns/FocussedLayout/components/LayoutProvider/FocussedLayoutProvider.tsx
```diff
@@ -17,7 +17,7 @@ export const useLayout = (): ILayoutContext => {
   const context = useContext(LayoutContext);
   if (!context) {
     throw new Error(
-      'useOrganisation must be used within a LayoutContextProvider',
+      'useLayout must be used within a FocussedLayoutContextProvider',
```
javadkh2, line 20: if it only works for `FocussedLayoutContextProvider` I'd name it `useFocussedLayout`.

javadkh2 (changes requested):
I think the main issue is in the [TxList](https://github.com/kadena-community/kadena.js/blob/4486f66769b6d0b2a5cd36b5ec89c4a2b66a3615/packages/apps/dev-wallet/src/pages/transaction/components/TxList.tsx#L173) component.
It uses a layout-specific component `<FocussedLayoutHeaderAside/>` that should only be used at the page level, not inside a component that can appear on different pages with different layouts.
I believe we need some refactoring to address this properly.

PR #3113, opened 2025-10-02
Improve/iterations-in-encryption

javadkh2:
This PR increases the PBKDF2 iterations from 1,000 to 310,000 to align more closely with OWASP recommendations for secure password-based key derivation.

### DIFF
--- a/packages/libs/hd-wallet/src/utils/crypto.ts
+++ b/packages/libs/hd-wallet/src/utils/crypto.ts
@@ -1,4 +1,4 @@
-export type BinaryLike = string | ArrayBuffer | Uint8Array | Buffer;
+export type BinaryLike = string | ArrayBuffer | Uint8Array;
 
 /**
  *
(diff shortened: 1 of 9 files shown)


packages/libs/hd-wallet/src/utils/crypto.ts
```diff
@@ -9,21 +9,36 @@ export type BinaryLike = string | ArrayBuffer | Uint8Array;
 export const randomBytes = (size: number) =>
   crypto.getRandomValues(new Uint8Array(size));
 
+export const ToArrayBuffer = (data: BinaryLike): ArrayBuffer => {
```
EnoF, line 12: is the capital `T` intended?

javadkh2: No will chanage it it

barthuijgen (approved):

PR #3132, opened 2026-05-02
refactor(hd-wallet): migrate ed25519-keygen to micro-key-producer

DeanVerjans:
The [ed25519-keygen](https://www.npmjs.com/package/ed25519-keygen) package has been deprecated on npm and should be migrated to [micro-key-producer](https://www.npmjs.com/package/micro-key-producer). Since its functionality has been merged into micro-key-producer, this change requires only minimal updates to resolve the deprecation warning.

### DIFF
--- a/packages/libs/hd-wallet/package.json
+++ b/packages/libs/hd-wallet/package.json
@@ -57,7 +57,7 @@
     "buffer": "6.0.3",
     "buffer-from": "1.1.2",
     "debug": "4.3.4",
-    "ed25519-keygen": "0.4.8"
+    "micro-key-producer": "0.8.6"
   },
   "devDependencies": {
     "@kadena-dev/eslint-config": "workspace:*",
--- a/packages/libs/hd-wallet/src/SLIP10/utils/sign.ts
+++ b/packages/libs/hd-wallet/src/SLIP10/utils/sign.ts
@@ -1,6 +1,6 @@
 import { signHash } from '@kadena/cryptography-utils';
 
-import { HDKey } from 'ed25519-keygen/hdkey';
+import { HDKey } from 'micro-key-producer/slip10.js';
 import { uint8ArrayToHex } from '../../utils/buffer-helpers.js';
 
 export interface ISignatureWithPublicKey {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -211,7 +211,7 @@ importers:
         version: 4.3.2(typescript@5.4.5)(vite@5.4.9(@types/node@22.9.1)(lightningcss@1.25.1)(terser@5.31.1))
       vitest:
         specifier: ^1.6.0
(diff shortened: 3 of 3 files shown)


DaisukeFlowers (commented):
Thanks for tackling this @DeanVerjans — swapping off a deprecated crypto dependency in the key-derivation path is exactly the kind of hygiene that's worth doing. I reviewed it and also validated it hands-on. Short version: **the change is correct and behavior-preserving — I verified the derived keys are byte-for-byte identical to the old library.** I have two things to flag, but neither is a reason to hold the code itself.

**What I checked.** I built the branch and ran the `hd-wallet` suite — **all 41 tests pass** (30 hd-wallet + 11 chainweaver), and API Extractor builds clean, so the `micro-key-producer/slip10.js` import doesn't break the type surface. More importantly, I wrote an equivalence harness that installs *both* `ed25519-keygen@0.4.8` and `micro-key-producer@0.8.6` and derives keys from identical seeds, mirroring `deriveKeyPair` exactly (`HDKey.fromMasterSeed(seed).derive(path, true)`):

- **15/15 derivations byte-for-byte identical** (`privateKey` *and* `publicKey`), across the real Kadena path `m/44'/626'/0'`, additional account indices, the max hardened index `2147483647'`, a deeper 5-level path, and 3 seeds (a fixed BIP-39 mnemonic + two raw seeds).
- Example — new lib, `abandon…about` mnemonic, `m/44'/626'/0'` → pubKey `0060ec71ef5df37ee922b272edf60590158938d6a6e0d385d506de913ad3f2be3d`, matching the old lib exactly.

So this is a safe, key-preserving migration — existing wallets derive the same accounts. It's also the vendor-recommended path: `ed25519-keygen`'s own deprecation notice points to `micro-key-producer`, same author.

**1. (Non-blocking, but worth doing here) The suite can't actually catch a derivation regression.** The reason I ran the harness is that the existing SLIP-10 tests (e.g. `kadenaSignWithSeed.test.ts`) generate a *random* mnemonic and only assert `sig.length > 0` — they never pin a known seed to an expected key. So a future derivation change could break every user's accounts and the suite would still pass green. Since I've already computed the equivalence vectors, could we fold a **known-answer test** into this PR — fixed seed → asserted exact `publicKey`/`privateKey` at a couple of paths? It locks in what I just verified and guards this path forever. I'm happy to hand over the vectors/harness.

**2. (Scope) `ed25519-keygen` isn't fully removed from the repo.** It's still a direct dependency of three other packages — `apps/dev-wallet`, `apps/chainweaver-wallet-connect-plugin`, and `apps/chainweaver-pact-console-plugin` — so the deprecation warning persists in the tree after this merges. Fine to keep this PR focused on `hd-wallet`; I'll open a follow-up issue to finish migrating those three so the goal is actually closed out.

**On CI:** the only red check is Vercel ("Authorization required to deploy") — an org/deploy-permission issue unrelated to this code; the `labeler` check passes. Flagging so it isn't mistaken for a real failure.

Net: the code is correct and verified. I'd merge it as-is; adding the known-answer test (item 1) would make it airtight, and item 2 is a follow-up I'll track. Nice cleanup.

